PR #6, opened 2017-12-20
Add command to create a Kubernetes cluster

itowlson:
Addresses https://github.com/Azure/vscode-kubernetes-tools/issues/2.  This implementation includes only creating ACS and AKS clusters, but it should be possible to add other Kubernetes implementations (EKS, GKE, minikube, etc.) over time.  (The code is a bit messy but I'd be happy to work with anyone wanting to contribute such a thing.)

### DIFF
--- a/src/createcluster.ts
+++ b/src/createcluster.ts
@@ -215,7 +215,22 @@ async function loginAsync(context: Context, subscription: string) : Promise<Erro
 
 // end TODO
 
+async function resourceGroupExists(context: Context, resourceGroupName: string) : Promise<boolean> {
+    const sr = await context.shell.exec(`az group show -n "${resourceGroupName}" -ojson`);
+    
+    if (sr.code === 0 && !sr.stderr) {
+        return sr.stdout !== null && sr.stdout.length > 0;
+    } else {
+        return false;
+    }
+ 
+}
+
 async function ensureResourceGroupAsync(context: Context, resourceGroupName: string, location: string) : Promise<Errorable<void>> {
+    if (await resourceGroupExists(context, resourceGroupName)) {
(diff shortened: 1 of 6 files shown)


src/createcluster.ts
```diff
@@ -0,0 +1,543 @@
...
+
+export const uriScheme : string = "k8screatecluster";
+
+export function operationUri(operationId: string) : Uri {
+    return Uri.parse(`${uriScheme}://operations/${operationId}`);
+}
+
+export function uiProvider(fs: FS, shell: Shell) : TextDocumentContentProvider & Advanceable {
+    return new UIProvider(fs, shell);
+}
```
khenidak: If we want this to be for any type of kubernetes cloud/on-prem then we probably need to externalize the rendering function per type.

itowlson: There's still a single entry point where we prompt for what kind of cluster to create.  We can do a better job of teasing apart the generic wizard logic, the common prompts, and the Azure-specific pages, but until we or a third party embarks on a new cluster type that would be pretty speculative.  I vote to leave for now, knowing that it will need refactoring later but that we will have better information about commonalities and differences then.

brendandburns:
LGTM, I'm going to merge this and we can iterate...

What do you think about releasing this plugin?

itowlson:
@brendandburns I'm working on getting a "Microsoft" publisher set up.  After that should be pretty quick to release, though I would quite like to get "Configure from AKS" (not just ACS) in there too, and we need to fix or stub out debugging because of VS Code API changes.  Target is to do a Microsoft branded publish this week even if it needs to be another pre-release.

khenidak:
:-) yes please! let us go with whatever we have then we can easily reiterate

Issue #7, opened 2017-12-22
Kubernetes logs should work for any pod


bnookala:
The current implementation of the `kubernetes logs` command will attempt to use the current working directory's name as a label selector when fetching pods to view logs from. This can fail if there is no pod that shares the same label as the directory name. See the following stack trace for details…

```
TypeError: Cannot read property 'metadata' of null
extension.js:492
	at selectPodForApp (/Users/bhargav/code/vscode-kubernetes-tools/out/src/extension.js:708:21)
	at findPodsForApp (/Users/bhargav/code/vscode-kubernetes-tools/out/src/extension.js:715:13)
	at kubectl.invoke (/Users/bhargav/code/vscode-kubernetes-tools/out/src/extension.js:489:13)
	at context.shell.exec.then (/Users/bhargav/code/vscode-kubernetes-tools/out/src/kubectl.js:77:72)
	at <anonymous>
```

bnookala:
Would be happy to take this on… :) I think I had something noticed this issue on the older extension as well.

itowlson:
@bnookala Closing this since your PR is merged (thanks!).  If this is incorrect, please accept my apologies and re-open!

PR #8, opened 2017-12-23
Use quick picker fetching pod logs.

bnookala:
Makes use of the quick picker to fetch logs for pods, removing the requirement that there be a pod in the namespace with a label equivalent to the working directory name. 

Fixes the bug noted in #7.

logging


brendandburns:
LGTM, thanks.

PR #16, opened 2018-01-02
Check the shell environment on pod to fix the terminal launch failure issue

testforstephen:
When the pod is running on an image without bash installed, launching terminal would fail. See the bug https://github.com/Azure/vscode-kubernetes-tools/issues/15

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -949,9 +949,11 @@ function execKubernetes() {
     execKubernetesCore(false);
 }
 
-function terminalKubernetes(explorerNode? : explorer.ResourceNode) {
+async function terminalKubernetes(explorerNode? : explorer.ResourceNode) {
     if (explorerNode) {
-        execTerminalOnPod(explorerNode.id, 'bash');
+        // For those images (e.g. built from Busybox) where bash may not be installed by default, use sh instead.
+        const bash = await checkBash(explorerNode.id);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -989,11 +991,22 @@ function execKubernetesCore(isTerminal) {
 }
 
 function execTerminalOnPod(podName : string, terminalCmd : string) {
-    const terminalExecCmd : string[] = ['exec', '-it', podName, terminalCmd];
+    const terminalExecCmd : string[] = ['exec', '-it', podName, '--', terminalCmd];
     const term = vscode.window.createTerminal(`${terminalCmd} on ${podName}`, kubectl.path(), terminalExecCmd);
     term.show();
 }
 
+function checkBash(podName : string): Promise<boolean> {
```
itowlson: Suggest changing this to `async` - I've been trying to run down the use of explicit promise management in favour of async/await because I feel it's a lot more readable.  To do this, you'll need to expose `kubectl.invokeAsync` on the interface (and I think you can/should remove the `ShellHandler` argument); I think it's worth it though.

itowlson: Rename to e.g. `isBashOnPod`.

testforstephen: Nice suggestion. Use async/await is more clean. Thanks.

testforstephen: fixed.

brendandburns:
LGTM, thanks! Once Ian approves we can merge...

PR #18, opened 2018-01-08
Settings for publishing to VS Marketplace

itowlson:


squillace:
@itowlson does it also have helm commands, or are they separate?

itowlson:
@squillace It includes Helm commands, which I have cunningly left out of the readme.  Thanks for the spot, will fix.

squillace:
as you can imagine, now I'm going to ask the same thing about the draft commands. I'm assuming they're not in this one. But due diligence.....

package.json
```diff
@@ -3,14 +3,21 @@
...
     },
     "license": "MIT",
     "categories": [
+        "Azure",
         "Other"
     ],
+    "keywords": [
+        "kubernetes",
+        "k8s",
+        "azure"
```
squillace: `helm` would belong here, then.

package.json
```diff
@@ -3,14 +3,21 @@
...
     "version": "0.1.0",
-    "publisher": "Microsoft",
+    "preview": true,
+    "publisher": "ms-kubernetes-tools",
     "engines": {
         "vscode": "^1.13.0"
     },
     "license": "MIT",
     "categories": [
+        "Azure",
```
squillace: I'm not clear what `categories` means here. If it means, "the publisher", ok. But these commands are not specific to Azure, are they? they work anywhere we can pick up your config file, correct? 

If that's right, then we need this to say something different, like `Kubernetes`. But I am ignorant what this section is supposed to mean.

itowlson:
Draft commands are included and are listed in the readme.

package.json
squillace: if draft commands are here, then you want draft in the keywords as well as helm.

PR #19, opened 2018-01-08
Use the new VS Code debug API

itowlson:
The old VS Code API for launching a debug session was removed a couple of releases ago, which broke our Kubernetes Debug command.  With this change we invoke the new API and it works again (at least, it works as well as it ever did!).

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "version": "0.1.0",
     "publisher": "Microsoft",
     "engines": {
-        "vscode": "^1.13.0"
+        "vscode": "^1.19.0"
     },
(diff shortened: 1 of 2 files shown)


squillace:
How will did it work? :-)

squillace:
How will did it work? :-)

itowlson:
@squillace It worked well for Node.js services.  For anything else... not so much.  (I tried getting it working for both Python and Ruby services but ran up against blockers in the case of Python and my own ignorance in the case of Ruby, and eventually decided not to burn more cycles on it until I understood All The Things better.)

PR #21, opened 2018-01-09
New kubernetes tree explorer

testforstephen:

- The hierarchy is similar as the k8s dashboard.
- User could switch cluster via the tree explorer.

See the gif for preview:
k8s_tree_explorer


package.json
```diff
@@ -145,6 +149,10 @@
                 {
                     "command": "extension.vsKubernetesRefreshExplorer",
                     "when": "view == extension.vsKubernetesExplorer"
+                },
+                {
+                    "command": "extension.vsKubernetesUseContext",
+                    "when": "view == extension.vsKubernetesExplorer"
```
itowlson: We seem to have two `when` clauses for the `extension.vsKubernetesUseContext` command: is this intentional?

package.json
```diff
@@ -244,6 +252,11 @@
                 "title": "Create Cluster",
                 "category": "Kubernetes"
             },
+            {
+                "command": "extension.vsKubernetesUseContext",
+                "title": "Set to current cluster",
```
itowlson: Is this consistent with the capitalisation of other commands?

src/explorer.ts
```diff
@@ -58,6 +64,20 @@ function isResource(obj: KubernetesObject) : obj is KubernetesResource {
     return !!(<KubernetesResource>obj).resourceId;
 }
 
+async function getClusters(kubectl: Kubectl, host: Host) : Promise<KubernetesObject[]> {
+    const shellResult = await kubectl.invokeAsync("config view -o json");
+    if (shellResult.code !== 0) {
+        return [];
```
itowlson: Should we do anything with stderr here?

src/explorer.ts
```diff
@@ -58,6 +64,20 @@ function isResource(obj: KubernetesObject) : obj is KubernetesResource {
...
+    }
+    const configJson = JSON.parse(shellResult.stdout);
```
itowlson: This is now a JS object, not JSON; consider calling it `config` or `kubectlConfig` or something like that.

src/explorer.ts
```diff
@@ -58,6 +64,20 @@ function isResource(obj: KubernetesObject) : obj is KubernetesResource {
     return !!(<KubernetesResource>obj).resourceId;
 }
 
+async function getClusters(kubectl: Kubectl, host: Host) : Promise<KubernetesObject[]> {
```
itowlson: Is `host` used in this function?

src/explorer.ts
```diff
@@ -91,4 +111,39 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
         this.resourceId = kind.abbreviation + '/' + id;
     }
-}
\ No newline at end of file
+}
+
+class VirtualKubernetesResource implements KubernetesObject, ResourceNode {
+    readonly resourceId: string;
+
+    constructor (readonly id : string, readonly kind : string, readonly metadata?: any) {
```
itowlson: Picky: inconsistent spacing before type annotations

src/extension.ts
```diff
@@ -1282,6 +1283,16 @@ async function createClusterKubernetes(request? : WizardUIRequest) {
     }
 }
 
+async function useContextKubernetes(explorerNode: explorer.ResourceNode) {
```
itowlson: As far as I know (and from what I found doing the first iteration of the tree), users can select commands from the palette even if the commands are targeted at the tree.  If that happens then `explorerNode` will be undefined.  It would be good to trap that situation and describe the intended usage (or display a pick list).

src/extension.ts
```diff
@@ -1282,6 +1283,16 @@ async function createClusterKubernetes(request? : WizardUIRequest) {
...
+    const targetContext = explorerNode.metadata.context;
+    const shellResult = await kubectl.invokeAsync(`config use-context ${targetContext}`);
+    if (!shellResult.code) {
```
itowlson: Inverted condition makes this hard to read.  Consider writing `if (shellResult.code === 0)`.

src/extension.ts
```diff
@@ -1282,6 +1283,16 @@ async function createClusterKubernetes(request? : WizardUIRequest) {
...
+        vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
+    } else {
+        vscode.window.showErrorMessage(`set current cluster failed: ${shellResult.stderr}`);
```
itowlson: Error messages should be sentence cased.

src/explorer.ts
```diff
@@ -91,4 +111,39 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+        }
+        return [];
+    }
+
+    getTreeItem() : vscode.TreeItem | Thenable<vscode.TreeItem> {
+        const treeItem = new vscode.TreeItem(this.id, vscode.TreeItemCollapsibleState.Collapsed);
+        treeItem.contextValue = null;
+        if (this.kind === 'cluster' && !this.metadata.active) {
+            treeItem.collapsibleState = vscode.TreeItemCollapsibleState.None;
+            treeItem.contextValue = "vsKubernetes.resource.cluster";
```
itowlson: Seems like this is not identifying that it's a cluster but specifically that it's an _inactive_ cluster?  If so, consider using a more self-documenting string.

src/explorer.ts
```diff
@@ -91,4 +111,39 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+
+class VirtualKubernetesResource implements KubernetesObject, ResourceNode {
+    readonly resourceId: string;
+
+    constructor (readonly id : string, readonly kind : string, readonly metadata?: any) {
+        this.resourceId = kind + "/" + id;
+    }
+
+    getChildren() : vscode.ProviderResult<KubernetesObject[]> {
+        if (this.kind === 'cluster') {
```
itowlson: Presumably this will never be called for inactive clusters because of the `TreeItemCollapsibleState.None`?  We don't (yet!) want to display resources for inactive clusters because that will require a more substantial rewriting of the infrastructure (which is not a bad thing, just a lot of extra effort!).

src/explorer.ts
```diff
@@ -91,4 +111,39 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+            return [
+                new KubernetesKind(kuberesources.allKinds.node),
+                new VirtualKubernetesResource('Workloads', 'workload'),
```
itowlson: I don't quite get why deployments, jobs and pods are grouped under 'Workloads'.  Is it the 'making it more similar to the k8s dashboard' thing mentioned in the PR comments?  (If so, that's great and I agree we should do it - just me not understanding.)

squillace:
OOOO I'm looking forward to this one. :-)

package.json
testforstephen: YES, the second `when` is intentional. It's used to restrict this command only to be visible at the tree explorer context menu, not listed at the drop down command palette via `Ctrl + Shift + P`.

package.json
testforstephen: Nice catch.

package.json
itowlson: Ooh, that's really handy.  I need to apply this to some of my existing commands I think!

src/explorer.ts
testforstephen: Correct, for inactive clusters, the `getChildren()` will never be called. Just show their names as placeholder, not expandable at the tree explorer.

src/explorer.ts
testforstephen: nice.

src/explorer.ts
testforstephen: nice.

src/explorer.ts
testforstephen: Right. It's same as the k8s dashboard ui. See https://github.com/kubernetes/dashboard

src/extension.ts
testforstephen: See my answer for your first comment, actually the `Set to Current Cluster` command will not be visible at the dropdown command palette.

src/extension.ts
testforstephen: nice. fixed.

src/extension.ts
testforstephen: nice.

src/explorer.ts
testforstephen: fixed.

src/explorer.ts
testforstephen: nice. fixed.

PR #22, opened 2018-01-10
Support configuring the extension from AKS

### LINKED ISSUE #3: Include AKS clusters in 'configure from...' command
Currently the command for configuring the extension only supports ACS.  We should support AKS (and it would be good to provide easier hooks if someone wants to add GKE or EKS).

itowlson:
We already had a 'Configure from ACS' command which would list ACS clusters in the user's Azure account and download and configure kubectl for a selected cluster.  This adds a choice of 'what cluster type do you want to point kubectl at', with options currently of AKS and ACS.  As with 'create cluster', it is intended that this should be extensible to custom clusters, GKE, EKS, etc. but this will currently be quite clunky - we will work with anyone who wants to implement this support, though, rather than speculatively refactoring now.

Fixes #3.

### DIFF
--- a/src/configurefromcluster.ts
+++ b/src/configurefromcluster.ts
@@ -76,6 +76,12 @@ async function next(context: Context, sourceState: OperationState<OperationStage
             const selectedClusterType : string = requestData;
             if (selectedClusterType == 'Azure Kubernetes Service' || selectedClusterType == 'Azure Container Service') {
                 const subscriptions = await getSubscriptionList(context, getClusterCommand(selectedClusterType));
+                if (!subscriptions.result.succeeded) {
+                    return {
+                        last: subscriptions,
+                        stage: OperationStage.PromptForSubscription
+                    };
+                }
                 const pctStateInfo = extend(subscriptions.result, (subs) => { return { clusterType: selectedClusterType, subscriptions: subs }; });
                 return {
                     last: { actionDescription: 'selecting cluster type', result: pctStateInfo },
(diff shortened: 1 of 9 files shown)


squillace:
This extension point should be robust and relatively easy to enable. How is it enabled here for others? Do they just submit updates directly to this extension, or do they create a new extension depending on this one?

itowlson:
@squillace They would need to submit a pull request to this extension, defining the UI steps and screens and the logic to populate them and action the user's choices.  The infrastructure for flowing from one step to the next is there, but I feel we don't yet know enough about how configuring other cluster types would work to design a robust API-style extension mechanism that would work for any environment.  It feels like the sort of thing we could put considerable effort into, and yet still have to come back to the first time a third party tried to hook it.  Sorry.  I see one or two obvious seams that I could prise apart, but I'd still need to work out how to stitch them back together.

squillace:
No apologies necessary. In the readme, or somewhere, I'd love to document exactly what you said above. It means we're happy to take implementations that extend this to work with GCE or any other competitor. It also means we're happy to rethink how we did this so that it can be made better. That's an entirely fine position to take. :-)

itowlson:
I might start a CONTRIBUTING.md for that perhaps.

squillace:
An entirely reasonable thing to do. As this will be the "defacto" kubernetes extension AND OSS for everyone, let's ensure that everyone knows they can play, and that they know how and that we're open to the fact that maybe we didn't do a great job at first whack, taht kind of thing!

src/configurefromcluster.ts
```diff
@@ -162,8 +211,8 @@ async function downloadCli(context: Context) : Promise<any> {
     }
 }
 
-async function getCredentials(context: Context, clusterName: string, clusterGroup: string) : Promise<any> {
-    const cmd = 'az acs kubernetes get-credentials -n ' + clusterName + ' -g ' + clusterGroup;
+async function getCredentials(context: Context, clusterType: string, clusterName: string, clusterGroup: string) : Promise<any> {
+    const cmd = `az ${getClusterCommandAndSubcommand(clusterType)} get-credentials -n ` + clusterName + ' -g ' + clusterGroup;
```
testforstephen: Suggest to use string interpolation syntax for the whole string.

src/configurefromcluster.ts
```diff
@@ -345,18 +425,44 @@ function advanceUri(operationId: string, requestData: string) : string {
...
+function listClustersFilter(clusterType: string): string {
+    if (clusterType == 'Azure Container Service') {
+        return '?orchestratorProfile.orchestratorType==`Kubernetes`';
+    }
+    return '';
+}
 
+function getListClustersCommand(context: Context, clusterType: string) : string {
+    let filter = listClustersFilter(clusterType);
+    let query = '[' + filter + '].{name:name,resourceGroup:resourceGroup}';
```
testforstephen: use string interpolation for string concat.

src/configurefromcluster.ts
```diff
@@ -3,10 +3,10 @@
 import { TextDocumentContentProvider, Uri, EventEmitter, Event, ProviderResult, CancellationToken } from 'vscode';
 import { Shell } from './shell';
 import { FS } from './fs';
-import { Advanceable, Errorable, UIRequest, StageData, OperationState, OperationMap, advanceUri as wizardAdvanceUri, selectionChangedScript as wizardSelectionChangedScript, script, waitScript, styles } from './wizard';
+import { Advanceable, Errorable, UIRequest, StageData, OperationState, OperationMap, advanceUri as wizardAdvanceUri, selectionChangedScript as wizardSelectionChangedScript, script, waitScript, styles, extend } from './wizard';
 import { Context, getSubscriptionList, setSubscriptionAsync } from './azure';
 
-export const uriScheme : string = "acsconfigure";
+export const uriScheme : string = "k8sconfigure";
```
testforstephen: somewhere the string constant is surrounded by single quote, and otherwhere double quote.  And also there are other code style consistent problems. Can we use travis ci to enable tslint validation before mergeing PR? If you're not familar with travis ci, i would like to provide a PR for this.

itowlson: A PR for CI would be great!  Thank you!  We'd also want to add this rule to the tslint config.

brendandburns:
LGTM, thanks!

PR #24, opened 2018-01-12
Add gulp script and .travis.yml to run tslint/build check

testforstephen:
1.  Add gulp script to run tslint check. (Run `gulp tslint` at cmd)
2. Add .travis.yml file 

TODO: 
1. In order to enable travis ci, need one of the repository administrators to activate it on https://travis-ci.org/.
3. After extending tslint latest rules (via ` "extends": "tslint:latest"`), there are many tslint errors in current code, should we need fix them, right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: node_js
+
+node_js:
+- 'node'
+
+os:
+- linux
+- osx
+
+before_install:
+- if [ $TRAVIS_OS_NAME == "linux" ]; then
+      export CXX="g++-4.9" CC="gcc-4.9" DISPLAY=:99.0;
+      sh -e /etc/init.d/xvfb start;
+      sleep 3;
+  fi
+
+install:
+- npm install -g vsce
+- npm install -g typescript
+- npm install -g gulp
+- npm install
+
+script:
+- gulp tslint
+- vsce package
+# - npm test --silent
--- a/tslint.json
+++ b/tslint.json
@@ -1,5 +1,4 @@
 {
-    "extends": "tslint:latest",
     "rules": {
         "only-arrow-functions": [
             true, "allow-declarations", "allow-named-functions"
--- a/tslint.json
+++ b/tslint.json
@@ -1,4 +1,5 @@
 {
+    "extends": "tslint:latest",
     "rules": {
         "only-arrow-functions": [
             true, "allow-declarations", "allow-named-functions"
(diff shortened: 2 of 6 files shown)


tslint.json
```diff
@@ -1,4 +1,5 @@
 {
+    "extends": "tslint:latest",
```
itowlson: If this is, as you say in the PR description, going to cause a large number of tslint warnings then let's leave it out for now and enable important warnings individually until we are in a position to do a blanket enable.  Or we could do a full tslint immediately, but I'd keep that as a separate PR.  We should avoid having a lot of errors in the build, and I don't want to overwhelm the CI changes by having a barrage of lint-related changes in the same PR.  Does that make sense?

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+install:
+- npm install -g vsce
+- npm install -g typescript
+- npm install -g gulp
+- npm install
+
+script:
+- gulp tslint
+- vsce package
+# - npm test --silent
```
itowlson: What would this do if we enabled it?  We do have some unit tests and it would be great if the CI could run them!

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+- 'node'
+
+os:
+- linux
+- osx
+
+before_install:
+- if [ $TRAVIS_OS_NAME == "linux" ]; then
+      export CXX="g++-4.9" CC="gcc-4.9" DISPLAY=:99.0;
+      sh -e /etc/init.d/xvfb start;
```
itowlson: What is this for?

testforstephen: It's for `npm test --silent` command, because this command needs launch a vscode window to execute unit test.

In headless linux, if we want to run any UI related command, need enable xvfb for that.

.travis.yml
testforstephen: Currently when i run `npm test --silent` on my local windows machine, there are some cases failed. I'm not sure whether it's an environment issue or cases issue? So disable it first.

.travis.yml
itowlson: That makes sense - thanks for the explanation!

.travis.yml
itowlson: Agree then, put it on hold.  It runs okay (for me!) via the VS Code "Launch Tests" debug config, but if the command line doesn't work then you're right, let's leave commented out in this PR.

tslint.json
testforstephen: Agree, we can disable the `latest` rules first. And enable it on separate PR later.

itowlson:
I've activated it on the Travis site.

PR #28, opened 2018-01-15
Improve output channel & show progress on status bar for kubectl job

testforstephen:

Implement several enhancements:
1. Let the extension to reuse the same kubenetes output channel to print message. (Previously every time executing a command, create a new output channel. This is not good experience because vscode doesn't provide ui to delete your output channel.)
2.  Syntax highlighting for output channel.
image

3. Show a progress message on status bar for kubectl job.
image


src/kubectl.ts
```diff
@@ -81,6 +85,21 @@ async function invoke(context : Context, command : string, handler? : ShellHandl
     await kubectlInternal(context, command, handler || kubectlDone(context));
 }
 
+async function invokeWithProgress(context : Context, command: string, progressMessage: string, handler?: ShellHandler): Promise<{}> {
```
itowlson: The combination of `async` + callback dates back to when I was starting to transition the code away from Node.js-style callbacks to TypeScript async/await.  We shouldn't really need both - we only have it in `kubectl.invoke` because a lot of the existing code relied on the callback (specifically the wrapper that injected `kubectlDone` as the default handler) and I didn't want to have to rewrite it all.  It would be desirable for new functions to use `async`, return a `Promise`, in this case a `Promise<ShellResult>`, and not take a callback.  It does bother me though if every call site has to call `kubectlDone`... so perhaps it's best to leave this as you have it for now, and address both it and `invoke` later.  Thoughts?

src/kubeChannel.ts
```diff
@@ -0,0 +1,13 @@
+import * as vscode from "vscode";
+
+export const kubeChannel = {
+    channel: vscode.window.createOutputChannel("Kubernetes"),
+
+    showOutput(message: any, title?: string) {
+        if (title) {
+            this.channel.appendLine(`[${title} ${(new Date()).toISOString().replace(/z|t/gi, ' ').trim()}]`);
```
itowlson: Can we extract a named variable here to make the transform more self-documenting?

syntaxes/kubernetes.output.tmLanguage
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
itowlson: The Helm tmLanguage file is in JSON format rather than XML.  Would it be impractical to align the two?  (It's not a big deal, anyone who can maintain one can maintain the other, it just looks a bit odd to me. *grin*)

src/kubeChannel.ts
```diff
@@ -0,0 +1,13 @@
+import * as vscode from "vscode";
+
+export const kubeChannel = {
+    channel: vscode.window.createOutputChannel("Kubernetes"),
+
+    showOutput(message: any, title?: string) {
```
itowlson: I've been preferring to export interfaces for this purpose, so that they can be mocked/faked in tests.  At the moment this is only used in `extension.ts`, though, so we can defer doing that.

src/kubeChannel.ts
```diff
@@ -0,0 +1,13 @@
+import * as vscode from "vscode";
+
+export const kubeChannel = {
+    channel: vscode.window.createOutputChannel("Kubernetes"),
```
itowlson: Feel like this should be private.

itowlson: And read-only.

package.json
```diff
@@ -361,13 +361,24 @@
                     ".tpl"
                 ],
                 "configuration": "./language-configuration.json"
+            },
+            {
+                "id": "kubernetes-output",
```
itowlson: Pardon my ignorance, but I don't understand how this identifies _our_ output as subject to the colorisation rules without affecting other output.  I realise this is probably a limitation of my understanding...!

testforstephen: The vscode output channel is using mimetype `text/x-code-output`. It seems every extension could extend the highlight rules, and it affects all outputs.

syntaxes/kubernetes.output.tmLanguage
testforstephen: Make sense. Json is more readable. Have changed it to json.

src/kubectl.ts
testforstephen: i would prefer to add two interfaces `invokeWithProgress` and `invokeAsyncWithProgress` for the corresponding `invoke` and `invokeAsync`. For old invoke code, we could choose to use `invokeWithProgress` with minor effort.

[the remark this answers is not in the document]
testforstephen: Nice catch.

src/kubeChannel.ts
testforstephen: nice.

src/kubeChannel.ts
testforstephen: nice.

src/kubeChannel.ts
testforstephen: nice.

testforstephen:
@itowlson Any concerns about the PR?

src/kubeChannel.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface IKubeChannel {
+    showOutput(message: any, title?: string);
+}
+
+class KubeChannel implements IKubeChannel {
+    private readonly channel: vscode.OutputChannel = vscode.window.createOutputChannel("Kubernetes");
+
+    showOutput(message: any, title?: string) {
+        if (title) {
+            const HIGHLIGHTING_TITLE = `[${title} ${(new Date()).toISOString().replace(/z|t/gi, ' ').trim()}]`;
```
itowlson: Local variables (even when constant) should be `camelCase` not `SHOUTY_CAPS`.  I also think it would be useful if you pulled out the `(new Date()).toISOString().replace(/z|t/gi, ' ').trim()` expression into a named variable - it takes a bit of eyeballing to figure out what it's doing and it would be good to save the reader that.

itowlson:
@testforstephen One tiny stylistic nitpick, otherwise good to go I reckon!  Thanks!

src/kubeChannel.ts
testforstephen: Nice suggestion. thanks.

PR #34, opened 2018-01-24
Option to configure extension with a newly created cluster

### LINKED ISSUE #33: [Configure from cluster] Unable to connect to the server: net/http: TLS handshake timeout
Reproduce steps:
1. Use `create cluster` command to create an aks cluster in east-us.
2. Then invoke `configure from cluster` command to configure `kubectl` context. The configure command succeeded, but then call any `kubectl` command would get `Unable to connect to the server: net/http: TLS handshake timeout`.

The reason is that:
1. `create cluster` command doesn't wait for the creating operation to finish. 
2. Even though aks cluster is not fully ready, `configure from cluster` wizard still could list the newly created cluster and execute `az aks get-credentails` operation. But actually `kubectl` command would throw errors.

Current extension doesn't provide any tool to help the user to query if the cluster is fully ready.  The user needs to login to azure portal to view the status.

Expected:  `create cluster` to wait for the completion of the create. Or before executing `az aks get-credentails`, calling `az aks wait` to wait for the cluster to be ready.

itowlson:
After creating a cluster, the user can now opt to wait for the cluster to come up, and configure kubectl (and thereby the extension) to point to the new cluster.

Fixes #33, except for the TLS handshake timeout issue which seems more fundamental... Also fixes an issue where if the `create` command failed synchronously (e.g. SSH configuration issues) then this was silently ignored and the extension reported that cluster creation was under way.

### DIFF
--- a/src/azure.ts
+++ b/src/azure.ts
@@ -104,10 +104,10 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
 }
 
 export async function configureCluster(context: Context, clusterType: string, clusterName: string, clusterGroup: string) : Promise<StageData> {
-    const downloadCliPromise = downloadCli(context, clusterType);
+    const downloadKubectlCliPromise = downloadKubectlCli(context, clusterType);
     const getCredentialsPromise = getCredentials(context, clusterType, clusterName, clusterGroup, 5);
 
-    const [cliResult, credsResult] = await Promise.all([downloadCliPromise, getCredentialsPromise]);
+    const [cliResult, credsResult] = await Promise.all([downloadKubectlCliPromise, getCredentialsPromise]);
 
     const result = {
         gotCli: cliResult.succeeded,
@@ -124,8 +124,8 @@ export async function configureCluster(context: Context, clusterType: string, cl
     };
 }
 
-async function downloadCli(context: Context, clusterType: string) : Promise<any> {
-    const cliInfo = installCliInfo(context, clusterType);
+async function downloadKubectlCli(context: Context, clusterType: string) : Promise<any> {
+    const cliInfo = installKubectlCliInfo(context, clusterType);
 
     const sr = await context.shell.exec(cliInfo.commandLine);
     if (sr.code === 0) {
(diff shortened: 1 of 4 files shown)


src/sleep.ts
```diff
@@ -0,0 +1,5 @@
+export function sleep(ms: number) : Promise<void> {
+    return new Promise<void>((resolve) => {
+        setTimeout(resolve, ms);
+    });
+}
```
khenidak: is this last line normal?

src/azure.ts
```diff
@@ -101,3 +102,102 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
         return { succeeded: false, result: null, error: [sr.stderr] };
     }
 }
+
+export async function configureCluster(context: Context, clusterType: string, clusterName: string, clusterGroup: string) : Promise<StageData> {
+    const downloadCliPromise = downloadCli(context, clusterType);
```
khenidak: nit: change name to kubectlcli (got me thinking what cli az?)

src/azure.ts
```diff
@@ -101,3 +102,102 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+            return {
+                succeeded: false,
+                error: sr.stderr
+            };
+        }
+    }
+}
+
+function installCliInfo(context: Context, clusterType: string) {
+    const cmdCore = `az ${getClusterCommandAndSubcommand(clusterType)} install-cli`;
```
khenidak: Shouldn't we check if cli already exists? I imagine users will many clusters and we don't need to re-download cli

src/sleep.ts
itowlson: @khenidak Don't understand this comment.  The last line is a closing brace to match the opening brace of the `sleep` function...?

src/azure.ts
itowlson: This is worth doing, but is a bit more invasive than just a rename.  Let's tackle it as a separate issue, rather than mixing it up in this PR.

PR #36, opened 2018-01-24
Add more context menus for kubernetes explorer

testforstephen:

Add more context menus for kubernetes explorer.
- Add icon for Cluster nodes
- Add `Create Cluster` and `Configure from Cluster` entries in tree navigation bar.
- For cluster node: 
  + Show Cluster Info
  + Delete from kubeconfig
- For namespace:
  + Set to Current Namespace
  + Get
(Namespace is newly added to tree, it lets user to easily switch namespace)
- For Node/Deployment/Pod/Services:
  + Get
  + Delete

See the gif below for preview.
k8s_explorer

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -1,11 +1,26 @@
+import * as path from 'path';
 import * as vscode from 'vscode';
+
 import * as shell from './shell';
 import { Kubectl } from './kubectl';
+import * as kubectlUtils from './kubectlUtils';
 import { Host } from './host';
 import * as kuberesources from './kuberesources';
 
 export function create(kubectl : Kubectl, host : Host) : KubernetesExplorer {
     return new KubernetesExplorer(kubectl, host);
+}
+
+export function createKindTreeNode(kind: kuberesources.ResourceKind) : TreeNode {
+    return new KubernetesKind(kind);
+}
+
+export function createResourceTreeNode(kind: kuberesources.ResourceKind, id: string, metadata?: any) : TreeNode {
+    return new KubernetesResource(kind, id, metadata);
+}
+
+export interface KubernetesObject {
+    readonly id : string;
 }
 
 export interface ResourceNode {
@@ -21,109 +36,111 @@ export class KubernetesExplorer implements vscode.TreeDataProvider<KubernetesObj
     constructor(private readonly kubectl : Kubectl, private readonly host : Host) {}
 
     getTreeItem(element: KubernetesObject) : vscode.TreeItem | Thenable<vscode.TreeItem> {
-        if (element instanceof VirtualKubernetesResource) {
+        if (element instanceof TreeNode) {
             return element.getTreeItem();
         }
-        const collapsibleState = isKind(element) ? vscode.TreeItemCollapsibleState.Collapsed: vscode.TreeItemCollapsibleState.None;
-        const label = isKind(element) ? element.kind.pluralDisplayName : element.id;
-        let treeItem = new vscode.TreeItem(label, collapsibleState);
-        if (isResource(element)) {
-            treeItem.command = {
-                command: "extension.vsKubernetesLoad",
-                title: "Load",
-                arguments: [ element ]
-            };
-            treeItem.contextValue = "vsKubernetes.resource";
-            if (element.resourceId.startsWith('pod')) {
-                treeItem.contextValue = "vsKubernetes.resource.pod";
-            }
-        }
-        return treeItem;
+        return new vscode.TreeItem(element.id, vscode.TreeItemCollapsibleState.None);
     }
 
     getChildren(parent? : KubernetesObject) : vscode.ProviderResult<KubernetesObject[]> {
-        if (parent instanceof VirtualKubernetesResource) {
-            return parent.getChildren();
-        } else if (parent) {
-            return getChildren(parent, this.kubectl, this.host);
+        if (parent) {
+            return (parent instanceof TreeNode) ? parent.getChildren(this.kubectl, this.host) : [];
         }
-        return getClusters(this.kubectl, this.host);
+        return this.getClusters(this.kubectl);
     }
 
     refresh(): void {
         this._onDidChangeTreeData.fire();
     }
 
-}
-
-function isKind(obj: KubernetesObject) : obj is KubernetesKind {
-    return !!(<KubernetesKind>obj).kind;
-}
-
-function isResource(obj: KubernetesObject) : obj is KubernetesResource {
-    return !!(<KubernetesResource>obj).resourceId;
-}
-
-async function getClusters(kubectl: Kubectl, host: Host) : Promise<KubernetesObject[]> {
-    const shellResult = await kubectl.invokeAsync("config view -o json");
-    if (shellResult.code !== 0) {
-        host.showErrorMessage(shellResult.stderr);
-        return [];
+    private async getClusters(kubectl: Kubectl) : Promise<KubernetesObject[]> {
+        const clusters = await kubectlUtils.getClusters(kubectl);
+        return clusters.map((c) => new VirtualKubernetesResource(c.name, 'cluster', c));
     }
-    const kubectlConfig = JSON.parse(shellResult.stdout);
-    const currentContext = kubectlConfig["current-context"];
-    const contexts = kubectlConfig.contexts;
-    return contexts.map((c) => new VirtualKubernetesResource(c.context.cluster, 'cluster', {
-        context: c.name,
-        active: c.name === currentContext
-    }));
 }
 
-async function getChildren(parent : KubernetesObject, kubectl: Kubectl, host: Host) : Promise<KubernetesObject[]> {
-    if (isKind(parent)) {
-        const childrenLines = await kubectl.asLines("get " + parent.kind.abbreviation);
+abstract class TreeNode implements KubernetesObject {
+    constructor(readonly id: string) {
+    }
+    abstract getChildren(kubectl: Kubectl, host : Host) : vscode.ProviderResult<KubernetesObject[]>;
+    abstract getTreeItem() : vscode.TreeItem | Thenable<vscode.TreeItem>;
+}
+
+class KubernetesKind extends TreeNode {
+    constructor(readonly kind: kuberesources.ResourceKind) {
+        super(kind.abbreviation);
+    }
+
+    async getChildren(kubectl: Kubectl, host : Host) : Promise<KubernetesObject[]> {
+        if (this.kind === kuberesources.allKinds.namespace) {
+            const namespaces = await kubectlUtils.getNamespaces(kubectl);
+            return namespaces.map((ns) => new KubernetesResource(this.kind, ns.name, ns));
+        }
+        const childrenLines = await kubectl.asLines("get " + this.kind.abbreviation);
         if (shell.isShellResult(childrenLines)) {
             host.showErrorMessage(childrenLines.stderr);
             return [ { id: "Error" } ];
         }
-        return childrenLines.map((l) => parse(parent.kind, l));
+        return childrenLines.map((l) => this.parse(this.kind, l));
     }
-    return [];
-}
 
-function parse(kind : kuberesources.ResourceKind, kubeLine : string) : KubernetesObject {
-    const bits = kubeLine.split(' ');
-    return new KubernetesResource(kind, bits[0]);
-}
+    getTreeItem() : vscode.TreeItem | Thenable<vscode.TreeItem> {
+        const treeItem = new vscode.TreeItem(this.kind.pluralDisplayName, vscode.TreeItemCollapsibleState.Collapsed);
+        treeItem.contextValue = `vsKubernetes.kind`;
+        return treeItem;
+    }
 
-interface KubernetesObject {
-    readonly id : string;
-}
-
-class KubernetesKind implements KubernetesObject {
-    readonly id: string;
-    constructor(readonly kind: kuberesources.ResourceKind) {
-        this.id = kind.abbreviation;
+    private parse(kind : kuberesources.ResourceKind, kubeLine : string) : KubernetesObject {
+        const bits = kubeLine.split(' ');
+        return new KubernetesResource(kind, bits[0]);
     }
 }
 
-class KubernetesResource implements KubernetesObject, ResourceNode {
+class KubernetesResource extends TreeNode implements ResourceNode {
     readonly resourceId: string;
-    constructor(kind: kuberesources.ResourceKind, readonly id: string) {
+    constructor(readonly kind: kuberesources.ResourceKind, readonly id: string, readonly metadata?: any) {
+        super(id);
         this.resourceId = kind.abbreviation + '/' + id;
     }
+
+    getChildren(kubectl: Kubectl, host : Host) : vscode.ProviderResult<KubernetesObject[]> {
+        return [];
+    }
+
+    getTreeItem() : vscode.TreeItem | Thenable<vscode.TreeItem> {
+        let treeItem = new vscode.TreeItem(this.id, vscode.TreeItemCollapsibleState.None);
+        treeItem.command = {
+            command: "extension.vsKubernetesLoad",
+            title: "Load",
+            arguments: [ this ]
+        };
+        treeItem.contextValue = `vsKubernetes.resource`;
+        if (this.kind === kuberesources.allKinds.namespace) {
+            treeItem.contextValue = `vsKubernetes.resource.${this.kind.abbreviation}`;
+            if (this.metadata.active) {
+                treeItem.label = "*" + treeItem.label;
+            } else {
+                treeItem.contextValue += ".inactive";
+            }
+        } else if (this.kind === kuberesources.allKinds.pod) {
+            treeItem.contextValue = `vsKubernetes.resource.${this.kind.abbreviation}`;
+        }
+        return treeItem;
+    }
 }
 
-class VirtualKubernetesResource implements KubernetesObject, ResourceNode {
+class VirtualKubernetesResource extends TreeNode implements ResourceNode {
     readonly resourceId: string;
 
     constructor (readonly id: string, readonly kind: string, readonly metadata?: any) {
+        super(id);
         this.resourceId = kind + "/" + id;
     }
 
-    getChildren() : vscode.ProviderResult<KubernetesObject[]> {
+    getChildren(kubectl: Kubectl, host : Host) : vscode.ProviderResult<KubernetesObject[]> {
         if (this.kind === 'cluster') {
             return [
+                new KubernetesKind(kuberesources.allKinds.namespace),
                 new KubernetesKind(kuberesources.allKinds.node),
                 new VirtualKubernetesResource('Workloads', 'workload'),
                 new KubernetesKind(kuberesources.allKinds.service)
@@ -140,10 +157,13 @@ class VirtualKubernetesResource implements KubernetesObject, ResourceNode {
 
     getTreeItem() : vscode.TreeItem | Thenable<vscode.TreeItem> {
         const treeItem = new vscode.TreeItem(this.id, vscode.TreeItemCollapsibleState.Collapsed);
-        treeItem.contextValue = null;
-        if (this.kind === 'cluster' && !this.metadata.active) {
-            treeItem.collapsibleState = vscode.TreeItemCollapsibleState.None;
-            treeItem.contextValue = "vsKubernetes.resource.cluster.inactive";
+        treeItem.contextValue = `vsKubernetes.resource.${this.kind}`;
+        if (this.kind === 'cluster') {
+            treeItem.iconPath = vscode.Uri.file(path.join(__dirname, "../../images/k8s-logo.png"));
+            if (!this.metadata.active) {
+                treeItem.collapsibleState = vscode.TreeItemCollapsibleState.None;
+                treeItem.contextValue += ".inactive";
+            }
         }
         return treeItem;
     }
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -11,22 +11,24 @@ export function create(kubectl : Kubectl, host : Host) : KubernetesExplorer {
     return new KubernetesExplorer(kubectl, host);
 }
 
-export function createKindTreeNode(kind: kuberesources.ResourceKind) : KubernetesTreeNode {
-    return new KubernetesKind(kind);
+export function createKubernetesResourceFolder(kind: kuberesources.ResourceKind) : KubernetesObject {
+    return new KubernetesResourceFolder(kind);
 }
 
-export function createResourceTreeNode(kind: kuberesources.ResourceKind, id: string, metadata?: any) : KubernetesTreeNode {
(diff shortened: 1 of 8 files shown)


test/fakes.ts
```diff
@@ -89,7 +90,13 @@ function fakeShellExec(settings : FakeShellSettings, cmd : string) : Promise<She
...
+    const invokeAsync = settings.invokeAsync || ((s : string) => ({ code: 0, stderr: "", stdout: ""}));
     return {
-        asLines: (cmd) => new Promise((resolve, reject) => resolve(asLines(cmd)))
+        asLines: (cmd) => wrapAsPromise(asLines(cmd)),
+        invokeAsync: (cmd) => wrapAsPromise(invokeAsync(cmd)),
     };
 }
+
+export function wrapAsPromise(result: any): Promise<any> {
+    return new Promise((resolve, reject) => resolve(result));
```
andxu: Promise.resovle(result)?

test/fakes.ts
```diff
@@ -89,7 +90,13 @@ function fakeShellExec(settings : FakeShellSettings, cmd : string) : Promise<She
 
 export function kubectl(settings : FakeKubectlSettings = {}) : any {
     const asLines = settings.asLines || ((s : string) => []);
+    const invokeAsync = settings.invokeAsync || ((s : string) => ({ code: 0, stderr: "", stdout: ""}));
     return {
-        asLines: (cmd) => new Promise((resolve, reject) => resolve(asLines(cmd)))
+        asLines: (cmd) => wrapAsPromise(asLines(cmd)),
```
andxu: wrapAsPromise doesn't need to wrapAsPromise, add async before (cmd)

test/fakes.ts
```diff
@@ -33,6 +33,7 @@ export interface FakeShellSettings {
 
 export interface FakeKubectlSettings {
     asLines? : (cmd: string) => string[] | ShellResult;
```
andxu: asLines signature has been changed.

testforstephen: This is for unit test. Create mocked kubectl instance.

test/fakes.ts
testforstephen: nice

src/extension.ts
```diff
@@ -1288,6 +1312,29 @@ async function useContextKubernetes(explorerNode: explorer.ResourceNode) {
...
+async function deleteContextKubernetes(explorerNode: explorer.ResourceNode) {
+    const answer = await vscode.window.showInformationMessage(`Do you want to delete the cluser '${explorerNode.id}' from the kubeconfig?`, "NO", "YES");
+    if (answer === "YES" && (await kubectlUtils.deleteCluster(kubectl, explorerNode.metadata))) {
+        vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
+    }
+}
+
+async function useNamespaceKubernetes(explorerNode: explorer.KubernetesObject) {
+    if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
+        vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
```
andxu: add return

test/fakes.ts
testforstephen: nice.

src/extension.ts
```diff
@@ -498,7 +502,7 @@ function loadKubernetesCore(value : string) {
         }
 
         const filename = value.replace('/', '-');
-        const filepath = path.join(vscode.workspace.rootPath, filename + '.json');
+        const filepath = path.join(vscode.workspace.rootPath || "", filename + '.json');
```
andxu: rootPath will be deprecated?

src/explorer.ts
```diff
@@ -21,109 +36,111 @@ export class KubernetesExplorer implements vscode.TreeDataProvider<KubernetesObj
...
+    }
 }
 
-function isKind(obj: KubernetesObject) : obj is KubernetesKind {
-    return !!(<KubernetesKind>obj).kind;
-}
-
-function isResource(obj: KubernetesObject) : obj is KubernetesResource {
-    return !!(<KubernetesResource>obj).resourceId;
+abstract class TreeNode implements KubernetesObject {
```
andxu: KubernetesTreeNode? TreeNode is too common.

src/extension.ts
testforstephen: There are other places using the deprecated workspace API, i prefer to change the legacy code in a new PR.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
+import * as vscode from "vscode";
```
itowlson: I'm trying to use the `Host` interface in new modules so that tests can inject mocks and verify that e.g. errors are displayed when expected.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
+import * as vscode from "vscode";
+import { Kubectl } from "./kubectl";
+
+export type Cluster = {
```
itowlson: This doesn't seem like a normal use of type aliasing (I had to look up the `type` keyword to work out what it did!).  Consider making these interfaces with read-only members.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+        vscode.window.showErrorMessage(shellResult.stderr);
+        return null;
+    }
+    return JSON.parse(shellResult.stdout);
+}
+
+export async function getClusters(kubectl: Kubectl): Promise<Cluster[]> {
+    const kubectlConfig = await getKubeconfig(kubectl);
+    if (!kubectlConfig) {
+        return;
```
itowlson: Not keen on a bald `return;` in a function that is meant to return a value.  (I'm surprised the compiler doesn't complain about this!  We should consider turning it on in the linter if that supports it.)  If we are comfortable with returning `undefined` then let's make that explicit by writing `return undefined;`.  Let's also be clear in that case about when we use `null` vs `undefined` - the internal `getKubeconfig` function returns `null` for failure, rather than `undefined`.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export type Namespace = {
+    name: string;
+    active: boolean;
+};
+
+async function getKubeconfig(kubectl: Kubectl): Promise<any> {
+    const shellResult = await kubectl.invokeAsync("config view -o json");
+    if (shellResult.code !== 0) {
+        vscode.window.showErrorMessage(shellResult.stderr);
```
itowlson: Should we `await` this?  I guess it's not necessary.

src/kubectlUtils.ts
itowlson: If the caller is not expected to handle `undefined/null` then consider returning an empty array... long term I'd like to get to all functions being clear about whether they are guaranteed to return a value, or only optionally (e.g. `Promise<Cluster[]>` vs `Promise<Cluster[] | null>`.  That might be a long road though!

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+            user: c.context.user,
+            active: c.name === currentContext
+        };
+    });
+}
+
+export async function deleteCluster(kubectl: Kubectl, cluster: Cluster): Promise<boolean> {
+    const deleteClusterResult = await kubectl.invokeAsyncWithProgress(`config delete-cluster ${cluster.name}`, "Deleting cluster...");
+    if (deleteClusterResult.code !== 0) {
+        await vscode.window.showErrorMessage(`Failed to delete the specified cluster ${cluster.name} from the kubeconfig: ${deleteClusterResult.stderr}`);
```
itowlson: Will this fit in the small VS Code message window?  Might want to look at the "Failed, see output window for why" cop-out that we use in a couple of other places where the size limit was a problem...

src/kubectlUtils.ts
itowlson: Looks like you do `await` message APIs in some other places in this file.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const deleteUserResult = await kubectl.invokeAsyncWithProgress(`config unset users.${cluster.user}`, "Deleting user...");
+    if (deleteUserResult.code !== 0) {
+        await vscode.window.showErrorMessage(`Failed to delete user info for cluster ${cluster.name} from the kubeconfig: ${deleteUserResult.stderr}`);
+        return false;
+    }
+
+    const deleteContextResult = await kubectl.invokeAsyncWithProgress(`config delete-context ${cluster.context}`, "Deleting context...");
+    if (deleteContextResult.code !== 0) {
+        vscode.window.showErrorMessage(`Failed to delete the specified cluster's context ${cluster.context} from the kubeconfig: ${deleteContextResult.stderr}`);
```
itowlson: This function seems a bit inconsistent about `await`-ing error messages or not.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+        return false;
+    }
+
+    vscode.window.showInformationMessage(`Successfully delete the associated cluster/context/user for cluster '${cluster.name}' from the kubeconfig.`);
```
itowlson: delete -> deleted

I would drop 'Successfully' - I don't think 'Successfully deleted' says anything more than 'Deleted' - and rephrase to something like "cluster X and associated data" - 'the associated cluster for cluster X' sounds odd.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    return true;
+}
+
+export async function getNamespaces(kubectl: Kubectl): Promise<Namespace[]> {
+    const shellResult = await kubectl.invokeAsync("get namespaces -o json");
+    if (shellResult.code !== 0) {
+        vscode.window.showErrorMessage(shellResult.stderr);
+        return [];
```
itowlson: Ah, cool.  If returning an empty array on failure works for this function, can we also use that idiom in `getClusters` please?  Unless there is a specific reason callers need a separate failure signal in `getClusters`?

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+    const nsObj = JSON.parse(shellResult.stdout);
+    const cns = await currentNamespace(kubectl);
```
itowlson: `currentNamespace` or `currentNS` or something - `c` is not sufficient.

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    return true;
+}
+
+export async function getNamespaces(kubectl: Kubectl): Promise<Namespace[]> {
+    const shellResult = await kubectl.invokeAsync("get namespaces -o json");
+    if (shellResult.code !== 0) {
+        vscode.window.showErrorMessage(shellResult.stderr);
+        return [];
+    }
+    const nsObj = JSON.parse(shellResult.stdout);
```
itowlson: Is the `Obj` suffix adding any value here?

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    const cns = await currentNamespace(kubectl);
+    return nsObj.items.map((item) => {
+        return {
+            name: item.metadata.name,
+            active: item.metadata.name === cns
+        };
+    });
+}
+
+export async function currentNamespace(kubectl: Kubectl): Promise<string> {
```
itowlson: Exported but not used outside the file?

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    const kubectlConfig = await getKubeconfig(kubectl);
+    if (!kubectlConfig) {
+        return "";
+    }
+    const ctxName = kubectlConfig["current-context"];
+    const currentContext = kubectlConfig.contexts.find((ctx) => ctx.name === ctxName);
+    if(!currentContext) {
```
itowlson: Space between `if` and opening bracket of condition

src/kubectlUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+        return "";
+    }
+    return currentContext.context.namespace || "default";
+}
+
+export async function switchNamespace(kubectl: Kubectl, namespace: string): Promise<boolean> {
+    const shellResult = await kubectl.invokeAsync("config current-context");
+    if (shellResult.code !== 0) {
+        vscode.window.showErrorMessage(shellResult.stderr);
```
itowlson: Will the `config current-context` error message make sense?  The user has selected "Set to Current Namespace", but the error message will have been written for someone who has issued a "get context" command?

test/explorer.test.ts
```diff
@@ -6,8 +7,8 @@ import * as fakes from './fakes';
 
 import { Host } from '../src/host';
 import { Shell, ShellResult } from '../src/shell';
-import { FS } from '../src/fs';
-import { create as explorerCreate } from '../src/explorer';
+import { fs } from '../src/fs';
```
itowlson: It looks like the original import wasn't used, but I'd still rather not import the implementation instead of the interface.  `../src/fs` is meant to be an abstraction layer that can be injected into a context.  Looking at the usage, you're not injecting fake responses to filesystem calls, so you can just call the system 'fs' package, e.g. `import * as sysfs from 'fs';` - don't bother with the abstraction as the filesystem call is part of the test implementation.

src/kubectlUtils.ts
itowlson: Should this module have unit tests?

src/kubectlUtils.ts
testforstephen: Nice

src/kubectlUtils.ts
testforstephen: nice

src/kubectlUtils.ts
testforstephen: Agree.

src/kubectlUtils.ts
testforstephen: Good suggestion, let me reference the similar UX in other places.

src/kubectlUtils.ts
testforstephen: nice.

src/explorer.ts
```diff
@@ -21,109 +36,111 @@ export class KubernetesExplorer implements vscode.TreeDataProvider<KubernetesObj
...
+    }
 }
 
-function isKind(obj: KubernetesObject) : obj is KubernetesKind {
-    return !!(<KubernetesKind>obj).kind;
-}
-
-function isResource(obj: KubernetesObject) : obj is KubernetesResource {
-    return !!(<KubernetesResource>obj).resourceId;
+abstract class KubernetesTreeNode implements KubernetesObject {
```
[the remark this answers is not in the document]
testforstephen: Previously i always try to keep the old structure, finally it makes the new code in messy. Agree with you, i need refactor this file to clearer logic.

src/explorer.ts
```diff
@@ -21,109 +36,111 @@ export class KubernetesExplorer implements vscode.TreeDataProvider<KubernetesObj
...
-        active: c.name === currentContext
-    }));
-}
 
-async function getChildren(parent : KubernetesObject, kubectl: Kubectl, host: Host) : Promise<KubernetesObject[]> {
-    if (isKind(parent)) {
-        const childrenLines = await kubectl.asLines("get " + parent.kind.abbreviation);
+    async getChildren(kubectl: Kubectl, host : Host) : Promise<KubernetesObject[]> {
+        if (this.kind === kuberesources.allKinds.namespace) {
+            const namespaces = await kubectlUtils.getNamespaces(kubectl);
```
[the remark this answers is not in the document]
testforstephen: yes, i need mark the current namespace with additional symbol (*)

src/extension.ts
```diff
@@ -1282,12 +1306,35 @@ async function useContextKubernetes(explorerNode: explorer.ResourceNode) {
...
+}
+
+async function deleteContextKubernetes(explorerNode: explorer.ResourceNode) {
+    const answer = await vscode.window.showInformationMessage(`Do you want to delete the cluser '${explorerNode.id}' from the kubeconfig?`, "NO", "YES");
+    if (answer === "YES" && (await kubectlUtils.deleteCluster(kubectl, explorerNode.metadata))) {
+        return vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
+    }
+}
+
+async function useNamespaceKubernetes(explorerNode: explorer.KubernetesObject) {
```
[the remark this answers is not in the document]
testforstephen: This is a good question. In my current implementation, for current namespace, i display an additional symbol `(*)` before the name. As to status bar, i need have a discussion with PM.

src/kubectlUtils.ts
testforstephen: i would prefer to remove `await` for showMessageBox in other places.

src/kubectlUtils.ts
testforstephen: The methods on this module are just invoking kubectl command, don't have much special logic. I would unit tests on lower priority.

src/extension.ts
```diff
@@ -45,6 +46,16 @@ const draft = draftCreate(host, fs, shell);
 const configureFromClusterUI = configureFromCluster.uiProvider(fs, shell);
 const createClusterUI = createCluster.uiProvider(fs, shell);
 
+const deleteMessageItems: vscode.MessageItem[] = [
+    {
+        title: "Delete"
+    },
+    {
+        title: "Close",
+        isCloseAffordance: true
```
itowlson: Hah, I didn't know you could override the Close button this way!  Nice find!  I suggest you name the button Cancel instead of Close in that case?  There are a couple of other places I could do with doing something similar - thanks for finding it!

src/extension.ts
```diff
@@ -1031,16 +1051,38 @@ function syncKubernetes() {
...
-            kubectl.invoke('delete ' + commandArgs);
+const deleteKubernetes = async (explorerNode? : explorer.ResourceNode) => {
+    if (explorerNode) {
+        const answer = await vscode.window.showWarningMessage(`Do you want to delete the resource '${explorerNode.resourceId}'?`, ...deleteMessageItems);
+        if (answer.isCloseAffordance) {
+            return;
         }
-    });
+        const shellResult = await kubectl.invokeAsyncWithProgress(`delete ${explorerNode.resourceId}`, `Deleting ${explorerNode.resourceId}...`);
+        if (shellResult.code !== 0) {
```
itowlson: Lines 1061-1067 look identical to lines 1076ff.  Should we extract this into a `reportDeleteResult` function perhaps?

src/extension.ts
```diff
@@ -1278,16 +1320,42 @@ async function createClusterKubernetes(request? : WizardUIRequest) {
     }
 }
 
-async function useContextKubernetes(explorerNode: explorer.ResourceNode) {
+async function useContextKubernetes(explorerNode: explorer.KubernetesObject) {
     const targetContext = explorerNode.metadata.context;
     const shellResult = await kubectl.invokeAsync(`config use-context ${targetContext}`);
     if (shellResult.code === 0) {
-        vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
+        return vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
```
itowlson: We do this in quite a few places - I would feel more comfortable if we extracted a `refreshExplorer()` method to decouple all these methods from the command ID.

itowlson: The containing method doesn't return a value, so a `return <expr>;` statement feels a bit odd.  Perhaps `refreshExplorer(); return;` instead?

src/extension.ts
```diff
@@ -1278,16 +1320,42 @@ async function createClusterKubernetes(request? : WizardUIRequest) {
...
+    }
+}
+
+async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject) {
+    const answer = await vscode.window.showWarningMessage(`Do you want to delete the cluster '${explorerNode.id}' from the kubeconfig?`, ...deleteMessageItems);
+    if (answer.isCloseAffordance) {
+        return;
+    }
+    if (await kubectlUtils.deleteCluster(kubectl, explorerNode.metadata)) {
+        return vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
```
itowlson: Again doesn't feel right returning a value (even a lack of value!) here...

src/kubectlUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+            active: c.name === currentContext
+        };
+    });
+}
+
+export async function deleteCluster(kubectl: Kubectl, cluster: Cluster): Promise<boolean> {
+    const deleteClusterResult = await kubectl.invokeAsyncWithProgress(`config delete-cluster ${cluster.name}`, "Deleting cluster...");
+    if (deleteClusterResult.code !== 0) {
+        kubeChannel.showOutput(`Failed to delete the specified cluster ${cluster.name} from the kubeconfig: ${deleteClusterResult.stderr}`, `Delete-${cluster.name}`);
+        vscode.window.showErrorMessage("Failed. See the output window for more details.");
```
itowlson: I think the first sentence might strip _too much_ context.  I'd suggest at minimum 'Delete failed.' and perhaps 'Delete ${cluster.name} failed.'  We're not completely strapped for space in the header - we just can't always handle what can be lengthy error messages.

itowlson: I think we capitalise 'output' elsewhere, as in 'See Output window for...'

src/kubectlUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (!currentContext) {
+        return "";
+    }
+    return currentContext.context.namespace || "default";
+}
+
+export async function switchNamespace(kubectl: Kubectl, namespace: string): Promise<boolean> {
+    const shellResult = await kubectl.invokeAsync("config current-context");
+    if (shellResult.code !== 0) {
+        kubeChannel.showOutput(`Failed. Cannot get the current context: ${shellResult.stderr}`, `Switch-namespace-${namespace}`);
```
itowlson: I'm not sure why the title is hyphenated - it's not a command, just a phrase/header?  Why not use spaces?

src/kubectlUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+        vscode.window.showErrorMessage("Failed. See the output window for more details.");
+        return false;
+    }
+    const updateResult = await kubectl.invokeAsyncWithProgress(`config set-context ${shellResult.stdout.trim()} --namespace="${namespace}"`,
+        "Switching namespace...");
+    if (updateResult.code !== 0) {
+        kubeChannel.showOutput(`Failed to switch the namespace: ${shellResult.stderr}`, `Switch-namespace-${namespace}`);
```
itowlson: Same comment on kubechannel output title.

src/extension.ts
testforstephen: `refreshExplorer` needs access a local variable `treeProvider` from `activate()`. Don't find a good way to extract it in the outside like other command handlers. Prefer to keep current design.

itowlson: Sorry I wasn't clear.  The implementation of `refreshExplorer()` would still be `vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");`.  It just prevents repeating the command-based implementation and the command ID in multiple places.

testforstephen: Got it. Nice.

testforstephen:
@itowlson Thanks for the review. I have fixed the issues, could you help have a look again?

PR #45, opened 2018-02-01
Command telemetry

itowlson:
This PR provides limited telemetry on extension usage.  It collects data about which commands are requested, but not about the inputs or results of those commands, nor about the use of other features such as intellisense or the tree view.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
         "draft",
         "azure"
     ],
+    "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
     "icon": "images/k8s-logo.png",
     "activationEvents": [
         "onCommand:extension.vsKubernetesCreate",
@@ -485,7 +486,8 @@
         "uuid": "^3.1.0",
         "yamljs": "0.2.10",
         "compare-versions": "^3.1.0",
-        "lodash": ">3"
+        "lodash": ">3",
+        "vscode-extension-telemetry": "^0.0.6"
     },
     "devDependencies": {
         "@types/mocha": "^2.2.32",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -153,7 +153,7 @@ export function activate(context) {
         vscode.languages.registerCodeLensProvider(HELM_REQ_MODE, new HelmRequirementsCodeLensProvider()),
 
         // Telemetry
-        new Reporter(context)
+        registerTelemetry(context)
     ];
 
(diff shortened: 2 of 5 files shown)


src/extension.ts
```diff
@@ -148,7 +150,10 @@ export function activate(context) {
         vscode.window.registerTreeDataProvider('extension.vsKubernetesExplorer', treeProvider),
 
         // Code lenses
-        vscode.languages.registerCodeLensProvider(HELM_REQ_MODE, new HelmRequirementsCodeLensProvider())
+        vscode.languages.registerCodeLensProvider(HELM_REQ_MODE, new HelmRequirementsCodeLensProvider()),
+
+        // Telemetry
+        new Reporter(context)
```
testforstephen: `new Reporter()` but don't assign it to any variable. It feels a little wierd to me. 
Could change it to `initializeReporter(context)`?

package.json
```diff
@@ -20,6 +20,7 @@
         "draft",
         "azure"
     ],
+    "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
```
testforstephen: is this a test key or the final production key?

src/extension.ts
itowlson: It is stored in the subscriptions array.  Agree it looks a bit weird because it's surrounded by expressions which are more obviously side-effectful, but I'd rather not change the name because this source is provided by the Azure VS Code SDK.  We could put a wrapper function in `extension.ts` if it feels that awkward though.

package.json
itowlson: It's a production key, but it seems that other open source Microsoft extensions do the same thing (e.g. https://github.com/Microsoft/vscode-docker/blob/master/package.json).  It gave me the jitters too, but I don't see a way of concealing it because even if we added it as part of a publish process it would still be in the `package.json` in the VSIX.

testforstephen: got it.

src/extension.ts
testforstephen: yeah, we could wrap it to a meaningful function in extension.ts. Here if i don't read the implementation, i don't get what's for `new Reporter()`.

itowlson: Good point.  If it's not readable then we should make it readable.  Added a `registerTelemetry` function.  Thanks!

squillace:
Is that the text at intend to use?

squillace:
#sign-off #lgtm

PR #46, opened 2018-02-05
Define debug framework and add java debug implementation

testforstephen:

Plan to split the local debug support to several PRs for review.
- <b>PR1: Define debug framework and add java debug implementation.</b>
- PR2: Add node debug implementation.
- PR3: Add debug (attach) support.
- PR4: Enable the new debug framework via kubernetes commands `Debug(Launch)` and `Debug(Attach)`
- PR5: Support docker-compose.yml ?


src/debug/debugService.ts
```diff
@@ -0,0 +1,243 @@
...
+            return;
+        } else if (!await this.debugProvider.isDebuggerInstalled()) { // Check the required debugger extension is installed or not.
+            return;
+        }
+
+        const cwd = workspaceFolder.uri.fsPath;
+        const imagePrefix = vscode.workspace.getConfiguration().get("vsdocker.imageUser", null);
+        const containerEnv= {};
+        const portInfo = await this.debugProvider.getDockerResolver().resolvePortsFromFile(this.dockerParser, containerEnv);
+        if (!portInfo.debug) {
```
andxu: Shall we provide a wizard to let user specify the port?

src/debug/debugService.ts
```diff
@@ -0,0 +1,243 @@
...
+        });
+    }
+
+    private async startDebugging(workspaceFolder: string, sessionName: string, proxyDebugPort: any, proxyAppPort: any, callback: () => Promise<any>) {
+        const disposables: vscode.Disposable[] = [];
+        disposables.push(vscode.debug.onDidStartDebugSession((debugSession) => {
+            if (debugSession.name === sessionName) {
+                kubeChannel.showOutput("The debug session is started, you could start debugging your application now.", "Start debugging");
+                if (proxyAppPort) {
+                    kubeChannel.showOutput(`The local proxy url for your service is http://localhost:${proxyAppPort}`, "Proxy for service");
```
andxu: give an output verbose of the app port specified by Dockerfile

src/debug/debugService.ts
```diff
@@ -0,0 +1,243 @@
...
+                if (!isProxyReady) {
+                    reject("Cannot setup port-forward.");
+                    return;
+                }
+                resolve();
+            });
+        });
+    }
+
+    private async startDebugging(workspaceFolder: string, sessionName: string, proxyDebugPort: any, proxyAppPort: any, callback: () => Promise<any>) {
```
andxu: rename callback

src/debug/debugService.ts
```diff
@@ -0,0 +1,243 @@
...
+        }));
+
+        disposables.push(vscode.debug.onDidTerminateDebugSession(async (debugSession) => {
+            if (debugSession.name === sessionName) {
+                disposables.forEach((d) => d.dispose());
+                await callback();
+            }
+        }));
+
+        const success = this.debugProvider.startDebugging(workspaceFolder, sessionName, proxyDebugPort);
```
andxu: +await

src/debug/dockerfileParser.ts
```diff
@@ -0,0 +1,78 @@
...
+export class DockerfileParser implements IDockerParser {
+    readonly dockerfile: IRawDockerfile;
+
+    constructor(private readonly dockerfilePath: string) {
+        this.dockerfile = new RawDockerfile(dockerfilePath);
+    }
+
+    getBaseImage(): string {
+        const fromEntries = this.dockerfile.getCommandEntries(["from"]);
+        const baseImagePaths = ((fromEntries.length ? String(fromEntries[0].args) : "").split(":")[0]).split("/");
```
itowlson: This is trying to do too much in one line, and it harms readability.  Suggest something like:

```
if (fromEntries.length === 0) {
  return "";
}
const baseImageTag = fromEntries[0].args;
const baseImageID = baseImageTag.split(":")[0];
const baseImageNameParts = baseImageTag.split("/");
```

src/debug/dockerfileParser.ts
```diff
@@ -0,0 +1,78 @@
...
+        return baseImagePaths[baseImagePaths.length - 1].toLowerCase();
```
itowlson: I would create a `last()` method for this but it's not essential.

src/debug/javaDebugProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+
+// Use the java debugger extension provided by microsoft team for java debugging.
+const defaultJavaDebuggerExtension = "vscjava.vscode-java-debug";
+
+export class JavaDebugProvider implements IDebugProvider {
+    constructor(readonly dockerResolver: IDockerResolver) {
+    }
+
+    public getDebuggerType(): string {
+        return "java";
```
itowlson: So this is what, the VS Code language ID?

src/debug/javaDebugProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public async isDebuggerInstalled(): Promise<boolean> {
+        if (vscode.extensions.getExtension(defaultJavaDebuggerExtension)) {
+            return true;
+        }
+        const answer = await vscode.window.showInformationMessage(`Please install java debugger extension '${defaultJavaDebuggerExtension}' before debugging.`, "Install Now");
```
itowlson: Possible wording: "Java debugging requires the 'Debugger for Java' extension. Would you like to install it now?"

We need to talk to @squillace about how he feels about baking this into the base extension.  It seems like the sort of thing he'd like to externalise, but I also recognise we need some way to point people at the right debugger for their language (if they don't have one, or one that works remotely, already).

src/debug/javaDockerResolver.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export class JavaDockerResolver implements IDockerResolver {
+    public isSupportedImage(baseImage: string): boolean {
+        return baseImage.indexOf("java") >= 0
+            || baseImage.indexOf("openjdk") >= 0
+            || baseImage.indexOf("oracle") >= 0;
+    }
+
+    public async resolvePortsFromFile(dockerParser: IDockerParser, env: {}): Promise<PortInfo> {
+        const portInfo: PortInfo = {
```
itowlson: const to me implies this won't be modified, which it clearly will be.  (I know the _actual_ restriction is that the _reference_ won't be modified, so perhaps I'm being silly and other readers won't take away the same assumption.)

src/debug/javaDockerResolver.ts
```diff
@@ -0,0 +1,71 @@
...
+        // Cannot resolve the debug port from Dockerfile, then ask user to specify it.
+        if (!portInfo.debug) {
+            const input = await vscode.window.showInputBox({
+                prompt: `Please specify debug port exposed by the Dockerfile (e.g. 5005)`,
+                placeHolder: "5005"
+            });
+            portInfo.debug = (input ? input.trim() : null);
+        }
+        if (!portInfo.debug) {
+            return portInfo;
```
itowlson: Seems like in this case we're not really returning anything usable - should we make this function return `Promise<PortInfo | undefined>` and return `undefined` here?

src/debug/javaDockerResolver.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+        
+        // Resolve the app port.
+        const dockerExpose = dockerParser.getExposedPorts();
+        if (portInfo.debug && dockerExpose.length) {
```
itowlson: The `portInfo.debug` test seems superfluous here?

src/debug/javaDockerResolver.ts
```diff
@@ -0,0 +1,71 @@
...
+            const possiblePorts = dockerExpose.filter((port) => port !== portInfo.debug);
+            if (possiblePorts.length === 1) {
+                portInfo.app = possiblePorts[0];
+            } else if (possiblePorts.length > 1) {
+                portInfo.app = await vscode.window.showQuickPick(possiblePorts, { placeHolder: "Please select the app port exposed at Dockerfile" });
```
itowlson: "Choose the application port you want to expose for debugging." perhaps?  ("at Dockerfile" doesn't sound right)

src/debug/javaDockerResolver.ts
```diff
@@ -0,0 +1,71 @@
...
+            }
+            // If the exposed port is a variable, then need set it in environment variables.
+            if (/\$\{?(\w+)\}?/.test(portInfo.app)) {
+                const varName = portInfo.app.match(/\$\{?(\w+)\}?/)[1];
+                env[varName] = defaultJavaAppPort;
+                portInfo.app = defaultJavaAppPort;
```
itowlson: Oh so this is why the app port isn't typed as number?  Can we at least type it as `number | string` and document that it can contain an environment variable?  Or better create a `PortNumber` type which can contain a number or an EV name and has well named members and suitable methods for telling which it is?

itowlson: Oh no wait, the portInfo.app is still a number, you just also add an EV that tells the debuggee to use that port number.

src/debug/providerRegistry.ts
```diff
@@ -0,0 +1,41 @@
+import * as vscode from "vscode";
+
+import { IDebugProvider, IDockerParser } from "./debugInterfaces";
+import { JavaDebugProvider } from "./javaDebugProvider";
+import { JavaDockerResolver } from "./javaDockerResolver";
+
+const supportedProviders: IDebugProvider[] = [
+    new JavaDebugProvider(new JavaDockerResolver())
```
itowlson: I don't think the registry needs to know that the JavaDebugProvider uses the JavaDockerResolver - why can't that be an internal implementation detail of the JavaDebugProvider?

src/debug/providerRegistry.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (supportedProviders.length < 1) {
+        return null;
+    } else if (supportedProviders.length === 1) {
+        return supportedProviders[0];
+    }
+
+    const providerItems = supportedProviders.map((provider) => {
+        return {
+            label: provider.getDebuggerType(),
+            provider
```
itowlson: Oh, so this is shorthand for `provider: provider`?  Today I learned...

src/debug/providerRegistry.ts
```diff
@@ -0,0 +1,41 @@
...
+        };
+    });
+    
+    const pickedProvider = await vscode.window.showQuickPick(<any[]> providerItems, { placeHolder: "Please select the debug type." });
```
itowlson: I don't know what 'debug type' means?

src/debug/providerRegistry.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+    return pickedProvider.provider;
+}
+
+export async function getDebugProvider(baseImage?: string): Promise<IDebugProvider> {
+    let debugProvider = null;
+    if (baseImage) {
+        debugProvider = supportedProviders.find((provider) => provider.getDockerResolver().isSupportedImage(baseImage));
+    }
+    if (!debugProvider) {
```
itowlson: We could also make a stab at working it out from the language ID.

src/debug/providerRegistry.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (!pickedProvider) {
+        return null;
+    }
+    return pickedProvider.provider;
+}
+
+export async function getDebugProvider(baseImage?: string): Promise<IDebugProvider> {
+    let debugProvider = null;
+    if (baseImage) {
+        debugProvider = supportedProviders.find((provider) => provider.getDockerResolver().isSupportedImage(baseImage));
```
itowlson: Just `return` here, then you can skip the if-check below.

src/debug/debugInterfaces.ts
```diff
@@ -0,0 +1,73 @@
...
+import { ShellResult } from "../shell";
+
+export interface PortInfo {
+    debug: string;
+    app: string;
+}
+
+export interface IDebugProvider {
+    /**
+     * The debugger type supported by the provider.
```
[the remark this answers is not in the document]
testforstephen: `debugger type` is actually the language id.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+        let proxyAppPort = 0;
+        if (appPort) {
+            proxyAppPort = await portfinder.getPortPromise({
+                port: appPort
+            });
+            portMapping.push(proxyAppPort + ":" + appPort);
+        }
+        // Find a free local port for forwarding data to remote debug port.
+        const proxyDebugPort = await portfinder.getPortPromise({
+            port: Math.max(10000, Number(proxyAppPort) + 1)
```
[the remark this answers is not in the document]
testforstephen: You'are right. `portfinder` doesn't guarantees `proxyAppPort` is occupied, that's why i start searching from `proxyAppPort + 1` to avoid finding the same free port.

`portfinder` will start searching from the specified base port.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    private async startDebugging(workspaceFolder: string, sessionName: string, proxyDebugPort: any, proxyAppPort: any, onTerminateCallback: () => Promise<any>) {
+        const disposables: vscode.Disposable[] = [];
+        disposables.push(vscode.debug.onDidStartDebugSession((debugSession) => {
+            if (debugSession.name === sessionName) {
+                kubeChannel.showOutput("The debug session is started, you could start debugging your application now.");
+                if (proxyAppPort) {
+                    kubeChannel.showOutput(`You can access the application via the local proxied socket port ${proxyAppPort}, e.g. http://localhost:${proxyAppPort}`);
+                    vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(`http://localhost:${proxyAppPort}`));
```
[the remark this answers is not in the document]
testforstephen: Actually this command will open it in a browser window.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+            return;
+        }
+
+        vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, async (p) => {
+            let appName;
+            try {
+                // Build/push docker image.
+                p.report({ message: "Building docker image..."});
+                kubeChannel.showOutput("Starting to build/push docker image...", "Docker build/push");
+                const shellOpts = Object.assign({ }, shell.execOpts(), { cwd });
```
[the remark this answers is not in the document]
testforstephen: `object literal shorthand`. It's recommended by many popular linters.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+                if (!imagePrefix) {
+                    // In order to allow local kubernetes cluster (e.g. minikube) to have access to local docker images,
+                    // need override docker-env before running docker build.
+                    const dockerEnv = await debugUtils.resolveDockerEnv(this.kubectl);
+                    shellOpts.env = Object.assign({}, shellOpts.env, dockerEnv);
+                }
+                const image = await debugUtils.buildAndPushDockerImage(DockerClient.docker, shellOpts, imagePrefix);
```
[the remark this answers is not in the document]
testforstephen: Yeah, it could be reused by other places. I could give it into a neutral XxxUtils name. But currently i don't want to touch the legacy code yet.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+            });
+            portMapping.push(proxyAppPort + ":" + appPort);
+        }
+        // Find a free local port for forwarding data to remote debug port.
+        const proxyDebugPort = await portfinder.getPortPromise({
+            port: Math.max(10000, Number(proxyAppPort) + 1)
+        });
+        portMapping.push(proxyDebugPort + ":" + debugPort);
+
+        let bin = vscode.workspace.getConfiguration('vs-kubernetes')['vs-kubernetes.kubectl-path'];
```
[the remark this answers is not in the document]
testforstephen: Nice. Could get it via `kubectl.path()`

src/debug/debugUtils.ts
```diff
@@ -0,0 +1,188 @@
...
+import { Kubectl } from "../kubectl";
+import { getKubeconfig } from "../kubectlUtils";
+import { shell, ShellResult } from "../shell";
+import { sleep } from "../sleep";
+
+/**
+ * Build the docker image first. If imagePrefix is not empty, push the image to remote docker hub, too.
+ * 
+ * @param dockerClient the possible dockerClient： docker or docker-compose.
+ * @param shellOpts any option available to Node.js's child_process.exec().
```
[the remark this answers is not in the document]
testforstephen: When run on minikube, docker build needs run with some special environment variables. So need expose shellOpts.

src/debug/debugUtils.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    const status = shellResult.stdout.split(/\s+/)[2];
+    kubeChannel.showOutput(`pod/${podName} status: ${status}`);
+    if (status === "Running") {
+        return;
+    } else if (status !== "ContainerCreating" && status !== "Pending" && status !== "Succeeded") {
+        const logsResult = await kubectl.invokeAsync(`logs pod/${podName}`);
+        kubeChannel.showOutput(`Failed to start the pod "${podName}", it's status is "${status}".\n
+            See more details from the pod logs:\n${logsResult.code === 0 ? logsResult.stdout : logsResult.stderr}`);
+        throw new Error(`Failed to start the pod "${podName}", it's status is "${status}".`);
```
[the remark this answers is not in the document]
testforstephen: Actually `showOutput` focus on showing the pod logs for diagnostics. Throw an error equals to `reject(xxx)`.

src/debug/debugUtils.ts
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Install a vscode extension programmatically.
+ * 
+ * @param extensionId the extension id.
+ */
+export async function installVscodeExtension(extensionId: string): Promise<boolean> {
+    const vscodeCliPath = path.join(path.dirname(process.argv0), "bin", "code");
+    const shellResult = await shell.exec(`"${vscodeCliPath}" --install-extension ${extensionId}`);
+    if (shellResult.code === 0) {
+        const restartAns = await vscode.window.showInformationMessage(`Extension '${extensionId}' was successfully installed. Restart to enable it.`, "Restart Now");
```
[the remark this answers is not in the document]
testforstephen: `Restart` preferred to me.

src/debug/javaDebugProvider.ts
testforstephen: Different debugger has different launch configuration definitions. If we allow the user to specify the debugger by themselves, how to get its debug configuration definition in our code?

That's why i don't depend on java debugger in `package.json` directly, but remind the user to install the required extension when he triggered the java debugging feature.

src/debug/javaDebugProvider.ts
testforstephen: Actually it's the debugger configuration type. Follow vscode development doc, every debugger extension needs register a type for itself. Although in most case, the type equals to the language id.

image

src/debug/debugUtils.ts
```diff
@@ -0,0 +1,188 @@
...
+    await sleep(1000);
+    await waitForRunningPod(kubectl, podName);
+}
+
+/**
+ * Install a vscode extension programmatically.
+ * 
+ * @param extensionId the extension id.
+ */
+export async function installVscodeExtension(extensionId: string): Promise<boolean> {
```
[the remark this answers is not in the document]
testforstephen: This utility is just to install the extension. The check should be done in caller side.

src/debug/debugService.ts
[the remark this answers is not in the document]
testforstephen: Will remove the `open` action, only print the port info in output window.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+        return {
+            proxyProcess: require('child_process').spawn(bin, ["port-forward", podName, ...portMapping]),
+            proxyDebugPort,
+            proxyAppPort
+        };
+    }
+
+    private async waitForProxyReady(proxyProcess: ChildProcess): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            const forwardingRegExp = /Forwarding\s+from\s+127\.0\.0\.1:/;
```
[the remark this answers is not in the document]
testforstephen: In regular expression, the space should be `\s`, and adding an additional `+` does no harm. Right?

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+                kubeChannel.showOutput(`Starting to run image ${image} on kubernetes cluster...`, "Run on kubernetes");
+                appName = await debugUtils.runDockerImageInK8s(this.kubectl, image, [portInfo.app, portInfo.debug], containerEnv);
+                kubeChannel.showOutput(`Finished to run image ${image} as a deployment ${appName} on kubernetes cluster.`);
+
+                // Find the running debug pod.
+                p.report({ message: "Finding the debug pod..."});
+                kubeChannel.showOutput(`Starting to find the pod by the label run=${appName}`, "Find debug pod");
+                const podList = await debugUtils.findPodsByLabel(this.kubectl, `run=${appName}`);
+                if (podList.items.length === 0) {
+                    throw new Error("Failed to find debug pod.");
```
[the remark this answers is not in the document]
testforstephen: There is a catch in the outer block, the error will be handled there.

src/debug/debugInterfaces.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { Kubectl } from '../kubectl';
+import { ShellResult } from "../shell";
+
+export interface PortInfo {
+    debug: string;
+    app: string;
+}
+
+export interface IDebugProvider {
```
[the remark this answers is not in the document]
testforstephen: In our previous project, linter tool suggest us to use `I` prefix for interface, so i keep that habit.

src/debug/dockerfileParser.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public searchInArgs(regularExpression: RegExp, scope?: string[]): RegExpMatchArray {
+        const scopeEntries = (scope ? this.getCommandEntries(scope) : this.commandEntries);
+        for (let entry of scopeEntries) {
+            const args = Array.isArray(entry.args) ? entry.args : [ String(entry.args) ];
+            for (let arg of args) {
+                const matches = arg.match(regularExpression);
+                if (matches && matches.length) {
+                    return matches;
```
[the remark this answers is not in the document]
testforstephen: Rename `getCommandArgs` to mergeCommandArgsOfType, it is only used to get all `expose` ports.

The launch command could be defined via `entrypoint java xxx` or `run java xxx`, so we just need to get first command that matches the regular expression.

src/debug/providerRegistry.ts
testforstephen: It's by purpose. Firstly the extension will guess the language type from image name. If it failed, then pop up a wizard to ask user to select language first.

src/debug/javaDockerResolver.ts
testforstephen: The original port parsed from dockerfile or regular expression is string type, so i'd like to keep it as string.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+        if (!workspaceFolder) {
+            return;
+        }
+        // TODO: Support docker-compose.yml
+        const dockerfilePath = path.join(workspaceFolder.uri.fsPath, "Dockerfile");
+        if (!fs.existsSync(dockerfilePath)) {
+            vscode.window.showErrorMessage(`No Dockerfile found at the workspace ${workspaceFolder.name}`);
+            return;
+        }
+        this.dockerParser = new DockerfileParser(dockerfilePath);
```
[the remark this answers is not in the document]
testforstephen: Agree. It should be a local variable.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+        this.debugProvider = await providerRegistry.getDebugProvider(this.dockerParser.getBaseImage());
+        if (!this.debugProvider) {
+            return;
```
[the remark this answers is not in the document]
testforstephen: No need. Because `getDebugProvider()` will fall back to a dropdown list to ask user to choose the debug provider. If after that, it is still empty, ignore it.

src/debug/debugService.ts
```diff
@@ -0,0 +1,251 @@
...
+        } else if (!await this.debugProvider.isDebuggerInstalled()) { // Check the required debugger extension is installed or not.
+            return;
```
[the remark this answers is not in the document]
testforstephen: No need. `isDebuggerInstalled` has fallback logic to ask user to install the required debugger.

src/debug/dockerfileParser.ts
```diff
@@ -0,0 +1,78 @@
...
+                targetEntries.push(entry);
+            }
+        });
+        return targetEntries;
+    }
+
+    public getCommandArgs(command: string): string[] {
+        let args = [];
+        this.commandEntries.forEach((entry) => {
+            if (entry.name.toLowerCase() === command) {
```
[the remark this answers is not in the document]
testforstephen: Yes, on purpose. Only used to get all `expose` ports.

src/debug/providerRegistry.ts
testforstephen: It's similar as `Select Environment`. In vscode, when your workspace doesn't contain `launch.json` file, and click F5 will ask user to select a environment first.

I would change the tooltip message to `Select Environment`.

src/debug/debugService.ts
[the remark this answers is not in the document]
itowlson: Bah, I misread - you are right, ignore me, sorry!

src/debug/debugService.ts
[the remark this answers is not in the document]
itowlson: I am uncomfortable having this sort of thing in `debugUtils`.  If you don't want to touch the existing code, then at least let's find a good home for the new code, and add a `TODO` to the existing code to say 'switch over to the shared implementation'.

src/debug/debugUtils.ts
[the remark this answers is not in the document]
itowlson: Extensions normally talk about 'reloading', e.g. when you add one from the gallery or from a VSIX.  Given the precedent for 'reload,' can you say more about why you prefer 'restart'?

src/debug/debugService.ts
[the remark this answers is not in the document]
itowlson: Well, it means you need a regex instead of a string contains check, and regexes are more error-prone to maintain.  But it's no big deal for this simple case!

src/debug/javaDockerResolver.ts
itowlson: That doesn't seem like a good reason.  If we followed that logic then everything read from a file or console would be string typed.  Typing it as number communicates what it is to readers of the code.  Even typing it as `number | string` would at least give readers some boundaries on what it might be.  Can you explain the benefit of typing it as `any`?

src/debug/debugService.ts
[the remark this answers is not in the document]
testforstephen: This issue is resolved in the latest commit. No `debugUtils`. I have split it into `dockerUtils` and `kubectlUtils`.

src/debug/debugUtils.ts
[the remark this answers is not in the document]
testforstephen: Ok, i revisit the vscode behavior, `Reload` is more accurate. Will do that.

src/debug/javaDockerResolver.ts
testforstephen: fixed.

src/debug/providerRegistry.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!pickedProvider) {
+        return null;
+    }
+    return pickedProvider.provider;
+}
+
+export async function getDebugProvider(baseImage?: string): Promise<IDebugProvider> {
+    let debugProvider = null;
+    if (baseImage) {
+        debugProvider = supportedProviders.find((provider) => provider.isSupportedImage(baseImage));
```
itowlson: Is it ever possible for a base image to be supported by more than one provider?

src/docker/dockerUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (imagePrefix) {
+        image = imagePrefix + "/" + image;
+    }
+    return image;
+}
+
+async function findVersion(cwd: string): Promise<string> {
+    const shellOpts = Object.assign({ }, shell.execOpts(), { cwd });
+    const shellResult = await shell.execCore('git describe --always --dirty', shellOpts);
+    return shellResult.code !== 0 ? "latest" : shellResult.stdout.trim();
```
itowlson: Reverse this condition to avoid the reader having to go, "if it's not zero then ..., but if it's _not_ not zero..."

src/docker/dockerUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+    const currentCluster = await getCurrentClusterConfig(kubectl);
+    if (!currentCluster || !currentCluster.server || !currentCluster.certificateAuthority) {
+        return {};
+    }
+    
+    if (/^https/.test(currentCluster.server)) {
+        dockerEnv["DOCKER_TLS_VERIFY"] = 1;
+    }
+    const serverUrl = url.parse(currentCluster.server);
+    dockerEnv["DOCKER_HOST"] = `tcp://${serverUrl.hostname}:2376`;
```
itowlson: If I'm pointing at a cluster in the cloud (rather than a local minikube), what is going to happen here?  Do we need to trap for this situation?

src/docker/dockerUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return dockerEnv;
+}
+
+async function dockerApiVersion(): Promise<string> {
+    const versionResult = await shell.exec(`docker version --format "{{.Client.APIVersion}}"`);
+    if (versionResult.code === 0) {
+        return versionResult.stdout.trim();
+    }
+    return "1.23";
```
itowlson: Where does this magic string come from?  Can we make it a named constant?

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,73 @@
+import * as fs from 'fs';
+import { CommandEntry, parse as rawDockerfileParser} from 'docker-file-parser';
```
itowlson: `parse as rawDockerfileParser` sounds like you're aliasing a verb to a noun, which feel like it's going to be misleading.  Can we find a better alias?  Do we need to alias it at all?

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,73 @@
+import * as fs from 'fs';
+import { CommandEntry, parse as rawDockerfileParser} from 'docker-file-parser';
+
+import { IDockerParser } from "./parser";
+import { shell, ShellResult } from "../shell";
+
+class RawDockerfile {
+    private commandEntries: CommandEntry[];
```
itowlson: Can this be declared `readonly`?

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    constructor(private readonly dockerfilePath: string) {
+        const dockerData = fs.readFileSync(dockerfilePath, 'utf-8');
+        this.commandEntries = rawDockerfileParser(dockerData, { includeComments: false });
+    }
+
+    public getCommandsOfType(...commands: string[]): CommandEntry[] {
+        return this.commandEntries.filter((entry) => {
+            const cmdName = entry.name.toLowerCase();
+            return commands.find((key) => key === cmdName);
```
itowlson: Why the name 'key'?

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,73 @@
...
+        let args = [];
+        this.commandEntries.forEach((entry) => {
+            if (entry.name.toLowerCase() === command) {
+                args = args.concat(entry.args);
+            }
+        });
+        return args;
+    }
+
+    public searchInArgs(regularExpression: RegExp, scope?: string[]): RegExpMatchArray {
```
itowlson: Why is this called `scope` instead of `commands` as it was in `getCommandsOfType`?

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,73 @@
...
+        const scopeEntries = (scope ? this.getCommandsOfType(...scope) : this.commandEntries);
+        for (const entry of scopeEntries) {
+            const args = Array.isArray(entry.args) ? entry.args : [ String(entry.args) ];
```
itowlson: The use of `concat` in `mergeCommandsOfType` made it look like `entry.args` was always an array, but here you test it first.  Should it also be tested in `mergeCommandsOfType`?

Unit tests for these methods would be very reassuring.  I'm not sure how you're testing them at the moment so I don't know if I'm worrying about nothing or if it's stuff you've checked and proved to work.

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,73 @@
...
+    private readonly dockerfile;
+
+    constructor(private readonly dockerfilePath: string) {
+        this.dockerfile = new RawDockerfile(dockerfilePath);
+    }
+
+    getBaseImage(): string {
+        const fromEntries = this.dockerfile.getCommandsOfType("from");
+        if (fromEntries.length === 0) {
+            return "";
```
itowlson: Seems like we should maybe return null or undefined here (changing the return type accordingly) rather than returning a valid string which might not be error-checked by the caller.

src/extensionUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Install a vscode extension programmatically.
+ * 
+ * @param extensionId the extension id.
+ */
+export async function installVscodeExtension(extensionId: string): Promise<boolean> {
+    const vscodeCliPath = path.join(path.dirname(process.argv0), "bin", "code");
+    const shellResult = await shell.exec(`"${vscodeCliPath}" --install-extension ${extensionId}`);
+    if (shellResult.code === 0) {
+        const answer = await vscode.window.showInformationMessage(`Extension '${extensionId}' was successfully installed. Restart to enable it.`, "Reload Now");
```
itowlson: "Restart to enable it" -> "Reload..."

src/extensionUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+        if (answer === "Restart Now") {
```
itowlson: Needs to change to "Reload Now" surely?

src/kubectlUtils.ts
```diff
@@ -14,7 +15,12 @@ export interface Namespace {
     readonly active: boolean;
 }
 
-async function getKubeconfig(kubectl: Kubectl): Promise<any> {
+export interface ClusterConfig {
+    readonly server: string;
+    readonly certificateAuthority: string;
+}
+
+export async function getKubeconfig(kubectl: Kubectl): Promise<any> {
```
itowlson: Prefer not to export this if it is returning `any` - can we wrap the features that callers need in more strongly typed methods?

src/kubectlUtils.ts
```diff
@@ -111,3 +130,82 @@ export async function switchNamespace(kubectl: Kubectl, namespace: string): Prom
...
+ * @param image the docker image.
+ * @param exposedPorts the exposed ports.
+ * @param env the additional environment variables when running the docker container.
+ * @return the deployment name.
+ */
+export async function runAsDeployment(kubectl: Kubectl, image: string, exposedPorts: number[], env: any): Promise<string> {
+    let imageName = image.split(":")[0];
+    let imagePrefix = imageName.substring(0, imageName.lastIndexOf("/")+1);
+    let baseName = imageName.substring(imageName.lastIndexOf("/")+1);
+    const deploymentName = `${baseName}-debug-${Date.now()}`;
```
itowlson: This doesn't seem to be in the debug directory, so it shouldn't be hardwired to use debug deployment names.

src/kubectlUtils.ts
```diff
@@ -111,3 +130,82 @@ export async function switchNamespace(kubectl: Kubectl, namespace: string): Prom
...
+    let runCmd = [
+        "run",
+        deploymentName,
+        `--image=${image}`,
+        imagePrefix ? "" : "--image-pull-policy=Never",
+        ...exposedPorts.map((port) => port ? `--port=${port}` : ""),
```
itowlson: When would an exposed port ever be falsy?

src/debug/debugProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+     * @param workspaceFolder the workspace folder path.
+     * @param sessionName the debug session name.
+     * @param port the debugging port exposed by the target program.
+     * @return A thenable that resolves when debugging could be successfully started.
+     */
+    startDebugging(workspaceFolder: string, sessionName: string, port: number): Promise<boolean>;
+
+        /**
+     * The docker image is supported by the provider or not.
+     *
```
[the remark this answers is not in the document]
testforstephen: Nice catch. The two methods above have no parameters, the comment has already explained the return type, i would ignore the duplicated `@return` description. Will add `@param` and `@return` to the three methods below.

src/debug/debugService.ts
```diff
@@ -0,0 +1,273 @@
...
+        if (!workspaceFolder) {
+            return;
+        }
+        // TODO: Support docker-compose.yml
+        const dockerfilePath = path.join(workspaceFolder.uri.fsPath, "Dockerfile");
+        if (!fs.existsSync(dockerfilePath)) {
+            vscode.window.showErrorMessage(`No Dockerfile found in the workspace ${workspaceFolder.name}`);
+            return;
+        }
+        const dockerParser = new DockerfileParser(dockerfilePath);
```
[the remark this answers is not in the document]
testforstephen: Nice suggestion. I will refactor the `dockerParser` interface to two minor interfaces `IDockerParser` and `IDockerfile`. `IDockerParser` keeps a method `parse(filepath): IDockerfile`, and `IDockerfile` keeps three methods `getBaseImage()`, `getExposedPorts()`, `searchLaunchArgs`.

src/debug/debugProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+        /**
+     * The docker image is supported by the provider or not.
+     * 
+     */
+    isSupportedImage(baseImage: string): boolean;
+
+    /**
+     * Resolve the debug port info from the dockerfile.
+     */
+    resolvePortsFromFile(dockerParser: IDockerParser, env: {}): Promise<PortInfo>;
```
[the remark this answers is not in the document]
testforstephen: They are not equivalent. `{}` means it should be an `Object`. `any` could be any thing. Here `{}` is more accurate.

src/debug/debugService.ts
```diff
@@ -0,0 +1,273 @@
...
+        this.debugProvider = await providerRegistry.getDebugProvider(dockerParser.getBaseImage());
+        if (!this.debugProvider) {
+            return;
+        } else if (!await this.debugProvider.isDebuggerInstalled()) {
+            return;
```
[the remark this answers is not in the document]
testforstephen: Correct. `debugProvider.isDebuggerInstalled()` will inform the user earlier.  It'll be `false` only in one case, that the user clicks `Cancel` when the debugger tells the user to install the required debugger extension.

src/debug/javaDebugProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+        const dockerExpose = dockerParser.getExposedPorts();
+        if (dockerExpose.length) {
+            const possiblePorts = dockerExpose.filter((port) => port !== rawDebugPortInfo);
+            if (possiblePorts.length === 1) {
+                rawAppPortInfo = possiblePorts[0];
+            } else if (possiblePorts.length > 1) {
+                rawAppPortInfo = await vscode.window.showQuickPick(possiblePorts, { placeHolder: "Choose the application port you want to expose for debugging." });
+            }
+            // If the exposed port is a variable, then need set it in environment variables.
+            if (/\$\{?(\w+)\}?/.test(rawAppPortInfo)) {
```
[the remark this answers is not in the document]
testforstephen: Nice. I like to show an error message for malformed variable.

src/debug/providerRegistry.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    const providerItems = supportedProviders.map((provider) => {
+        return {
+            label: provider.getDebuggerType(),
+            provider
+        };
+    });
+    
+    const pickedProvider = await vscode.window.showQuickPick(<any[]> providerItems, { placeHolder: "Select the environment" });
```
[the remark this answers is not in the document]
testforstephen: Here i just refer to vscode message, when you click `F5` to start debugging, it'll ask you to `select the environment` first.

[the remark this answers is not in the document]
testforstephen: Of course, we can define a specific type that extends `vscode.QuickPickItem`. Use `any` is a brief implementation.

src/debug/providerRegistry.ts
testforstephen: This's a good concern. This is involved with how intelligent our code can infer the application type. In the `javaDebugProvider#isSupportedImage`, i didn't add too many guess logic, but just consider the classic key words `java`, `openjdk`, `oracle` as a java image. If in such cases, it points to multiple providers, i think it makes sense to the first provider matched by the loop.

src/docker/dockerUtils.ts
testforstephen: This is the current docker version, just want to add a default version. Will assign it to a named constant.

src/docker/dockerfileParser.ts
testforstephen: No need to test it in `mergeCommandsOfType`, because `concat([1])` equals `concat(1)`.

I have test the logic multiple times by running the real application, it worked. For command `FROM`, the arg is just a string, for `EXPOSE/ENTRYPOINT etc.`, the args is array.  Will consider to add some unit tests later.

src/kubectlUtils.ts
testforstephen: Yes, it's an internal helper method in this file, no need to export it.

src/debug/debugService.ts
```diff
@@ -0,0 +1,273 @@
...
+        // TODO
+        return;
+    }
+
+    private async buildDockerImageStep(imagePrefix: string, shellOpts: any): Promise<string> {
+        kubeChannel.showOutput("Starting to build/push docker image...", "Docker build/push");
+        if (!imagePrefix) {
+            // In order to allow local kubernetes cluster (e.g. minikube) to have access to local docker images,
+            // need override docker-env before running docker build.
+            const dockerEnv = await dockerUtils.resolveDockerEnv(this.kubectl);
```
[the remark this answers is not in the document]
testforstephen: It does the same thing as `minikube docker-env`. It's used to get docker env of the kubernetes cluster, and set it to the shell context of running `docker build` via local `docker` client so that the built image is generated in the kubernetes docker directly. 

   This just worked for local `minikube`. For remote kubernetes cluster, the user need specify `imagePrefix` via `vsdocker.imageUser` settings and pushed to `docker hub` or `ACR`.

src/docker/dockerUtils.ts
testforstephen: Actually for cloud, such as `aks`, this method will return an emtpy `{}` because its certificate-authority in kubeconfig is empty. It just has value in local minikube.

src/debug/javaDebugProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }
+
+        return {
+            debugPort: Number(rawDebugPortInfo),
+            appPort: Number(rawAppPortInfo)
+        };
+    }
+
+    public async resolvePortsFromContainer(kubectl: Kubectl, pod: string, container: string): Promise<PortInfo> {
```
itowlson: Consider removing until ready to implement

src/debug/providerRegistry.ts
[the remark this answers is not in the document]
itowlson: I'd feel more comfortable with defining a type.  I know it's a bit fiddly though.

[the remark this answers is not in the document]
itowlson: Okay

src/debug/providerRegistry.ts
itowlson: That sounds fine - we can address that later if it comes up.

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,79 @@
...
+            const args = Array.isArray(entry.args) ? entry.args : [ String(entry.args) ];
+            for (const arg of args) {
+                const matches = arg.match(regularExpression);
+                if (matches && matches.length) {
+                    return matches;
+                }
+            }
+        }
+
+        return null;
```
itowlson: I'd normally return an empty array rather than a null (which requires special handling).

src/docker/dockerfileParser.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+class Dockerfile implements IDockerfile {
+    private readonly dockerfile;
```
itowlson: Declare a type for this field.

src/docker/dockerfileParser.ts
testforstephen: In my understanding, if returning an empty array, the caller still needs use the `array.length` to check it's a valid result. Returning `null` could let the user to boolean expression to check it.

itowlson: But if the caller needs to explicitly check for no matches, don't they need to check the empty array case anyway?  (I know you currently check `matches.length` so that it never returns an empty array, but that's not something I'd trust to remain true as other people maintain the function.)  Either an empty array does 'the right thing' inherently (e.g. a loop scenario) or the caller needs to check for it just in case.

(Or we need a test which asserts if the function ever returns an empty array, so that maintainers are alerted to this detail of the contract.)

testforstephen: Agree. thanks for your opinion.

PR #47, opened 2018-02-06
Support config maps and secrets

testforstephen:

Feature https://github.com/Azure/vscode-kubernetes-tools/issues/30 to support Config Maps and Secrets on tree explorer. @kyliel @squillace

k8s_explorer_configmap

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -130,7 +130,7 @@ class KubernetesWorkloadFolder extends KubernetesFolder {
 
 class KubernetesConfigFolder extends KubernetesFolder {
     constructor() {
-        super("config", "Configs");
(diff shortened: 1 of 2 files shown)


src/explorer.ts
```diff
@@ -127,6 +128,19 @@ class KubernetesWorkloadFolder extends KubernetesFolder {
     }
 }
 
+class KubernetesConfigFolder extends KubernetesFolder {
+    constructor() {
+        super("config", "Configs");
```
itowlson: For the display name I suggest something like "Configuration" rather than "Configs".

testforstephen: Nice. Fixed.

Issue #43, opened 2018-01-30
Run Kubernetes command in "TERMINAL" window in VS Code


kyliel:
Directly show and run that command in “TERMINAL” window. For example, I have a Kubernetes cluster which have 2 pods. I select one pod and click “Get”, the command “kubectl get pod/xxx -o wide” will be executed in “TERMINAL” window. And I also could see the output after this command is executed. Such an experience is also adopted by docker extension, maven extension, Ansible extension and others in VS Code. 

![1](https://user-images.githubusercontent.com/29885760/35552007-e25e894c-05cc-11e8-9077-99100833e608.jpg)

kyliel:
@testforstephen will work on it.

testforstephen:
Migrate @itowlson's comments:

> @itowlson: I feel we need a clearer design philosophy around how this is meant to work. When is a command meant to be run in a terminal? How many terminals do we create? If there are multiple terminals then how does the user tell which is which? We can use issue #43 to discuss possible approaches (and how other extensions do it) if that would help.

@kyliel @squillace  Any opinions about this?

testforstephen:
> When is a command meant to be run in a terminal?

For ui operations, if it is be done by one single kubectl command, then i'd like to run it in terminal. 
The typical cases are `create/get/delete/describe/logs/expose/exec/scale/cluster-info`

Otherwise, hide the detail by using the normal kubectl.invoke.
The typical cases are `run/sync/diff/debug`

> How many terminals do we create? If there are multiple terminals then how does the user tell which is which?

I'd like to share the terminal. Although `vscode-docker` extension always run command in a new terminal, but i feel like the shared terminal is more friendly because k8s extension provides more ui commands (> 20) than docker.

kyliel:
@itowlson, And it is better to get commands in command palette run in terminal tool. For example, if I select "Kubernetes Create", a command "kubectl create -f xxx.yaml" is shown on the terminal and executed on the terminal. Then I know what is happening.

itowlson:
@kyliel Yeah, I think we're all agreed that it's desirable, I just want to align on the _design_ and _behaviour_ so Jinbo doesn't get dragged back and forth through a series of reviews with one person saying "I think it should work like this" and then someone else saying "no, _I_ think it should work like _this_."  My question is about refining the detail, not questioning the broad principle.

kyliel:
[...]

PR #48, opened 2018-02-06
Run kubernetes command in vscode TERMINAL window

testforstephen:

Fix issue #43. @kyliel @squillace


src/extension.ts
```diff
@@ -421,14 +437,14 @@ function maybeRunKubernetesCommandForActiveWindow(command) {
                         vscode.window.showErrorMessage("Save failed.");
                         return;
                     }
-                    kubectl.invoke(`${command} "${editor.document.fileName}"`);
+                    runKubectlCommandInTerminal("kubectl", `${command} "${editor.document.fileName}"`);
```
itowlson: I feel like we're losing some of the encapsulation of the `kubectl` module here.  Have you considered a design where you provide e.g. an `invokeOptions` argument to `kubectl.invoke`, and if the caller specifies a terminal then run it in the terminal?  Open to other ideas too!

src/extension.ts
```diff
@@ -556,22 +572,21 @@ function exposeKubernetes() {
         cmd += ' --port=' + ports[0];
     }
 
-    kubectl.invoke(cmd);
+    runKubectlCommandInTerminal("kubectl-expose", cmd);
```
itowlson: Seem to be using slightly random terminals depending on which command is run and whether it internally goes through the 'active window', 'temp file' or other code path.  All that stuff should be hidden from the user.  If you want commands to share a terminal then maybe have _all_ commands share that terminal; if you want each command to start its own terminal then give them more specific names.  (I think the latter could result in a _lot_ of terminals though.)

src/extension.ts
```diff
@@ -1353,12 +1350,7 @@ async function useContextKubernetes(explorerNode: explorer.KubernetesObject) {
 
 async function clusterInfoKubernetes(explorerNode: explorer.KubernetesObject) {
     const targetContext = explorerNode.metadata.context;
-    const shellResult = await kubectl.invokeAsync(`cluster-info`);
-    if (shellResult.code === 0) {
-        kubeChannel.showOutput(shellResult.stdout, `cluster-info for ${explorerNode.id}`);
-    } else {
-        vscode.window.showErrorMessage(`Failed to get cluster info: ${shellResult.stderr}`);
-    }
+    runKubectlCommandInTerminal(`cluster-info for ${targetContext}`, "cluster-info");
```
itowlson: This terminal name has a completely different pattern from other names.

src/extension.ts
testforstephen: I prefer to add a new method `invokeInTerminal` in `kubectl` interface.

src/extension.ts
testforstephen: Nice. I'd like the shared terminal suggestion. 

Besides, i have removed the terminal command for 'temp file', but for normal `apply file`, i'd like to show it in terminal. 

And for other ui operations (get/delete/describe/logs/expose/exec/scale/cluster-info), if it is be done by one single kubectl command, then i'd like to run it in terminal. Otherwise, hide the detail by using the normal `kubectl.invoke`.

src/extension.ts
testforstephen: I prefer to let all commands to share the terminal.

testforstephen:
@itowlson @kyliel

> When is a command meant to be run in a terminal?

For ui operations, if it is be done by one single kubectl command, then i'd like to run it in terminal. 
The typical cases are `create/get/delete/describe/logs/expose/exec/scale/cluster-info`

Otherwise, hide the detail by using the normal kubectl.invoke.
The typical cases are `run/sync/diff/debug`

> How many terminals do we create? If there are multiple terminals then how does the user tell which is which?

I'd like to share the terminal. Although `vscode-docker` extension always run command in a new terminal, but i feel like the shared terminal is more friendly because k8s extension provides more ui commands (> 20) than docker.

testforstephen:
@itowlson Per last meeting discussed, we changed `kubectl` command to TERMINAL first, and revisit how to keep the command history later. Could your help have a look to see if this PR is OK?

src/extension.ts
```diff
@@ -946,19 +944,14 @@ function getPorts() {
 
 function describeKubernetes(explorerNode? : explorer.ResourceNode) {
     if (explorerNode) {
-        describeKubernetesCore(explorerNode.resourceId);
+        kubectl.invokeInTerminal(`describe ${explorerNode.resourceId}`);
```
itowlson: This will no longer display a separate progress indicator - I think that's okay but just checking that this is your intent.

src/extension.ts
```diff
@@ -1075,32 +1067,21 @@ async function refreshExplorer() {
...
-
 const deleteKubernetes = async (explorerNode? : explorer.ResourceNode) => {
     if (explorerNode) {
         const answer = await vscode.window.showWarningMessage(`Do you want to delete the resource '${explorerNode.resourceId}'?`, ...deleteMessageItems);
         if (answer.isCloseAffordance) {
             return;
         }
-        const shellResult = await kubectl.invokeAsyncWithProgress(`delete ${explorerNode.resourceId}`, `Deleting ${explorerNode.resourceId}...`);
-        await reportDeleteResult(explorerNode.resourceId, shellResult);
+        kubectl.invokeInTerminal(`delete ${explorerNode.resourceId}`);
```
itowlson: Hmm, this is a case where I don't much like the terminal experience.  It requires the user to pay attention to a noisy text window to see the outcome, whereas the old way (info message for success and error message for failure) gave much more obvious feedback.  (Scale would probably be another example of this.)  It's fine to point the user to the terminal window output if they want to grovel through the error details, but it would be good to surface the success/failure indicators more simply.  `kubectl` is an implementation detail, which it's useful to surface but shouldn't be in the user's face all the time.

src/kubectl.ts
```diff
@@ -27,6 +33,7 @@ class KubectlImpl implements Kubectl {
     }
 
     private readonly context : Context;
+    private terminal : any;
```
itowlson: Do we not have better type info than `any`?

src/kubectl.ts
```diff
@@ -43,12 +50,28 @@ class KubectlImpl implements Kubectl {
...
+        return invokeInTerminal(this.context, command, terminal);
+    }
     asLines(command : string) : Promise<string[] | ShellResult> {
         return asLines(this.context, command);
     }
     path() : string {
         return path(this.context);
     }
+    private getSharedTerminal() : any {
+        if (!this.terminal) {
```
itowlson: Would it be clearer to call this `sharedTerminal` or something like that?

src/extension.ts
testforstephen: yes, this is by purpose. In terminal, user could get the current status of the command by himself, so no need to show an additional progress indicator.

src/extension.ts
testforstephen: yeah, i saw the ux agreement in yesterday's meeting, have reverted to the old design.

testforstephen:
@itowlson PR updated, Any other comments?

src/extension.ts
```diff
@@ -421,14 +421,14 @@ function maybeRunKubernetesCommandForActiveWindow(command) {
                         vscode.window.showErrorMessage("Save failed.");
                         return;
                     }
-                    kubectl.invoke(`${command} "${editor.document.fileName}"`);
+                    kubectl.invokeInTerminal(`${command} "${editor.document.fileName}"`);
```
itowlson: This looks like it will affect Create, Apply and Diff as well as Get?  Per Kylie's notes, I think the former three are meant to _not_ go via the terminal, though Get is.

src/extension.ts
```diff
@@ -556,22 +556,21 @@ function exposeKubernetes() {
         cmd += ' --port=' + ports[0];
     }
 
-    kubectl.invoke(cmd);
+    kubectl.invokeInTerminal(cmd);
```
itowlson: Expose should not go via terminal, per Kylie's notes.

src/extension.ts
testforstephen: Fixed. Already undo the changes.

src/extension.ts
testforstephen: Fixed. Already undo the changes.

PR #52, opened 2018-02-08
Adding support for kubectl dashboard

### LINKED ISSUE #9: Support kubectl proxy
Would be great to see support for the Kubernetes dashboard via`kubectl proxy`. I think the idealized behavior here would be:

1. Invoke `kubectl proxy`
2. Either open the user's default browser, or do an HTML preview pane with and navigate to the dashboard address: `http://localhost:8001/api/v1/namespaces/kube-system/services/kubernetes-dashboard/proxy/`
3. If we open the preview pane, see if we can hook into a close event on the pane to shut down the proxy after the user is done with it.

Thoughts?

bnookala:
See below for demo:

kubeproxy

Runs `kubectl proxy` in a terminal window created by the extension, and then attempts to open the Kubernetes dashboard in the user's browser. 

Fixes #9

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
         "onCommand:extension.vsKubernetesRemoveDebug",
         "onCommand:extension.vsKubernetesConfigureFromCluster",
         "onCommand:extension.vsKubernetesCreateCluster",
+        "onCommand:extension.vsKubernetesDashboard",
         "onCommand:extension.helmTemplate",
         "onCommand:extension.helmTemplatePreview",
         "onCommand:extension.helmLint",
@@ -350,6 +351,11 @@
                 "category": "Kubernetes"
             },
             {
+                "command": "extension.vsKubernetesDashboard",
+                "title": "Open Dashboard",
+                "category": "Kubernetes"
+            },
+            {
                 "command": "extension.draftVersion",
                 "title": "Version",
                 "description": "Get the version of the local Draft client.",
@@ -477,17 +483,18 @@
         "test": "node ./node_modules/vscode/bin/test"
     },
     "dependencies": {
-        "shelljs": "^0.7.7",
-        "js-yaml": "^3.8.2",
-        "dockerfile-parse": "^0.2.0",
-        "k8s": "^0.4.12",
-        "tmp": "^0.0.31",
-        "pluralize": "^4.0.0",
-        "uuid": "^3.1.0",
-        "yamljs": "0.2.10",
         "compare-versions": "^3.1.0",
+        "dockerfile-parse": "^0.2.0",
+        "js-yaml": "^3.8.2",
+        "k8s": "^0.4.12",
         "lodash": ">3",
-        "vscode-extension-telemetry": "^0.0.6"
+        "opn": "^5.2.0",
+        "pluralize": "^4.0.0",
+        "shelljs": "^0.7.7",
+        "tmp": "^0.0.31",
+        "uuid": "^3.1.0",
+        "vscode-extension-telemetry": "^0.0.6",
+        "yamljs": "0.2.10"
     },
     "devDependencies": {
         "@types/mocha": "^2.2.32",
--- a/src/components/kubectl/proxy.ts
+++ b/src/components/kubectl/proxy.ts
@@ -1,0 +1,95 @@
+'use strict';
+
+import * as vscode from 'vscode';
+import * as opn from 'opn';
+
+import {createReadStream} from 'fs';
+import {resolve} from 'path';
+import {fs} from '../../fs';
+
+const KUBE_DASHBOARD_URL = "http://localhost:8001/ui/";
+const TERMINAL_NAME = "Kubernetes Proxy";
+const PROXY_OUTPUT_FILE = resolve(__dirname, 'proxy.out');
+
+// The instance of the terminal running Kubectl Proxy.
+let terminal:vscode.Terminal;
+
+/**
+ * Runs `kubectl proxy` in a terminal process spawned by the extension, and opens the Kubernetes
+ * dashboard in the user's preferred browser.
+ */
+export async function dashboardKubernetes ():Promise<void> {
+    // If we've already got a terminal instance, just open the dashboard.
+    if (terminal) {
+        opn(KUBE_DASHBOARD_URL);
+        return;
+    }
+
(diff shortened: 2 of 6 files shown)


squillace:
oooo me like. Nice fix! we'll get to the az aks browse in another extension.

bnookala:
@squillace I'll open a ticket for that!

bnookala:
@squillace oOoo - are we doing a whole separate AKS extension?

src/components/kubectl/proxy.ts
```diff
@@ -0,0 +1,139 @@
...
+    // Maybe we've bound to the port already outside of the extension?
+    vscode.window.showErrorMessage("Could not start Kubernetes API Proxy. Is it already running?");
+    terminal.dispose();
+};
+
+/**
+ * Promisified version of node's `fs.unlink`
+ * @param path
+ */
+const promisedUnlink = (path:string):Promise<null> => {
```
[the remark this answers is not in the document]
itowlson: For most of these we could wickedly just use the sync versions as they should be quick enough to not interrupt the UI... we already do this in a lot of places!

src/components/kubectl/proxy.ts
```diff
@@ -0,0 +1,139 @@
...
+const PROXY_OUTPUT_FILE = resolve(__dirname, 'proxy.out');
+
+// The instance of the terminal running Kubectl Proxy.
+let terminal:vscode.Terminal;
+
+/**
+ * Runs `kubectl proxy` in a terminal process spawned by the extension, and opens the Kubernetes
+ * dashboard in the user's preferred browser.
+ */
+export async function proxyKubernetes () {
```
itowlson: return type is `Promise<void>`?

src/components/kubectl/proxy.ts
```diff
@@ -0,0 +1,139 @@
...
+    // to listen for when the server is ready.
+    terminal.sendText(`kubectl proxy | tee ${PROXY_OUTPUT_FILE}`, true);
+    terminal.show();
+}
+
+/**
+ * Called when the terminal window is closed by the user.
+ * @param proxyTerminal
+ */
+const onClosedTerminal = async (proxyTerminal:vscode.Terminal) => {
```
itowlson: The extension has traditionally been inconsistent about this but is there a particular reason for preferring `const foo = (bar) => { ... };` to `function foo(bar) { ... }`?

package.json
```diff
@@ -349,6 +350,11 @@
                 "title": "Set as Current Namespace",
                 "category": "Kubernetes"
             },
+            {
+                "command": "extension.vsKubernetesProxy",
+                "title": "Start Kubernetes Proxy, open the Kubernetes Dashboard",
```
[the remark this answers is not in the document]
bnookala: Oh good point! I forgot about the category convention here. As far as as Dashboard vs. Proxy… we can implement both commands ie 'Open Dashboard' and 'Start API Proxy' but point them to the same activating function, and only open the dashboard when running the the dashboard command - thoughts?

src/components/kubectl/proxy.ts
bnookala: I (slightly) prefer arrow syntax, as it appears cleaner to me. The use of `const` makes the function immutable, which I think is good style.

src/components/kubectl/proxy.ts
```diff
@@ -0,0 +1,139 @@
...
+        // Let the proxy warm up a bit… otherwise we might hit a browser's error page.
+        setTimeout(() => {
+            opn(KUBE_DASHBOARD_URL);
+        }, 2500);
+        return;
+    }
+
+    // Maybe we've bound to the port already outside of the extension?
+    vscode.window.showErrorMessage("Could not start Kubernetes API Proxy. Is it already running?");
+    terminal.dispose();
```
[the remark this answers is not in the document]
bnookala: Looks like it does… I'll remove that line. https://github.com/Microsoft/vscode/issues/10925#issuecomment-247428328

package.json
```diff
@@ -349,6 +350,11 @@
                 "title": "Set as Current Namespace",
                 "category": "Kubernetes"
             },
+            {
+                "command": "extension.vsKubernetesDashboard",
+                "title": "Open the Dashboard",
```
itowlson: super picky: just `Dashboard` or `Open Dashboard`?

src/components/kubectl/proxy.ts
```diff
@@ -0,0 +1,95 @@
...
+        PROXY_OUTPUT_FILE,
+        {encoding: 'utf8'}
+    ).on('data', onStreamData);
+
+    terminal = vscode.window.createTerminal(TERMINAL_NAME);
+    vscode.window.onDidCloseTerminal(onClosedTerminal);
+
+    // stdout is also written to a file via `tee`. We read this file as a stream
+    // to listen for when the server is ready.
+    terminal.sendText(`kubectl proxy | tee ${PROXY_OUTPUT_FILE}`, true);
```
itowlson: TIL `tee` works in Powershell - neat!

src/components/kubectl/proxy.ts
```diff
@@ -0,0 +1,95 @@
...
+    terminal.show();
```
itowlson: Is it worth passing `true` here so that the terminal doesn't steal focus?  It is kind of moot since we're about to open an enormous browser window which will _steal the heck_ out of the focus... but when the user Alt+Tabs back to VS Code, it might be good to have focus still on the editor?

bnookala: I think this is required to make the terminal run the command: https://code.visualstudio.com/docs/extensionAPI/vscode-api#Terminal, otherwise it waits for the user to do it themselves.

package.json
bnookala: let's go with the second one here.

src/components/kubectl/proxy.ts
itowlson: I tried a trivial test and the command still seemed to be executed if you passed `true` for `preserveFocus`, at least on Windows.  Do you see something different?  If so, then I agree we can't pass the option.

bnookala: er derp; i thought this comment was with regards to the sendText call. I tried the change and it seems to work fine :)

PR #53, opened 2018-02-08
utility functions for integrating kubernetes yaml support feature.

andxu:
https://github.com/Azure/vscode-kubernetes-tools/issues/42

### DIFF
--- a/src/yaml-languange-support/schema-utils.ts
+++ b/src/yaml-languange-support/schema-utils.ts
@@ -1,0 +1,37 @@
+'use strict';
+
+/**
+ * An utility function for splitting text into lines.
+ */
+export function splitLines(text) {
+    return text.split(/\r\n?|\n/);
+} 
+
+/**
+ * A very easy utility function for decides whether the text is support to be a kubernetes yaml file, further
+ * implementations is needed to be more precise on the result rather than the regex test.
+ */
+export function isKubernetesYamlFile(text) {
+    return !!(/^kind\s*\:/g.exec(text) ||
+        /[^a-zA-Z]kind\s*\:/g.exec(text) ||
+        /^apiVersion\s*\:/g.exec(text) ||
+        /[^a-zA-Z]apiVersion\s*\:/g.exec(text)
+    )
+}
+
+/**
+ * An utility function for get apiVersion and kind from 'x-kubernetes-group-version-kind' node in kubernetes schema.
+ */
+export function parseApiVersionKind(groupKindNode) {
+    if (!groupKindNode.version) {
+        throw new Error("Invalid 'x-kubernetes-group-version-kind' node, 'version' is missing.");
+    }
+
+    if (!groupKindNode.kind) {
+        throw new Error("Invalid 'x-kubernetes-group-version-kind' node, 'kind' is missing.");
+    }
+    return {
+        apiVersion: (groupKindNode.group ? groupKindNode.group + '/' : '') + groupKindNode.version, 
+        kind: groupKindNode.kind
+    };
+}
--- a/src/yaml-languange-support/schema-utils.ts
+++ b/src/yaml-languange-support/schema-utils.ts
@@ -16,7 +16,7 @@ export function isKubernetesYamlFile(text) {
         /[^a-zA-Z]kind\s*\:/g.exec(text) ||
         /^apiVersion\s*\:/g.exec(text) ||
         /[^a-zA-Z]apiVersion\s*\:/g.exec(text)
-    )
+    );
 }
 
 /**


src/yaml-languange-support/schema-utils.ts
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
itowlson: typo in directory path `yaml-langua>n<ge...` -> `yaml-language...`

src/yaml-languange-support/schema-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * An utility function for splitting text into lines.
+ */
+export function splitLines(text) {
```
itowlson: All functions should have parameter and return types please.

src/yaml-languange-support/schema-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    return text.split(/\r\n?|\n/);
+} 
+
+/**
+ * A very easy utility function for decides whether the text is support to be a kubernetes yaml file, further
+ * implementations is needed to be more precise on the result rather than the regex test.
+ */
+export function isKubernetesYamlFile(text) {
+    return !!(/^kind\s*\:/g.exec(text) ||
+        /[^a-zA-Z]kind\s*\:/g.exec(text) ||
```
itowlson: What is the intent of the `[^a-zA-Z]` for here?  If the intention is to allow `  kind:` (indented) or `- kind:` or something like that then can we be more specific in the regex?

src/yaml-languange-support/schema-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+        /^apiVersion\s*\:/g.exec(text) ||
+        /[^a-zA-Z]apiVersion\s*\:/g.exec(text)
+    );
+}
+
+/**
+ * An utility function for get apiVersion and kind from 'x-kubernetes-group-version-kind' node in kubernetes schema.
+ */
+export function parseApiVersionKind(groupKindNode) {
```
itowlson: I'm not sure what a `groupKindNode` is or where we get one from.  I take it this is an `any`, perhaps coming from a JSON or YAML load?  And we are extracting the interesting values into a well-typed structure?

src/yaml-languange-support/schema-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (!groupKindNode.version) {
+        throw new Error("Invalid 'x-kubernetes-group-version-kind' node, 'version' is missing.");
```
itowlson: What are we planning to do with these errors?  The messages are extremely technical - few users will know what an `x-kubernetes-group-version-kind` is - so we can't display them...

src/yaml-languange-support/schema-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    if (!groupKindNode.kind) {
+        throw new Error("Invalid 'x-kubernetes-group-version-kind' node, 'kind' is missing.");
+    }
+    return {
+        apiVersion: (groupKindNode.group ? groupKindNode.group + '/' : '') + groupKindNode.version,
```
itowlson: How is this stringised value used?  Is this `group/version` representation something that the Kubernetes API/tooling recognises or will we need to split it out again?

itowlson:
I noticed this PR is from a branch in the main repo.  The workflow we've been using it to fork the repo into personal GitHub accounts, create branches in those accounts, and create PRs from there.  This keeps the Azure repo free of WIP branches, and puts us on the same workflow as community contributors (most people do not have push access to the Azure org repo).  Could you adopt that workflow for future PRs if it works for you please?  Happy to discuss a better workflow if people don't like the current one though!

src/yaml-languange-support/schema-utils.ts
andxu: I would like to match `kind:` and not `akind:`, I have updated the regex in newer PR.

src/yaml-languange-support/schema-utils.ts
andxu: It is from kubernetes schema, see: https://github.com/garethr/openshift-json-schema/blob/master/v3.6.0/deployment.json#L2

src/yaml-languange-support/schema-utils.ts
andxu: I updated the map in new PR's code comment

andxu:
I will

andxu:
Close this one as it is directly from the repo: https://github.com/Azure/vscode-kubernetes-tools, will open a new PR based on my personal repo.

PR #55, opened 2018-02-09
utility functions for integrating kubernetes yaml support feature, 

andxu:
Fix as code reviewer suggests.

### DIFF
--- a/src/yaml-language-support/schema-models.ts
+++ b/src/yaml-language-support/schema-models.ts
@@ -1,0 +1,34 @@
+/**
+ * The model for 'x-kubernetes-group-version-kind' node in kubernetes schema, a sample schema would look like:
+ * <pre><code>
+ *   "x-kubernetes-group-version-kind": [
+ *   {
+ *       "group": "apps",
+ *       "kind": "Deployment",
+ *       "version": "v1beta1"
+ *   }]
+ * </code></pre>
+ *
+ * Above node specifies the schema to match the following yaml:
+ * <pre><code>
+ * apiVersion: extensions/v1beta1
+ * kind: Deployment
+ * metadata:
+ *  name: my-app
+ * spec:
+ *   replicas: 2
+ * </code></pre>
+ */
+export interface KubernetesGroupKindNode {
+    group?: string;
+    version: string;
+    kind: string;
+}
+
+/**
+ * The base model for all kubernetes manifest objects, eg: Pod, Service, Deployment.
+ */
+export interface KubernetesResourceObjectBase {
+    apiVersion: string;
+    kind: string;
+}
--- a/src/yaml-language-support/schema-utils.ts
+++ b/src/yaml-language-support/schema-utils.ts
@@ -1,0 +1,41 @@
+'use strict';
+
+import { KubernetesGroupKindNode, KubernetesResourceObjectBase } from "./schema-models";
+
+/**
+ * An utility function for splitting text into lines.
+ */
+export function splitLines(text: string): string[] {
+    return text.split(/\r\n?|\n/);
+} 
+
+/**
+ * A very easy utility function for decides whether the text is support to be a kubernetes yaml file, further
+ * implementations is needed to be more precise on the result rather than the regex test.
+ */
+export function isKubernetesYamlFile(text: string): boolean {
+    return !!(/\bkind\s*:/g.exec(text) ||
+        /\bapiVersion\s*:/g.exec(text));
+}
+
+/**
+ * An utility function for get apiVersion and kind from 'x-kubernetes-group-version-kind' node in kubernetes schema.
+ */
+export function parseApiVersionKind(groupKindNode: KubernetesGroupKindNode): KubernetesResourceObjectBase {
+    if (!groupKindNode.version) {
+        throw new Error("Encounter invalid kubernetes schema with no 'version' information.");
+    }
+
+    if (!groupKindNode.kind) {
+        throw new Error("Encounter invalid kubernetes schema with no 'kind' information");
+    }
+
+    // apiVersion := [%group%/] ? %version% 
+    // eg:  {"group": "apps", "kind": "Deployment","version": "v1beta1" } will map to:
+    // apiVersion: extensions/v1beta1, kind: Deployment
+
+    return {
+        apiVersion: (groupKindNode.group ? groupKindNode.group + '/' : '') + groupKindNode.version, 
+        kind: groupKindNode.kind
+    };
+}


andxu:
https://github.com/Azure/vscode-kubernetes-tools/issues/42

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+        throw new Error("Encounter invalid kubernetes schema with no 'version' information.");
+    }
+
+    if (!groupKindNode.kind) {
+        throw new Error("Encounter invalid kubernetes schema with no 'kind' information");
+    }
+
+    // apiVersion := [%group%/] ? %version% 
+    // eg:  {"group": "apps", "kind": "Deployment","version": "v1beta1" } will map to:
+    // apiVersion: extensions/v1beta1, kind: Deployment
```
itowlson: The code doesn't do this though.  With that GroupKindNode, the apiVersion would be `apps/v1beta`, not `extensions/v1beta1`, wouldn't it?

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    return {
+        apiVersion: (groupKindNode.group ? groupKindNode.group + '/' : '') + groupKindNode.version,
```
itowlson: Precedence for the ternary operator isn't well known, so consider using brackets to make this more explicit: `(groupKind.group ? (groupKind.group + '/') : '')`

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+    return text.split(/\r\n?|\n/);
+} 
+
+/**
+ * A very easy utility function for decides whether the text is support to be a kubernetes yaml file, further
+ * implementations is needed to be more precise on the result rather than the regex test.
+ */
+export function isKubernetesYamlFile(text: string): boolean {
+    return !!(/\bkind\s*:/g.exec(text) ||
+        /\bapiVersion\s*:/g.exec(text));
```
itowlson: If I read this right it will match any text that includes the string `kind:`.  Should we consider making this an `&&` instead of an `||` check?

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/**
+ * An utility function for get apiVersion and kind from 'x-kubernetes-group-version-kind' node in kubernetes schema.
+ */
+export function parseApiVersionKind(groupKindNode: KubernetesGroupKindNode): KubernetesResourceObjectBase {
```
itowlson: The naming seems a bit odd here - returning `KubernetesResourceObjectBase` makes me think it's returning the resource object through a polymorphic interface - but all it's actually doing is returning the apiVersion/kind record.  So it's not clear to me whether `KubernetesResourceObjectBase` is meant to be a base type for k8s objects (in which case this function is returning the wrong thing), or a record type for apiVersion/kind (in which case its name is wrong).  Perhaps this will be clearer once we see it being used but it's hard to know looking at this code in isolation.

src/yaml-language-support/schema-models.ts
```diff
@@ -0,0 +1,34 @@
...
+ * <pre><code>
+ * apiVersion: extensions/v1beta1
+ * kind: Deployment
+ * metadata:
+ *  name: my-app
+ * spec:
+ *   replicas: 2
+ * </code></pre>
+ */
+export interface KubernetesGroupKindNode {
```
itowlson: The members should probably be `readonly`.

src/yaml-language-support/schema-models.ts
```diff
@@ -0,0 +1,34 @@
...
+    group?: string;
+    version: string;
+    kind: string;
+}
+
+/**
+ * The base model for all kubernetes manifest objects, eg: Pod, Service, Deployment.
+ */
+export interface KubernetesResourceObjectBase {
```
itowlson: Again should probably be `readonly`?

itowlson: Not sure we need `Object` here - wouldn't `KubernetesResourceBase` work just as well?  Or are you specifically trying to distinguish a manifest or descriptor from the resource itself?  In which case maybe prefer `Manifest` or something to the generic `Object`?

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An utility function for splitting text into lines.
+ */
+export function splitLines(text: string): string[] {
+    return text.split(/\r\n?|\n/);
+} 
+
+/**
+ * A very easy utility function for decides whether the text is support to be a kubernetes yaml file, further
```
itowlson: Comment reads a bit oddly - suggest e.g. "A simple function for deciding whether a piece of text is a Kubernetes YAML file."

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+ * implementations is needed to be more precise on the result rather than the regex test.
+ */
+export function isKubernetesYamlFile(text: string): boolean {
+    return !!(/\bkind\s*:/g.exec(text) ||
+        /\bapiVersion\s*:/g.exec(text));
+}
+
+/**
+ * An utility function for get apiVersion and kind from 'x-kubernetes-group-version-kind' node in kubernetes schema.
```
itowlson: I don't think the documentation needs to state that it's a utility function - we know it's a function from the keyword, and that it's a utility from the module name *grin*.  Consider just "Gets apiVersion and kind from..."

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+export function parseApiVersionKind(groupKindNode: KubernetesGroupKindNode): KubernetesResourceObjectBase {
+    if (!groupKindNode.version) {
+        throw new Error("Encounter invalid kubernetes schema with no 'version' information.");
+    }
+
+    if (!groupKindNode.kind) {
+        throw new Error("Encounter invalid kubernetes schema with no 'kind' information");
+    }
+
+    // apiVersion := [%group%/] ? %version%
```
itowlson: What's this syntax?

src/yaml-language-support/schema-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+    return !!(/\bkind\s*:/g.exec(text) ||
+        /\bapiVersion\s*:/g.exec(text));
+}
+
+/**
+ * An utility function for get apiVersion and kind from 'x-kubernetes-group-version-kind' node in kubernetes schema.
+ */
+export function parseApiVersionKind(groupKindNode: KubernetesGroupKindNode): KubernetesResourceObjectBase {
+    if (!groupKindNode.version) {
+        throw new Error("Encounter invalid kubernetes schema with no 'version' information.");
```
itowlson: "Encounter" reads like a command here.  Consider just describing the error condition: "Invalid Kubernetes schema: no 'version' attribute."

andxu:
Close since another pr deprecates this one.

PR #61, opened 2018-02-23
Fixes for Apply command

### LINKED ISSUE #60: "Kubernetes: Apply" does not do anything
I create new test deployment file like:

```
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: test
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
        - command:
            - sh
            - -c
            - while true; do sleep 86400; done
          image: busybox
          name: test
      terminationGracePeriodSeconds: 1
```
Next I run "Kubernetes: Apply" command and nothing is changed...
I expect deployment creation after this command or even error message.
I can manage my cluster (get logs, deployments, pods, etc.) through this extension so I think config is OK and ```kubectl``` is reachable.
Also I have try to open existing deployment, change it and run same command - same effect, nothing is changed.
Maybe I do something wrong?

itowlson:
This fixes a couple of glitches in the Apply command:

1. If the user had not selected any text, it silently exited.  It now applies the whole content of the active document.

2. If the resource did not exist, it exited with a 'resource not found' error.  This is inconsistent with `kubectl apply` which will create the resource if it doesn't exist.  It now warns that this will create a new resource instead of modifying an existing one.

3. If the resource existed, it displayed a diff of the active document against a JSON representation of the resource.  If the active document was YAML, it was hard to spot the differences.  It now detects the type of the active document, and diffs against the appropriate representation.

Fixes #60.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1147,7 +1147,7 @@ const diffKubernetes = (callback) => {
             fileName = file;
             if (vscode.window.activeTextEditor && vscode.window.activeTextEditor.document) {
                 const langId = vscode.window.activeTextEditor.document.languageId.toLowerCase();
-                if (langId == "yaml" || langId == "helm") {
+                if (langId === "yaml" || langId === "helm") {
                     fileFormat = "yaml";
                 }
             }
@@ -1164,7 +1164,7 @@ const diffKubernetes = (callback) => {
         kubectl.invoke(` get -o ${fileFormat} ${kindName}`, (result, stdout, stderr) => {
             if (result == 1 && stderr.indexOf('NotFound') >= 0) {
                 vscode.window.showWarningMessage(`Resource ${kindName} does not exist - this will create a new resource.`, 'Create').then((choice) => {
-                    if (choice == 'Create') {
+                    if (choice === 'Create') {
                         maybeRunKubernetesCommandForActiveWindow('create -f');
                     }
                 });
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -462,12 +462,10 @@ function getTextForActiveWindow(callback) {
     if (editor.selection) {
         text = editor.document.getText(editor.selection);
 
-        if (text.length === 0) {
+        if (text.length > 0) {
+            callback(text, null);
             return;
         }
-
-        callback(text, null);
-        return;
     }
 
     if (editor.document.isUntitled) {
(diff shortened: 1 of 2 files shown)


src/shell.ts
```diff
@@ -55,6 +57,13 @@ function combinePath(basePath : string, relativePath : string) {
     return basePath + separator + relativePath;
 }
 
+function fileUri(filePath : string) : vscode.Uri {
+    if (isWindows()) {
+        return vscode.Uri.parse('file:///' + filePath.replace(/\\/g, '/'));
```
bnookala: perhaps we can use node's path module to make this work?

src/shell.ts
```diff
@@ -55,6 +57,13 @@ function combinePath(basePath : string, relativePath : string) {
...
+    }
+    return vscode.Uri.parse('file://' + filePath);
```
bnookala: likewise

src/extension.ts
```diff
@@ -1154,19 +1161,27 @@ const diffKubernetes = (callback) => {
             return;
         }
 
-        kubectl.invoke(` get -o json ${kindName}`, (result, stdout, stderr) => {
-            if (result !== 0) {
+        kubectl.invoke(` get -o ${fileFormat} ${kindName}`, (result, stdout, stderr) => {
+            if (result == 1 && stderr.indexOf('NotFound') >= 0) {
+                vscode.window.showWarningMessage(`Resource ${kindName} does not exist - this will create a new resource.`, 'Create').then((choice) => {
+                    if (choice == 'Create') {
```
bnookala: Use ===

also… is the choice necessary if there's only a single option int he warning message (`create`)?

src/extension.ts
```diff
@@ -1137,13 +1135,22 @@ const diffKubernetes = (callback) => {
...
             kindName = findKindNameForText(data);
-            fileName = path.join(os.tmpdir(), 'local.json');
+            fileName = path.join(os.tmpdir(), `local.${fileFormat}`);
             fs.writeFile(fileName, data, handleError);
         } else if (file) {
             kindName = findKindName();
             fileName = file;
+            if (vscode.window.activeTextEditor && vscode.window.activeTextEditor.document) {
+                const langId = vscode.window.activeTextEditor.document.languageId.toLowerCase();
+                if (langId == "yaml" || langId == "helm") {
```
bnookala: === for equality

src/extension.ts
itowlson: The choice is necesssary because there's a 'Close' option automatically supplied by VS Code (which we treat as cancellation).

src/shell.ts
itowlson: Hmm, I don't see anything in the path module that helps with the `file:` side of things -- am I missing something?  We could probably use it to avoid the backslash replacement but I'm not sure there's much value in that -- what do you reckon?

PR #62, opened 2018-02-24
add yaml schema support 

andxu:
1. fix the TODO item at helm.hoverProvider.ts: right now, the line `foo: {{foo}}` may match both the action and the resource def
2. add yaml parser dependency


src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { parse, findNodeAtPosition, util as yamlUtil } from 'node-yaml-parser';
+
+export interface YamlCache {
+    // the document model list represents the yaml text
+    yamlDocs: any[];
+
+    // lineLens contains the converted line length of each lines, it is used for converting from
+    // vscode position to the inner position in yaml element model.
+    lineLens: number[];
```
itowlson: I read this as 'lens' (as in CodeLens).  `lineLengths` would be clearer.

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
+import * as vscode from 'vscode';
+
+import { parse, findNodeAtPosition, util as yamlUtil } from 'node-yaml-parser';
+
+export interface YamlCache {
+    // the document model list represents the yaml text
```
itowlson: I don't understand this comment.  Does this represent a single YAML document, or a list of YAML documents?  What is meant by 'the document model list'?

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+    yamlDocs: any[];
+
+    // lineLens contains the converted line length of each lines, it is used for converting from
+    // vscode position to the inner position in yaml element model.
+    lineLens: number[];
+
+    // the version of the document to avoid duplicate work on the same text
```
itowlson: "the document" implies a YamlCache represents a document, not a cache.  Should we call it something like YamlDocument or YamlCachedDocument or YamlParsedDocument or something like that?

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+    version: number;
+}
+
+export interface YamlMatchedElement {
+    // the document model list represents the yaml text
+    readonly yamlDocs: any[];
+
+    // the found node at the given position(usually at the edit/hover place)
```
itowlson: 'the given position' lacks context here - nothing else in the interface talks about positions, and the name talks about matching rather than positions.  I'd suggest something like 'the node which was found as the successful match' perhaps?  Or if it is about positions then maybe rename the interface and call this e.g. `nodeAtPosition` or even just `node`?

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+    // vscode position to the inner position in yaml element model.
+    lineLens: number[];
+
+    // the version of the document to avoid duplicate work on the same text
+    version: number;
+}
+
+export interface YamlMatchedElement {
+    // the document model list represents the yaml text
+    readonly yamlDocs: any[];
```
itowlson: I don't understand how an array of documents fits into a "matched element" interface.

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+    // the document which contains the node at given position
+    readonly matchedDocument: any;
+}
+
+/**
+ * A yaml interpreter parse the yaml text and find the matched ast node from vscode location.
+ */
+ class YamlLocator {
+    
+    private _cache:{ [key:string]: YamlCache; }  = {};
```
itowlson: Perhaps document that this is mapping URIs to cached documents?

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+        if (!this._cache[key]) {
+            this._cache[key] = <YamlCache> {version: -1};
+        }
+
+        if (this._cache[key].version !== textDocument.version) {
+            const { documents, lineLens } = parse(textDocument.getText());
+            this._cache[key].yamlDocs = documents;
+            this._cache[key].lineLens = lineLens;
+            this._cache[key].version = textDocument.version;
+        }
```
itowlson: Consider extracting the first part of this method into a separate method whose name documents what it does, e.g. `ensureCached(...)` (?).

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+        };
+    }
+
+ }
+
+ // a global instance of yaml locator
+const yamlLocator = new YamlLocator();
+
+ // yamlUtil contains the important util functions for handling with yaml elements
+export { yamlLocator, yamlUtil };
```
itowlson: Why does this module need to re-export `yamlUtil`?  Can't consumers import it directly from `node-yaml-parser`?  Are there more specific wrappers we can provide instead?

src/yaml-language-support/yaml-support.ts
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public getYamlElement(textDocument: vscode.TextDocument, pos: vscode.Position): YamlMatchedElement {
+        const key: string = textDocument.uri.toString();
+        if (!this._cache[key]) {
+            this._cache[key] = <YamlCache> {version: -1};
+        }
+
+        if (this._cache[key].version !== textDocument.version) {
+            const { documents, lineLens } = parse(textDocument.getText());
+            this._cache[key].yamlDocs = documents;
```
itowlson: Okay, so the idea here is that a single file may contain multiple YAML documents, and that's why there's the `yamlDocs: any[]` member?  That makes sense, but I think we need to make the documentation clearer.

src/yaml-language-support/yaml-support.ts
itowlson: Elsewhere we generally do `export class`, `export interface` etc. rather than having a separate `export` statement.

src/helm.hoverProvider.ts
```diff
@@ -40,6 +42,15 @@ export class HelmTemplateHoverProvider implements vscode.HoverProvider {
...
 
         if (this.notInAction(doc, pos, word)) {
+            try {
+                // have a test against whether the word is a key element in yaml
+                const { matchedNode } = yamlLocator.getYamlElement(doc, pos);
+                if (!yamlUtil.isKey(matchedNode)) {
+                    return Promise.resolve(null);
+                }
+            } catch (ex) {
+                console.log(ex);
```
itowlson: Is this desired behaviour?  As far as the user is concerned, this results in silently swallowing the exception.

src/helm.hoverProvider.ts
```diff
@@ -40,6 +42,15 @@ export class HelmTemplateHoverProvider implements vscode.HoverProvider {
         }
 
         if (this.notInAction(doc, pos, word)) {
+            try {
+                // have a test against whether the word is a key element in yaml
```
itowlson: Can we document why we care about this, rather than just what check we're performing?

src/helm.hoverProvider.ts
```diff
@@ -40,6 +42,15 @@ export class HelmTemplateHoverProvider implements vscode.HoverProvider {
...
+                const { matchedNode } = yamlLocator.getYamlElement(doc, pos);
+                if (!yamlUtil.isKey(matchedNode)) {
+                    return Promise.resolve(null);
+                }
+            } catch (ex) {
+                console.log(ex);
+            }
```
itowlson: You mention this is resolving a TODO comment but the diff does not seem to show any TODO comment being removed - am I missing something?

src/helm.hoverProvider.ts
itowlson: Hmm, it's a hover provider, so yeah, silence is probably best... but in that case we should be avoiding having exceptions at all.  Should we perhaps wrap this in a meaningfully named helper method (e.g. `isPositionInKey`) that internally suppresses exceptions if they are truly going to happen?

src/yaml-language-support/yaml-support.ts
andxu: sure, I will change it to lineLengths later

src/yaml-language-support/yaml-support.ts
andxu: a yaml file will be able to contain multiple yaml documents separated by ---, so a list of documents is used to represents the yaml file ,I will update the comments

src/yaml-language-support/yaml-support.ts
andxu: the yamlDocs represents a parsed results of the yaml file, I keep it in a cache to avoid duplicate parse works since a number of places need the parsed results of the yaml file, eg: hover, complete, schema provider.

src/yaml-language-support/yaml-support.ts
andxu: good advice for 'nodeAtPosition'

src/yaml-language-support/yaml-support.ts
andxu: yes

src/yaml-language-support/yaml-support.ts
andxu: I will remove the yamlDocs from YamlMatchedElement

src/yaml-language-support/yaml-support.ts
andxu: the utils provided by yaml-parser have some util functions for coping with yaml nodes, the user can reference it directly, I will remove the pass through export.

src/yaml-language-support/yaml-support.ts
andxu: I will add more comments to the call of parse function.

src/helm.hoverProvider.ts
andxu: isPositionInKey will hardly tell us with false for not in key position or exception, so I think the error wrap should be suppressed here.

andxu:
@itowlson any ideas about this PR?

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+
+// the schema for kubernetes
+const _definitions = {};
+
+// load the kubernetes schema and make some modifications to $ref node
+export function loadSchema(schemaFile: string) {
+    const schemaRaw = require(schemaFile);
+    const definitions = schemaRaw.definitions;
+    const toFixed = [];
+    for (let key of Object.keys(definitions)) {
```
itowlson: Strongly favour `const key of...` - there should be no reason to modify the range variable.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+import * as vscode from 'vscode';
+import { yamlLocator } from "./yaml-locator";
+import { VSCODE_YAML, KUBERNETES_SCHEMA, KUBERNETES_GROUP_VERSION_KIND, GROUP_VERSION_KIND_SEPARATOR } from "./yaml-constant";
+
+// the schema for kubernetes
+const _definitions = {};
+
+// load the kubernetes schema and make some modifications to $ref node
+export function loadSchema(schemaFile: string) {
+    const schemaRaw = require(schemaFile);
```
itowlson: Is `require` idiomatic?  I like the conciseness, but I usually think of `require` as loading a code module.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+    const toFixed = [];
+    for (let key of Object.keys(definitions)) {
+        let node = definitions[key];
+        const name = key;
+        if (node[KUBERNETES_GROUP_VERSION_KIND]) {
+            let groupKindNode = node[KUBERNETES_GROUP_VERSION_KIND];
+            delete node[KUBERNETES_GROUP_VERSION_KIND];
+            if (groupKindNode.length) {
+                groupKindNode.forEach((groupKindNodeItem) => {
+                    const apiVersion = ((groupKindNodeItem.Group ? groupKindNodeItem.Group + '/' : '') + groupKindNodeItem.Version);
```
itowlson: I'm starting to find this hard to follow.  Pull out some named helper functions here to describe the intent at the various levels.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+                name,
+                ...node
+            };
+        }
+
+        toFixed.push(node['properties']);
+    }
+
+    for (let fix of toFixed ) {
+        patchOnRef(fix);
```
itowlson: It's hard to ascertain what's going on here - none of the names `toFixed`, `fix` or `patchOnRef` really give me a handle on the intent.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+export async function registerYamlSchemaSupport() {
+    loadSchema(KUBERNETES_SCHEMA);
+    let yamlPlugin;
```
itowlson: Seems like this initial block is trying to find and activate the RedHat YAML extension?  If so, pull it out into a helper function called e.g. `activateYamlExtension`.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+                yamlPlugin = await Promise.resolve(ext.activate());
+                break;
+            } catch (error) {
+                console.log('Failed to activate the Yaml Extension: ' + error);
+            }
+            break;
+        }
+    }
+    if (!yamlPlugin) {
+        console.log('vscode-yaml plugin doesn\'t expose the entry to register custom schema provider.');
```
itowlson: What does this mean?  The user won't see this message so this will silently return `undefined`... what is the impact of that?

package.json
```diff
@@ -493,12 +493,14 @@
...
         "k8s": "^0.4.12",
         "lodash": ">3",
+        "node-yaml-parser": "^0.0.6",
         "opn": "^5.2.0",
         "pluralize": "^4.0.0",
         "shelljs": "^0.7.7",
         "tmp": "^0.0.31",
         "uuid": "^3.1.0",
         "vscode-extension-telemetry": "^0.0.6",
+        "vscode-uri": "^1.0.1",
```
itowlson: Should we be declaring a package dependency on the RedHat YAML extension?

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+    for (let fix of toFixed ) {
+        patchOnRef(fix);
+    }
+}
+
+export async function registerYamlSchemaSupport() {
+    loadSchema(KUBERNETES_SCHEMA);
+    let yamlPlugin;
+    const installedExtensions: any[] = vscode.extensions.all;
+    for (let ext of installedExtensions) {
```
itowlson: Prefer `const` to `let` (check throughout).

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+        if (ext.id === VSCODE_YAML) {
```
itowlson: Do we need to do the loop/if check?  Can we not use `vscode.extensions.getExtension`?

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+            }
+            break;
+        }
+    }
+    if (!yamlPlugin) {
+        console.log('vscode-yaml plugin doesn\'t expose the entry to register custom schema provider.');
+        return;
+    }
+    // register for kubernetes schema provider
+    yamlPlugin.registerContributor('kubernetes', (resource) => {
```
itowlson: This is too much to be done in lambdas - the `registerContributor` signature gives me no clue as to what the `resource` and `uri` handlers mean, so it's hard for me to get my head around the complex contents of the lambdas.  Let's try to break this into intention-revealing named functions.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+            return undefined;
+        }
+    }, (uri) => {
+        let _uri = Uri.parse(uri);
+        if (!_uri.path || ! _uri.path.startsWith('/')) {
+            return '';
+        }
+        if (_uri.path.includes('+')) {
+            const manifestRef = _uri.path.slice(1).split('+').map((r) => {
+                return {$ref: `kubernetes://manifest/${r}`};
```
itowlson: You're clearly lining up URIs between here and the `resource` lambda so pull out the URI generation into a shared method.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+            });
+            return jsonToString({ oneOf: [...manifestRef] });
+        }
+        let manifestType = _uri.path.slice(1);
+        //apiVersionKind[0] is apiVersion and apiVersionKind[1] is kind
```
itowlson: Defining a type and a parser function, rather than using an array, would make this more self-documenting.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+        return undefined;
+    });
+}
+
+// convert '#/definitions/com.github.openshift.origin.pkg.build.apis.build.v1.ImageLabel' to
+// 'com.github.openshift.origin.pkg.build.apis.build.v1.ImageLabel'
+const getNameInDefinitions = ($ref: string): string => {
+    const prefix = '#/definitions/';
+    if ($ref.includes(prefix)) {
```
itowlson: Don't you need `startsWith` rather than `includes`, or are you wanting to slice from here even if it occurs in non-prefix positions?

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+        return $ref.slice($ref.indexOf(prefix) + prefix.length);
+    } else {
+        throw new Error('bad reference name.');
```
itowlson: This error is not helpful if the user ever sees it.  (If it's a "can't happen" error then please add a test that it _doesn't_.  **grin**)

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+            value.$ref = `kubernetes://manifest/${match.name.toLowerCase()}`;
+        } else {
+            throw new Error(`Unsupported $ref value ${value.$ref}.`);
+        }
+    }
+}
+
+// convert a json object to string
+function jsonToString(json: any): string {
+    return JSON.stringify(json);
```
itowlson: Does this add any value over `JSON.stringify`?  I don't feel we're adding any levels of abstraction or readability here.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+    if (_.isArray(node)) {
+        for (let subItem of node) {
+            patchOnRef(subItem);
+        }
+    }
+    if (!_.isObject(node)) {
+        return;
+    }
+    for(let key of Object.keys(node)) {
```
itowlson: `const`

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+        patchOnRef(node[key]);
+    }
+
+    const value = node;
+    if (value && value.$ref) {
+        const match = _definitions[getNameInDefinitions(value.$ref.toLowerCase())];
+        if (match) {
+            value.$ref = `kubernetes://manifest/${match.name.toLowerCase()}`;
```
itowlson: There's `toLowerCase`-ing going on here which I didn't see in the other references to `kubernetes://manifest`.  I'm assuming it does all line up somewhere behind the scenes, but I feel a bit at a loss as to how.  I realise that this sort of node traversal code is inherently abstract and hard to communicate clearly to readers, but it would be good if you could refactor it a bit to make the nature of the URIs and transformations clearer - introducing a few helper functions and un-inlining some of the blocks of code would go a long way I think.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+                // TODO, unwrap quotes
+                const apiVersion = rootMapping.mappings.find((map) => map.key && map.key.raw === 'apiVersion');
+                const kind = rootMapping.mappings.find((map) => map.key && map.key.raw === 'kind');
+                if (apiVersion && kind) {
+                    choices.push(apiVersion.value.raw + GROUP_VERSION_KIND_SEPARATOR + kind.value.raw);
+                }
+            });
+            choices = _.uniq(choices);
+            if (choices.length > 1) {
+                return 'kubernetes://combine/' + _.uniq(choices).join('+');
```
itowlson: I don't see this URL prefix referenced elsewhere.  Is it understood by the YAML extension by virtue of the `kubernetes` contributor being registered, and `combine` being a well-known keyword?  If so, you need to make it clearer that the `kubernetes` registration key and the URL scheme are related.

src/extension.ts
```diff
@@ -191,6 +193,8 @@ export function activate(context) {
     subscriptions.forEach((element) => {
         context.subscriptions.push(element);
     }, this);
+
+    return registerYamlSchemaSupport();
```
itowlson: `registerYamlSchemaSupport` is async so returns a promise - I don't think we want to return that from `activate` do we?

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,165 @@
...
+        if (matchedSchema) {
+            return jsonToString(matchedSchema);
+        }
+        return undefined;
+    });
+}
+
+// convert '#/definitions/com.github.openshift.origin.pkg.build.apis.build.v1.ImageLabel' to
+// 'com.github.openshift.origin.pkg.build.apis.build.v1.ImageLabel'
+const getNameInDefinitions = ($ref: string): string => {
```
itowlson: Not sure why we switch between `function foo() { ... }` and `const foo = (...) => { ... };`?

src/yaml-support/yaml-schema.ts
andxu: ok, require will have bad side effect of loading js rather than json.

src/extension.ts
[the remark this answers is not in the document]
andxu: I don't know this plugin will expose any api publicly, registerYamlSchemaSupport will not expose anything, I will fix on it.

src/yaml-support/yaml-constant.ts
```diff
@@ -0,0 +1,9 @@
+import * as path from 'path';
+
+export const VSCODE_YAML = 'redhat.vscode-yaml';
```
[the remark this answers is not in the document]
andxu: sure

src/yaml-support/yaml-schema.ts
andxu: sure, const is better

src/extension.ts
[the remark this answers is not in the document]
andxu: Then I need to make activate async since registerYamlSchemaSupport is async mothod.

export async function activate(context) {

src/yaml-support/yaml-schema.ts
andxu: I will change it to function style.

src/extension.ts
andxu: registerYamlSchemaSupport  is async, the result is not import and should be discarded.

src/yaml-support/yaml-schema.ts
andxu: it is useless for the phase combine, just a placeholder for uri host, we want the schema name as the `path` part rather than the host part.

src/yaml-support/yaml-schema.ts
andxu: The old version of vscode-yaml plugin without my custom schema code will not expose the register... apis, just ignore it and the yaml schema will not be supported.

package.json
andxu: Sure

src/extension.ts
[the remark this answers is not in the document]
itowlson: I don't think you need to make `activate` async unless you need to await on the result of `registerYamlSchemaSupport`.

src/yaml-support/yaml-schema.ts
itowlson: It sounds like this would result in the user not getting the schema support, and not knowing why.  I feel we should either declare a package dependency with minimum version, or display a warning explaining why we're not providing (intellisense / schema validation / whatever the schema is giving us) and how they can fix it.

bnookala:
@andxu @itowlson Do you think we should be using k8s semantic versioning for the `swagger.json` file?

I'd imagine a situation in which a user might be on 1.7.* cluster, and the swagger.json file we've bundled is using definitions for a 1.9.* cluster. 

Additionally, this might open up opportunities to use the configured cluster's kubernetes version to load the appropriate swagger.json file.

itowlson:
@bnookala The Explain command already has code to download and cache the Swagger from the cluster (though we should enhance it to handle switching clusters following Jinbo's recent work), if that helps.

andxu:
@bnookala thank you very much for the advice, in this pr, I think the latest schema will be enough not to make changes too much, I will create a issue named providing the version compatible schema for kubernetes, another concern about version is some users may not have a kubernetes available and they want to edit yaml files and some users have multiple clusters with different versions, we can discussion all concerns incuding how to get user version from cluster or preference in that issue.

andxu:
@itowlson @bnookala  any other ideas about this PR?

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+    VSCODE_YAML_EXTENSION_ID, KUBERNETES_SCHEMA, KUBERNETES_GROUP_VERSION_KIND, GROUP_VERSION_KIND_SEPARATOR,
+    KUBERNETES_SCHEMA_FILE
+} from "./yaml-constant";
+import * as util from "./yaml-util";
+
+// the schema for kubernetes
+const _definitions = {};
+
+// load the kubernetes schema and make some modifications to $ref node
+export function loadSchema(schemaFile: string) {
```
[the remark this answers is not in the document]
andxu: added void.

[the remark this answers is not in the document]
andxu: close the expose.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+    const schemaRaw = util.loadJson(schemaFile);
+    const definitions = schemaRaw.definitions;
+    const toFixed = [];
```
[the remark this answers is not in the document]
andxu: does convertRefToSchemaURIs name the action?

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+    for (const name of Object.keys(definitions)) {
+        const currentSchema = definitions[name];
+        if (currentSchema[KUBERNETES_GROUP_VERSION_KIND] && currentSchema[KUBERNETES_GROUP_VERSION_KIND].length) {
+            // if the schema contains 'x-kubernetes-group-version-kind'. then it is a direct kubernetes manifest,
```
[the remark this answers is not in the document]
andxu: Done as suggested

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+
+        // fix on each node in properties for $ref since it will directly reference '#/definitions/...'
+        // we need to convert it into schema like 'kubernetes://schema/...'
+        // we need also an array to collect them since we need to get schema from _definitions, at this point, we have
+        // not finished the process of add schemas to _definitions, call patchOnRef will fail for some cases.
+        toFixed.push(currentSchema['properties']);
+    }
+
+    for (const fix of toFixed ) {
+        patchOnRef(fix);
```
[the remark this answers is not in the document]
andxu: Yes, all $refs to `#/definitions/...`  needs to be replaced since there is no central definitions in our schema provided to vscode-yaml.

src/yaml-support/yaml-schema.ts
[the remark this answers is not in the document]
itowlson: Yes - as a reader, it would certainly set me on the right track to understanding.  For the array, I'd prefer a noun or adjective, but if it's hard to come up with a good one, then at least this would point me in the direction of what the array was for.

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+    const prefix = '#/definitions/';
+    if ($ref.startsWith(prefix)) {
+        return $ref.slice(prefix.length);
+    } else {
+        return prefix;
+    }
+}
+
+// patch on schema $ref with values like 'kubernetes://schema/...'
+function patchOnRef(node) {
```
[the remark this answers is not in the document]
andxu: replaceDefinitionRefsWithYamlSchemaUris ?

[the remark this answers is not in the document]
itowlson: Works for me - thanks!

src/yaml-support/yaml-schema.ts
```diff
@@ -1,128 +1,184 @@
...
+    readonly 'x-kubernetes-group-version-kind'?: any[];
+    readonly properties?: { [key: string]: any; };
+}
+
+// The function signature exposed by vscode-yaml:
+// 1. the requestSchema api will be called by vscode-yaml extension to decide whether the schema can be handled by this
+// contributor, if it returns undefined, means it doesn't support this yaml file, vscode-yaml will ask other contributors
+// 2. the requestSchemaContent api  will give the parameter uri returned by the first api, and ask for the json content(after stringify) of
+// the schema
+declare type YamlSchemaContributor = (schema: string,
```
itowlson: YES!  Just these three lines clarify things _enormously_.  Thank you!

src/yaml-support/yaml-util.ts
```diff
@@ -90,24 +95,29 @@ export function equalIgnoreCase(a: string, b: string) {
...
+// get the string value in a javascript object with key(may be case sensitive due to the third parameter)
+function getStringValue(node, key: string, ignoreCase: StringComparison = StringComparison.Ordinal): string {
     if (!node) {
         return undefined;
     }
     if (node.hasOwnProperty(key)) {
-        return node[key];
+        return <string>node[key];
     }
     if (ignoreCase) {
```
itowlson: This needs to be `if (ignoreCase === StringComparison.OrdinalIgnoreCase)` I think.  (You might want to rename the `ignoreCase` parameter to `comparisonOption` or `caseMode` or something like that.)

src/yaml-support/yaml-schema.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+}
+
+// find the schema saved at _definitions
+function findSchemaByIdOrName(key) {
+    // property keys in _definitions are always low case
+    key = key.toLowerCase();
+    let matchedSchema = _definitions[key];
+    if (!matchedSchema) {
+        // if we cannot find any schema in _definitions for manifestType, we will iterate the values in  _definitions to
```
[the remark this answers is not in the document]
andxu: findSchemaByIdOrName is removed.

src/yaml-support/yaml-util.ts
```diff
@@ -0,0 +1,114 @@
...
+//      key2: value2
+//
+export function getPropertyValue(mapNode: YamlMap, key: string, ignoreCase: boolean = false) {
+    // TODO, unwrap quotes
+    return mapNode.mappings.find((mapping) => mapping.key &&
+        (ignoreCase ? key === mapping.key.raw : equalIgnoreCase(key, mapping.key.raw)));
+}
+
+// get the value in a javascript object with key(may be case sensitive due to the third parameter)
+function getValue(node, key: string, ignoreCase: boolean = false) {
```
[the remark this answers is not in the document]
andxu: done

src/yaml-support/yaml-util.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// create a $ref schema for kubernetes manifest
+export function makeRefOnKubernetes(id: string) {
+    return { $ref: makeKubernetesUri(id) };
+}
+
+
+// extract id, apiVersion, kind from x-kubernetes-group-version-kind node in schema
+export function getKubernetesGroupVersionKind(groupKindNodeItem) {
```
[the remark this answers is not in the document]
andxu: groupKindNodeItem is a swagger json node, it has group, kind, version or Group, Kind, Version in different swagger versions, so i suggest to use any, the output is string

PR #66, opened 2018-03-06
API-based cluster provider

itowlson:
Previously the 'Create Cluster' and 'Configure from Cluster' commands were implemented by an unholy hellscape of intricate, entangled code involving impressive abuse of VS Code command URLs and a truly staggering number of backslashes established only after weeks of trial and error.  This created something of a barrier to entry for anyone wanting to extend the system to support Minikube, GKE, EKS, etc.

This PR defines a registration API and a HTTP protocol for third-party cluster providers, and ports the Azure provider to that system.  The Azure provider is refactored into a HTTP server component and a utility layer that invokes the Azure and Kubernetes CLIs.


bnookala:
YESSSSSS

CONTRIBUTING.md
```diff
@@ -10,8 +10,20 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
...
+* Get the VS Code Kubernetes extension's API object via the `vscode.extensions.getExtension<T>` and `Extension<T>.activate()` functions.
+* The API object has a field named `clusterProviderRegistry`.  This is an object you can use to register the type(s) of cluster you support.  Call the registry's `register` method, passing an object with the following fields:
+  * `id` (string): an identifier for the type of cluster.  This needs to be unique across all providers, as we use it to dispatch the user's selected cluster type to the appropriate handler.  E.g. for Azure Kubernetes Service we use `aks`.
+  * `displayName` (string): the display name for the type of cluster.  E.g. for Azure Kubernetes Service we use (wait for it) `Azure Kubernetes Service`.
+  * `port` (number): the port on which your provider will serve the requisite configuration pages.  We reserve port 44010 for the internal 'select cluster type' page provider.  Also, the Azure provider uses port 44011, so please don't use that either.
```
squillace: Ports 44010 and 44011 are reserved for the internal 'select cluster type' page provider and the Azure provider.

bnookala: I wonder if we'll need to reserve these ports, since the cluster registry is the entry point for creation and configuration of clusters. I'm thinking that If we find an unbound port via node, we can bind to it, and pass it back to the clusterRegistry.register() function, eliminating the "reserved" port thing altogether. A quick npm search reveals this module, which might be useful: https://www.npmjs.com/package/get-port

testforstephen:
Agree with @bnookala , writing raw HTML is not a good experience and hard to maintain, actually we could take vscode as a browser and host the web page by a standalone nodejs express server. In our last project https://github.com/Microsoft/vscode-arduino, we have tried to use React to construct visual web view in vscode, it's easier to define css and implement web view.

itowlson:
@bnookala @testforstephen This sounds like a great idea, but it's beyond by current level of aptitude with these tools.  It seems like it would be a great way to get my head around them, but I'm reluctant to tackle it now because we have limited time.  And the pages are so simple that the raw HTML is not _too_ bad to write... but definitely for anything more complex it would be good!

src/azure.ts
```diff
@@ -102,8 +120,211 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+    return '';
+}
+
+function getListClustersCommand(context: Context, clusterType: string) : string {
+    let filter = listClustersFilter(clusterType);
+    let query = `[${filter}].{name:name,resourceGroup:resourceGroup}`;
+    if (context.shell.isUnix()) {
+        query = `'${query}'`;
+    }
+    return `az ${getClusterCommand(clusterType)} list --query ${query} -ojson`;
```
testforstephen: typos. `-ojson` -> `-o json`

src/azure.ts
```diff
@@ -102,8 +120,211 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+    if (sr.code === 0 && !sr.stderr) {
+        const clusters : ClusterInfo[] = JSON.parse(sr.stdout);
+        return { succeeded: true, result: clusters, error: [] };
+    } else {
+        return { succeeded: false, result: [], error: [sr.stderr] };
+    }
+}
+
+function listClustersFilter(clusterType: string): string {
+    if (clusterType == 'acs') {
```
testforstephen: `==` -> `===`

src/azure.ts
```diff
@@ -102,8 +120,211 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+    return `az ${getClusterCommand(clusterType)} list --query ${query} -ojson`;
+}
+
+async function listLocations(context: Context) : Promise<Errorable<Locations>> {
+    let query = "[].{name:name,displayName:displayName}";
+    if (context.shell.isUnix()) {
+        query = `'${query}'`;
+    }
+
+    const sr = await context.shell.exec(`az account list-locations --query ${query} -ojson`);
```
testforstephen: `-ojson` -> `-o json`

src/azure.ts
```diff
@@ -102,8 +120,211 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+}
+
+export async function listAcsLocations(context: Context) : Promise<Errorable<ServiceLocation[]>> {
+    const locationInfo = await listLocations(context);
+    if (!locationInfo.succeeded) {
+        return { succeeded: false, result: [], error: locationInfo.error };
+    }
+    const locations = locationInfo.result;
+
+    const sr = await context.shell.exec(`az acs list-locations -ojson`);
```
testforstephen: `-ojson` -> `-o json`

src/azure.ts
```diff
@@ -102,8 +120,211 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+}
+
+function locationDisplayNamesEx(production: string[], preview: string[], locationInfo: Locations) : ServiceLocation[] {
+    let result = locationDisplayNames(production, false, locationInfo) ;
+    result = result.concat(locationDisplayNames(preview, true, locationInfo));
+    return result;
+}
+
+export async function listVMSizes(context: Context, location: string) : Promise<Errorable<string[]>> {
+    const sr = await context.shell.exec(`az vm list-sizes -l "${location}" -ojson`);
```
testforstephen: `-ojson` -> `-o json`

src/azure.ts
```diff
@@ -102,8 +120,211 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+        const response : any[] = JSON.parse(sr.stdout);
+        const result = response.map((r) => r.name as string);
+        return { succeeded: true, result: result, error: [] };
+    } else {
+        return { succeeded: false, result: [], error: [sr.stderr] };
+    }
+}
+
+async function resourceGroupExists(context: Context, resourceGroupName: string) : Promise<boolean> {
+    const sr = await context.shell.exec(`az group show -n "${resourceGroupName}" -ojson`);
```
testforstephen: `-ojson` -> `-o json`

src/azure.ts
```diff
@@ -102,8 +120,211 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
...
+    };
+}
+
+async function listClustersAsync(context: Context, clusterType: string) : Promise<Errorable<ClusterInfo[]>> {
+    let cmd = getListClustersCommand(context, clusterType);
+    const sr = await context.shell.exec(cmd);
+
+    if (sr.code === 0 && !sr.stderr) {
+        const clusters : ClusterInfo[] = JSON.parse(sr.stdout);
+        return { succeeded: true, result: clusters, error: [] };
```
testforstephen: i saw there were many places to return the same formatted success or failure data. Can we  provide two util functions for them? such as `wrapAsSuccessResult(result: T): Errorable<T>`, `wrapAsFailureResult(result: T, error: string[]): Errorable<T>`.

bnookala:
@itowlson agreed! I ~ feel ~ that we can make this a bit better, but I I will look into it.

src/azure.ts
```diff
@@ -2,7 +2,7 @@
```
testforstephen: this file looks like an azure utility, shouldn't we move it to the folder `src/components/clusterprovider/azure`?

src/azure.ts
itowlson: `-ojson` is accepted as far as I can tell...?

src/azure.ts
itowlson: Is this a style preference?  I don't have strong feelings either way but have just developed the habit of `-ojson`.

CONTRIBUTING.md
itowlson: We already have portfinder in the project from the debugging work, but I tried using it to replace the literal ports and it didn't work (the second server was assigned the same port as the first, even though the first one was already listening).  @testforstephen any advice on the successful use of portfinder?  If I can't get it working, I'll leave it as fixed ports, as it will be a non-breaking change to make them dynamic later.

itowlson:
@bnookala @testforstephen Ready for round two!  Some comments not actioned for reasons discussed in the replies - let me know if you feel strongly about them.

src/azure.ts
testforstephen: i take it as typos by mistake, it's OK for your usage.

CONTRIBUTING.md
itowlson: Brilliant!  Thank you!

src/components/clusterprovider/azure/azure.ts
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    if (forCommand == 'acs') {
+        prereqCheckSSHKeys(context, errors);
+    }
+
+    return errors;
+}
+
+async function azureCliVersion(context: Context) : Promise<string> {
```
bnookala: i feel like we should use https://github.com/npm/node-semver here

src/components/clusterprovider/azure/azure.ts
```diff
@@ -0,0 +1,388 @@
...
+        const homeDir = process.env['HOME'];
+        const installDir = homeDir + '/bin';
+        const installFile = installDir + '/kubectl';
+        const cmd = `mkdir -p "${installDir}" ; ${cmdCore} --install-location="${installFile}"`;
+        return { installFile: installFile, commandLine: cmd };
+    }
+}
+
+function getClusterCommand(clusterType: string) : string {
+    if (clusterType == 'Azure Container Service' || clusterType == 'acs') {
```
bnookala: ===

src/components/clusterprovider/azure/azure.ts
```diff
@@ -0,0 +1,388 @@
...
+        return 'acs';
+    }
+    return 'aks';
+}
+
+function getClusterCommandAndSubcommand(clusterType: string) : string {
+    if (clusterType == 'Azure Container Service' || clusterType == 'acs') {
```
bnookala: ===

src/components/clusterprovider/clusterproviderutils.ts
```diff
@@ -0,0 +1,50 @@
...
+                visibility: hidden;
+            }
+            .vscode-dark #theme-canary {
+                display: none;
+                visibility: visible;
+            }
+        </style>
+        <script>
+        function styleme() {
+            // TODO: shameful, shameful, shameful
```
bnookala: indeed

src/extension.ts
```diff
@@ -191,6 +196,11 @@ export function activate(context) {
     subscriptions.forEach((element) => {
         context.subscriptions.push(element);
     }, this);
+
+    return {
+        apiVersion: '0.1',
+        clusterProviderRegistry: clusterProviderRegistry
+    };
```
bnookala: can you talk about what this does? never seen this pattern of returning an object from the activate function before

itowlson:
I don't want to remove ACS while AKS is in preview and only in certain regions.  Even then there might be cases where people want to have more access to their control plane, I dunno.  I expect to make AKS the default once it's GA.

src/extension.ts
itowlson: The return value of `activate` is an object that other extensions can interact with if they call `getExtension<T>` and `extension<T>.activate`.  So it acts as a way to surface an API to your extension.  https://code.visualstudio.com/docs/extensionAPI/vscode-api#_extensions

src/components/clusterprovider/azure/azure.ts
itowlson: We currently use https://www.npmjs.com/package/compare-versions.  I'm not sure what `semver` offers to justify switching.

bnookala: argh derp;  i forgot that the output for `az --version` looks like:

```
azure-cli (2.0.27)

acr (2.0.21)
acs (2.0.26)
advisor (0.1.2)
appservice (0.1.26)
backup (1.0.6)
batch (3.1.10)
batchai (0.1.5)
billing (0.1.7)
cdn (0.0.13)
cloud (2.0.12)
cognitiveservices (0.1.10)
command-modules-nspkg (2.0.1)
configure (2.0.14)
consumption (0.2.1)
container (0.1.18)
core (2.0.27)
cosmosdb (0.1.19)
dla (0.0.18)
dls (0.0.19)
eventgrid (0.1.10)
extension (0.0.9)
feedback (2.1.0)
find (0.2.8)
interactive (0.3.16)
iot (0.1.17)
keyvault (2.0.18)
lab (0.0.17)
monitor (0.1.2)
network (2.0.23)
nspkg (3.0.1)
profile (2.0.19)
rdbms (0.0.12)
redis (0.2.11)
reservations (0.1.1)
resource (2.0.23)
role (2.0.19)
servicefabric (0.0.10)
sql (2.0.21)
storage (2.0.25)
vm (2.0.26)

Python location '/usr/local/opt/python3/bin/python3.6'
Extensions directory '/Users/bhargav/.azure/cliextensions'

Python (Darwin) 3.6.4 (default, Jan 23 2018, 14:22:14)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]

Legal docs and information: aka.ms/AzureCliLegal
````
… nvm my suggestion then - i was wondering why you were using the regex parsing, and now i get it.

PR #76, opened 2018-03-08
Offer all pods for exec/terminal

### LINKED ISSUE #65: Command "Kubernetes exec" and "kubernetes terminal" throw errors "Couldn't find any relevant pods."
Repro steps:

1. open vscode with kubernetes plugin ,press F1, select "Kubernetes exec" or "kubernetes terminal"
2. type a command like bash 

Expect: a bash open in terminal 

Actual:
image

But I can open terminal by right click on the pod.
image

itowlson:
When you run Exec or Terminal, the extension uses a label that it applies during Run to locate pods associated with the current application.  However, this causes an unhelpful error if the application was not deployed with the extension.  With this PR, if there are no pods identified for the current application, the extension displays a list of all pods, and the user can exec or open a terminal on any of them.

To simplify the functional change, I converted the original code from callback style to async/await.  Reviewers can see these two commits separately to check the fidelity of the conversion.  It would be nice to more strongly type the pod info after deserialisation (there was a latent bug if the app had more than one pod, which types would have caught), but I propose to defer this for now.

Fixes #65.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -620,30 +620,37 @@ function findVersionInternal(fn) {
     });
 }
 
-function findPods(labelQuery, callback) {
-    kubectl.invoke(` get pods -o json -l ${labelQuery}`, (result, stdout, stderr) => {
-        if (result !== 0) {
-            vscode.window.showErrorMessage('Kubectl command failed: ' + stderr);
-            return;
-        }
-        try {
-            let podList = JSON.parse(stdout);
-            callback(podList);
-        } catch (ex) {
-            console.log(ex);
-            vscode.window.showErrorMessage('unexpected error: ' + ex);
-        }
-    });
+async function findPods(labelQuery: string) : Promise<FindPodsResult> {
+    const sr = await kubectl.invokeAsync(` get pods -o json -l ${labelQuery}`);
+
+    if (sr.code !== 0) {
+        vscode.window.showErrorMessage('Kubectl command failed: ' + sr.stderr);
+        return { succeeded: false, pods: [] };
+    }
+
+    try {
+        const podList = JSON.parse(sr.stdout);
+        return { succeeded: true, pods: podList.items };
+    } catch (ex) {
+        console.log(ex);
+        vscode.window.showErrorMessage('unexpected error: ' + ex);
+        return { succeeded: false, pods: [] };
+    }
 }
 
-function findPodsForApp(callback) {
+async function findPodsForApp() : Promise<FindPodsResult> {
     let appName = path.basename(vscode.workspace.rootPath);
-    findPods(`run=${appName}`, callback);
+    return await findPods(`run=${appName}`);
 }
 
-function findDebugPodsForApp(callback) {
+async function findDebugPodsForApp() : Promise<FindPodsResult> {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -1045,31 +1072,32 @@ async function isBashOnPod(podName : string): Promise<boolean> {
...
-        selectContainerForPod(pod, (container) => {
-            if (!container) {
-                return;
-            }
-            let pieces = container.image.split(':');
-            if (pieces.length !== 2) {
-                vscode.window.showErrorMessage(`unexpected image name: ${container.image}`);
-                return;
-            }
+async function syncKubernetes() : Promise<void> {
```
bnookala: i have no idea what this function is doing, and that's ok because i just realized we have https://github.com/Azure/vscode-kubernetes-tools/issues/68 to clear this up 🙃

src/extension.ts
```diff
@@ -852,44 +867,60 @@ function findPod(callback) {
...
+async function selectPod(scope: PodSelectionScope, fallback: PodSelectionFallback) : Promise<any | null> {
+    const findPodsResult = scope === PodSelectionScope.App ? await findPodsForApp() : await findAllPods();
+    
+    if (!findPodsResult.succeeded) {
+        return null;
+    }
+
+    const podList = findPodsResult.pods;
+
+    if (podList.length === 0) {
```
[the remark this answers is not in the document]
itowlson: I am... not fond of that idiom, especially when right alongside an `if (podList.length === 1)`.  I'm checking for various values of `length`.  That one of them happens to coincide with a falsey integer conceals the pattern more than clarifying it.

Though I suppose if I felt that strongly, I ought to convert the whole thing to a `switch` statement.

src/extension.ts
itowlson: And there's a PR on the way for that, but it depends on this one (because of the refactoring)...

src/extension.ts
```diff
@@ -869,44 +884,60 @@ function findPod(callback) {
...
-                callback(null);
-                return;
-            }
+    for (const element of podList) {
+        names.push(`${element.metadata.namespace}/${element.metadata.name}`);
+    }
 
-            let ix = value.indexOf('/');
-            let name = value.substring(ix + 1);
+    const value = await vscode.window.showQuickPick(names);
```
testforstephen: Suggest to use `vscode.window.showQuickPick<T extends vscode.QuickPickItem>(items: T[])`, no need to loop the array again to query the original element after picking. 

See the sample https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/debug/providerRegistry.ts#L11

src/extension.ts
```diff
@@ -978,32 +1009,30 @@ function describeKubernetesCore(kindName : string) {
...
     for (const element of pod.spec.containers) {
         names.push(element.name);
     }
 
-    vscode.window.showQuickPick(names).then((value) => {
-        for (const element of pod.spec.containers) {
-            if (element.name !== value) {
-                continue;
-            }
+    const value = vscode.window.showQuickPick(names);
```
testforstephen: Same as the last comment.

PR #81, opened 2018-03-09
update node-yaml-parser version since lineLength name change.

andxu:
1. update node-yaml-parser version since lineLength name change.
2. fix a definition ref error


src/yaml-support/yaml-schema.ts
```diff
@@ -100,7 +100,7 @@ function requestYamlSchemaContentCallback(uri: string): string {
         const manifestRefList = manifestType.split('+').map(util.makeRefOnKubernetes);
         return JSON.stringify({ oneOf: manifestRefList });
     }
-    const schema = _definitions[manifestType];
+    const schema = _definitions[manifestType.toLowerCase()];
```
itowlson: This is where using a dedicated lookup type rather than an object would come in handy, e.g.

```
class Definitions {
  private object _definitions;
  function lookup(key: string) : any /* or whatever it is */ {
    return this._definitions[key.toLowerCase()];
  }
}
```

Then you don't need to worry about remembering to call `toLowerCase()` at every call site.

andxu: I will  refacter it later

Issue #68, opened 2018-03-06
Sync command is hard to understand


itowlson:
The Sync command takes your local repo back to the commit corresponding to the version of the application in the cluster.  This is really nice for debugging, hotfixing, etc.  However, the command only works properly if the app was deployed using the extension (and therefore uses the magic image tagging convention implemented by the extension), and on failure displays unhelpful messages such as `git checkout failed: 1`.  We should:

* Review if Sync is the best name for this command
* Document the broader workflow in which Sync is useful
* Document the behaviour and the convention it relies on
* Provide better error messages on failure

itowlson:
I've addressed these to a great extent in #84 though there's still scope for better docs, so leaving this open for now.

PR #84, opened 2018-03-11
Improved Sync command messages and error handling

itowlson:
It was not clear what the Sync command was doing, and the errors were often opaque.  This PR provides a confirmation message of what the command is about to do (check out a Git commit based on the image version), and provides clearer explanations when it fails.

It goes some way towards fixing #68, but we still need to join it up more clearly with other commands through process and/or documentation.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -40,6 +40,7 @@ import { Reporter } from './telemetry';
 import * as telemetry from './telemetry-helper';
 import * as extensionapi from './extension.api';
 import {dashboardKubernetes} from './components/kubectl/proxy';
+import { Git } from './components/git/git';
 import { DebugSession } from './debug/debugSession';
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -1081,31 +1083,36 @@ async function isBashOnPod(podName : string): Promise<boolean> {
 
 async function syncKubernetes() : Promise<void> {
```
testforstephen: From the command name `Kubernetes: Sync`, i totally don't understand what it did. Can we change it to `Git-Checkout` or `Checkout` directly?

itowlson: I agree it's a pretty cryptic name.  However, I think of checkout as an implementation detail (admittedly a pretty darned blatant one!).  I've proposed `Sync Working Copy to Cluster`, and added some docs.  Thoughts?

testforstephen: Yeah, the new name is better.

PR #86, opened 2018-03-12
When doing a Get for the resource open in the editor, use the terminal

### LINKED ISSUE #85: Show the result for command palette "Kubernetes: Get" in TERMINAL

itowlson:
Fixes #85.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -583,14 +583,8 @@ function getKubernetes(explorerNode? : any) {
         const id = explorerNode.resourceId || explorerNode.id;
         kubectl.invokeInTerminal(`get ${id} -o wide`);
     } else {
-        // TODO Show the result for get command in TERMINAL window.
-        let kindName = findKindName();
-        if (kindName) {
-            maybeRunKubernetesCommandForActiveWindow('get --no-headers -o wide -f ');
-            return;
-        }
         findKindNameOrPrompt(kuberesources.commonKinds, 'get', { nameOptional: true }, (value) => {
-            kubectl.invoke(" get " + value + " -o wide --no-headers");
+            kubectl.invokeInTerminal(` get ${value} -o wide`);
         });
     }
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -562,20 +562,21 @@ function loadKubernetesCore(value : string) {
 }
 
 function exposeKubernetes() {
-    let kindName = findKindName();
-    if (!kindName) {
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -583,14 +583,8 @@ function getKubernetes(explorerNode? : any) {
         const id = explorerNode.resourceId || explorerNode.id;
         kubectl.invokeInTerminal(`get ${id} -o wide`);
     } else {
-        // TODO Show the result for get command in TERMINAL window.
-        let kindName = findKindName();
-        if (kindName) {
-            maybeRunKubernetesCommandForActiveWindow('get --no-headers -o wide -f ');
-            return;
-        }
         findKindNameOrPrompt(kuberesources.commonKinds, 'get', { nameOptional: true }, (value) => {
```
testforstephen: Should improve the workflow for `findKindNameOrPrompt` function.
1. When i closed all editors in vscode, run `Kubernetes: Get` will fail with message `No active editor`. Should not pop up this message.
2. In `promptKindName` function, it pops up a input box with placeholder saying `Empty string to be prompted`, actually i have no idea what to do here. Can we remove this logic and show the resource type pickup box directly?

itowlson: 1. Good catch, thank you!
2. No, I don't think we can, because that would prevent people from typing their resource id (e.g. `deploy/nginx`) directly into the box.  We _could_ reverse the order, e.g. have a drop down with resource types _and_ a special "I'll enter the resource ID directly" option.  I'm not sure I much like that though, and I definitely don't want to slip such a change (which would affect several other commands) into this PR.

itowlson: Fixed point 1 (note this has knock-on effects for other commands, which I think are beneficial, but I also feel we need to avoid going too far down a rabbit hole in this PR).

testforstephen: 2. In current design, the user is hard to discover the follow-up pick up box at the first glance. I like the idea to add a special "I'll enter the resource ID directly" option.

itowlson: That's fair enough, but I'd like to keep it out of scope for this PR, as it affects multiple other commands in a way that is more than a bug fix and is somewhat a matter of taste, and therefore merits discussion and agreement!  Let's raise an issue for changing the 'pick from list or type an ID' flow as a separate piece of work.

testforstephen: Agree. Let's create an issue to track the discussion.

PR #88, opened 2018-03-12
 create a KubernetesSchemaHolder to store/lookup schema.

andxu:
create a KubernetesSchemaHolder to store/lookup schema.


src/yaml-support/yaml-schema.ts
```diff
@@ -25,30 +25,102 @@ export interface KubernetesSchema {
...
-// the schema for kubernetes
-const _definitions: { [key: string]: KubernetesSchema; } = {};
-
-// load the kubernetes schema and make some modifications to $ref node
-function loadSchema(schemaFile: string): void {
-    const schemaRaw = util.loadJson(schemaFile);
-    const definitions = schemaRaw.definitions;
-    for (const name of Object.keys(definitions)) {
-        saveSchemaWithManifestStyleKeys(name, definitions[name]);
+export class KubernetesSchemaHolder {
```
itowlson: Does this need to be `export`?  The original functions weren't exported, so I feel like this may be used only within this module, not sure?  If it does need to be exported, consider exporting an interface containing only the functions required for other modules to interact with it (probably just `lookup`?).

itowlson: Blank line between previous type declaration and this one, I'd suggest.

andxu: We need not expose it at all.

andxu: blank line added

PR #89, opened 2018-03-12
Use JVM environment variables to enable java debug options

testforstephen:

This PR provides a third-way to enable java debug options, and also a workaround for the bug https://github.com/Azure/vscode-kubernetes-tools/issues/83

### DIFF
--- a/src/debug/debugUtils.ts
+++ b/src/debug/debugUtils.ts
@@ -13,13 +13,13 @@ async function promptForPort(promptMessage: string, defaultPort: string): Promis
 }
 
 export async function promptForDebugPort(defaultPort: string): Promise<string> {
-    return await promptForPort("What debug port your app is enabled on?", defaultPort);
+    return await promptForPort("Please specify debug port exposed for debugging", defaultPort);
 }
 
 export async function promptForAppPort(ports: string[], defaultPort: string, env: {}): Promise<string> {
     let rawAppPortInfo: string;
     if (ports.length === 0) {
-        return await promptForPort("What port your app listens on?", defaultPort);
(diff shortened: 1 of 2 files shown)


src/debug/debugUtils.ts
```diff
@@ -2,30 +2,38 @@ import * as vscode from "vscode";
...
+        prompt: promptMessage,
+        placeHolder: defaultPort,
+        value: defaultPort
     });
 
     return input && input.trim();
 }
 
+export async function promptForDebugPort(defaultPort: string): Promise<string> {
+    return await promptForPort("What debug port your app is enabled on?", defaultPort);
```
itowlson: 'Enabled' reads oddly to me here.  I prefer your original "Please specify debug port exposed for debugging", though I guess it is a bit tautologous.  "Please specify the port for application debugging" perhaps?

src/debug/debugUtils.ts
```diff
@@ -2,30 +2,38 @@ import * as vscode from "vscode";
...
+}
+
 export async function promptForAppPort(ports: string[], defaultPort: string, env: {}): Promise<string> {
     let rawAppPortInfo: string;
-    if (ports.length === 1) {
+    if (ports.length === 0) {
+        return await promptForPort("What port your app listens on?", defaultPort);
```
itowlson: Grammar pickiness: "What port does your application listen on?"

src/debug/debugUtils.ts
testforstephen: Previous app port prompt message is a bit similar as the debugging port message, and easy to lead mistake. That's why i refine the message. Refining the app port prompt message is enough. Have reverted this message to the original.

PR #90, opened 2018-03-12
Narrow helm selectors

andxu:


package.json
```diff
@@ -460,9 +461,15 @@
                     "helm"
                 ],
                 "extensions": [
-                    ".yaml",
                     ".tpl"
                 ],
+                "filenamePatterns": [
+                    "values.yaml",
+                    "Chart.yaml",
+                    "values.yaml",
```
testforstephen: DUP

src/extension.ts
```diff
@@ -84,7 +84,7 @@ export async function activate(context) : Promise<extensionapi.ExtensionAPI> {
     const completionProvider = new HelmTemplateCompletionProvider();
     const completionFilter = [
         "helm",
-        {language: "yaml", pattern: "**/templates/*.yaml"},
+        // templates/*.yaml should be helm templates, so we needn't select them
```
itowlson: Do we need to keep this commented-out line?

package.json
```diff
@@ -56,7 +56,8 @@
         "onCommand:extension.draftCreate",
         "onCommand:extension.draftUp",
         "onView:extension.vsKubernetesExplorer",
-        "onLanguage:helm"
+        "onLanguage:helm",
+        "onLanguage:yaml"
```
itowlson: Do we need to activate on arbitrary YAML files?  Are we providing Helm intellisense on arbitrary YAML as well as on files that match the Helm language filter?

andxu: We need onLanguage:yaml because kubernetes  manifest files are yaml language, we will be activated to be able to call registerYamlSchemaSupport, otherwise, user will be very confuse because activating the extension or not will have different behaviors for kubernetes manifest files when sometimes he see the hoved document and sometimes not(because our extension is not activated )

andxu: for your question: Are we providing Helm intellisense on arbitrary YAML as well as on files that match the Helm language filter
on arbitrary YAML , if user has chosen yaml as the language, then we are providing Helm intellisense only when yaml is under `templates` folder(we can add more filters after detecting the helm/chart folder structure later). On the other side, we try to provide kubernetes manifest support (validation/auto-completion) only when we recognize it as kubernetes manifest support by kind and apiVersion

src/extension.ts
andxu: removed

Issue #80, opened 2018-03-09
Image published to docker hub without user consent


akaroml:
After running Kubernetes: Run, I found my image published to docker hub without asking. This makes private images public without users acknowledgment and can cause serious privacy issues. We need to ask users whether they agree to upload images to docker hub.

On the other hand, docker hub can be avoided. When minikube is used, local docker registry can be used to store the images without traveling all the way to docker hub.

itowlson:
This is a good point.  We need to work out the desired workflow here.  Is this a project setting for the container registry to use for getting the images onto the cluster?  Is project level or user level better?  Is a project likely to contain multiple workflows of this kind (e.g. dev-test on minikube via local registry, then integration test on cloud via ACR)?  If so how do we facilitate users switching between local and cloud workflows?

cc @squillace

itowlson:
Hmm, `vscode-docker` treats a blank registry path as Docker Hub, so doesn't _really_ have a way to express "use the local registry" (because it doesn't need to think about implicit pushes as a way of getting containers to a cluster).  That said, if neither `defaultRegistry` _nor `defaultRegistryPath` are provided, then we can reasonably treat that as local, because users won't have permission to push to Docker Hub without a path prefix.  (`vscode-docker`'s tagging code implicitly recognises this: in such as situation the tag will be unqualified.)

testforstephen:
About the user consent message, we could add a button `"Never Show Again"` to allow user to close it.

itowlson:
I would expect a one-off warning (or once per project) to be the behaviour.  If we want an "Ask me every time" option then that adds _further_ complexity.

itowlson:
The proposal is as follows:

* When you perform the _first_ push to a new cluster, it will ask you which registry you want to use with that cluster.
* For subsequent pushes to that cluster, it will continue to use the same registry.

Things to consider:

* How do we record which registry is associated with any given cluster?  Options:
  * Metadata in kubeconfig.  The `kubectl config set*` commands seem pretty hostile to this.
  * VS Code project or user settings.  This should be easy, but is a separate piece of info to cart around.  At project level it could be checked in, which is nice, but then the user would need to re-enter it for every project - is that a burden?
  * Store it on the cluster somehow.  I am not sure what options we have for annotations at the cluster level.
[...]

PR #93, opened 2018-03-12
Better error messages if docker push fails during Run

### LINKED ISSUE #77: Kubernetes: Run failed with no further direction
Tried Kubernetes: Run and got the notification:
screen shot 2018-03-09 at 11 07 09 am

And here's the message in Output window:
screen shot 2018-03-09 at 11 06 10 am

It would be nice if we can provide further directions to resolve the problem. I could not tell the cause from the error message.

itowlson:
When `docker push` failed during a Kubernetes Run command, we would just say "Image push failed" and dump standard error.  The Docker error often did not help users to understand what the problem might be.  This PR checks for known messages and suggests possible remediations (e.g. user id not configured, user not logged in to Docker).

Fixes #77.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -710,6 +710,18 @@ function runKubernetes() {
     });
 }
 
+function diagnosePushError(exitCode: number, error: string) : string {
+    if (error.includes("denied")) {
+        const user = vscode.workspace.getConfiguration().get("vsdocker.imageUser", null);
+        if (user) {
+            return "Failed to push to Docker Hub. Try running docker login.";
+        } else {
+            return "Failed to push to Docker Hub. Try setting vsdocker.imageUser.";
+        }
+    }
+    return 'Image push failed.';
+}
+
 function buildPushThenExec(fn) {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -721,9 +733,9 @@ function buildPushThenExec(fn) {
                         vscode.window.showInformationMessage(image + ' pushed.');
```
testforstephen: Before executing `push`, it's better to add user consent `This command needs to push your docker image to docker hub, do you want to continue?`

itowlson: Yeah, this has been raised as #80.  As noted there, this also gets us thinking about giving options for ACR etc., private registries and local registry (for minikube), and aligning with how vscode-docker handles this stuff.  Adding to the complexity, a consent prompt entails tracking that the user has consented so we don't bug them every time.  I'd like to restrict the scope of this PR to giving better errors in the existing failure cases, and tackle the broader issue of our interaction with registries in a subsequent one.

testforstephen: Ok to do it on the new PR.

PR #95, opened 2018-03-13
[WIP] Add initial draft debug to extension

radu-matei:


src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
...
+    // TODO - make ports parameters
+    const term = vscode.window.createTerminal('draft connect', `bash`, ['-c', `${draftPath} connect -p 8080:8080 -p 9229:9229 ; bash`]);
+    term.show(true);
+    
+    vscode.window.showInformationMessage("Attaching debugger...")
+
+    // we know there is a ready pod, this shoud be enough time for 
+    // the portforward stuff to create the connection 
+    // TODO - how can this be avoided? what happens if network is busy?
+    await sleep(5000)
```
radu-matei: This can probably be avoided by checking the ports at the OS level.

package.json
```diff
@@ -378,6 +381,24 @@
                 "description": "Create a new Draft project",
                 "category": "Draft"
             },
+            {
+                "command": "extension.draftCreateDebug",
+                "title": "Create (debug pack)",
```
itowlson: I think we usually title-case command names, e.g. `Create (Debug Pack)`.

package.json
```diff
@@ -378,6 +381,24 @@
...
+                "description": "Create a new Draft project using the Node Debug pack",
+                "category": "Draft"
+            },
+            {
+                "command": "extension.draftUpDebug",
+                "title": "up --debug",
```
itowlson: `Up (Debug)` perhaps?  We don't usually use the CLI command quite so literally. *grin*

package.json
```diff
@@ -378,6 +381,24 @@
...
+                "description": "Debug the Draft application",
+                "category": "Draft"
+            },
+            {
+                "command": "extension.draftDebug",
+                "title": "Debug",
+                "description": "Debug your Draft application",
```
itowlson: Description "Debug your Draft application" vs `draftUpDebug`'s "Debug the Draft application - hard to know what the difference is.

src/draft.ts
```diff
@@ -9,6 +9,7 @@ export interface Draft {
     isFolderMapped(path: string) : boolean;
     packs() : Promise<string[] | undefined>;
     invoke(args: string) : Promise<ShellResult>;
+    invokeAsync(command : string) : Promise<ShellResult>;
```
itowlson: `invoke` is already async (returns a Promise).  Do we truly need a new method here?  If so, can we choose a name that distinguishes it more clearly?  (And yes, this interface is inconsistent with the `kubectl` one, but that's historical, and `kubectl` will likely move towards purely async methods and drop the suffix.)

src/draft.ts
```diff
@@ -81,6 +86,11 @@ async function invoke(context : Context, args : string) : Promise<ShellResult> {
     }
 }
 
+async function invokeAsync(context : Context, command : string) : Promise<ShellResult> {
+    let cmd = context.binPath + ' ' + command;
```
itowlson: `const`

src/draft.ts
```diff
@@ -81,6 +86,11 @@ async function invoke(context : Context, args : string) : Promise<ShellResult> {
...
+    return await context.shell.exec(cmd);
```
itowlson: This looks the same as `invoke` except without the `checkPresent` call (which I feel we should keep...?).  It is written a bit differently (`return await exec()` rather than `result = exec(); return result;` but I think the differences turn out to be purely stylistic (except for the `checkPresent` one).

src/extension.ts
```diff
@@ -128,6 +129,10 @@ export function activate(context) {
         registerCommand('extension.draftCreate', execDraftCreate),
         registerCommand('extension.draftUp', execDraftUp),
 
+        registerCommand('extension.draftCreateDebug', execDraftCreateDebug),
```
itowlson: Suggest we keep all the Draft commands in one block, so remove blank line above?

src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
     }
 }
 
-function draftCreateResult(sr : ShellResult, hadPack : boolean) {
+async function execDraftCreateDebug() {
```
itowlson: Can we deduplicate across this and `execDraftCreate`?  It seems to be the same code and UI except for passing the `node-debug` pack instead of `undefined` in the last line.

src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
...
+    }
+    const proposedAppName = path.basename(vscode.workspace.rootPath);
+    const appName = await vscode.window.showInputBox({ value: proposedAppName, prompt: "Choose a name for the Helm release" });
+    if (appName) {
+        await execDraftCreateApp(appName, pack);
+    }
+}
+
+async function execDraftUpDebug() {
+    const draftPath = await draft.path();
```
itowlson: Unlike `execDraftUp`, this doesn't perform any checks that this is a Draft app.  Does it perform correctly if the user invokes it in a non-Drafted folder (or no folder)?  Or is it configured so you can't do that?

src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
...
+    const term = vscode.window.createTerminal('draft logs', `bash`, ['-c', `${draftPath} logs ; bash`]);
```
itowlson: Torn between 'this is okay because Draft is broken on Windows' and 'we should check the OS anyway so we don't get a nasty surprise when they fix it'.  *grin*  I think it's fine...

src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
...
+    term.show(true);
+
+    vscode.window.showInformationMessage("Starting draft up")
+    await draft.invokeAsync('up').then((result) => {
```
itowlson: Mixing of `await` and thenable-callback style looks odd to me.  (We have a few places where we do it with `kubectl`, but those are because the code is undergoing a long slow transition from thenables/callbacks to `async/await`.  I'd prefer not to propagate it further, unless of course there's good reason!)

src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
...
+
+}
+
+async function execDraftDebug() {
+
+    // exits with code 0 when it found a pod to connect to
+    // replicaCount is set to 1 in the debug pack, because with rolling updates, 
+    // it will find a running pod soon to be terminated, connect to it and then it gets killed
+    // TODO - fix this once helm install --wait is functional
+    await draft.invokeAsync('connect --wait')
```
itowlson: Should we display something while this is happening, or is it really quick?  We have code somewhere for displaying indicators in the status bar, or you could print something to the Output window.

src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
...
+    vscode.window.showInformationMessage("Starting draft up")
+    await draft.invokeAsync('up').then((result) => {
+        vscode.window.showInformationMessage("Finished draft up")
+        execDraftDebug() 
+    })
+
+}
+
+async function execDraftDebug() {
+
```
itowlson: Do we need to do the hygiene checks like 'is it a folder, is it a Draft app, etc.'?

src/extension.ts
```diff
@@ -1460,7 +1465,82 @@ async function execDraftCreateApp(appName : string, pack? : string) : Promise<vo
...
+    const debugConfiguration = {
+        type: 'node',
+        request: 'attach',
+        name: 'Attach to Process',
+        address: 'localhost',
+        // TODO - make this configurable - part of draft.toml?
+        port: 9229,
+        localRoot: vscode.workspace.rootPath,
+        remoteRoot: '/usr/src/app'
```
itowlson: Pardon my ignorance - is this directory part of the Draft spec, so guaranteed to work?

src/extension.ts
itowlson: Does this work on all versions of Draft or is there a minimum?  (Asking because I don't recall having seen it before, but I don't know much about Draft yet!)

radu-matei:
Closing this as it has become way out of date.

Thanks for all the feedback, will incorporate it in the new version coming soon 🎉

PR #101, opened 2018-03-16
Update kubernetes schema to version 1.9.3

andxu:


andxu:
The latest version 1.9.3 are compatible with old version with proper depreciated warning messages, I will first switch to this version and then load schema from user's environment/swagger api and then  notify user about the depreciated model/compatible issue between the yaml and the k8s server, this PR is the first one switch to version  1.9.3

PR #104, opened 2018-03-19
Code snippet for deployment and service for yaml language.

andxu:


src/yaml-support/yaml-snippet.ts
```diff
@@ -0,0 +1,55 @@
...
+        this._snippets  = fs.readdirSync(snippetRoot)
+            .filter((filename: string): boolean => filename.endsWith('-snippet.yaml'))
+            .map((filename: string): CodeSnippet =>
+                <CodeSnippet>yaml.safeLoad(fs.readFileSync(path.join(snippetRoot, filename), 'utf-8')));
+    }
+
+    // filter all internal code snippets using the parameter word
+    private _filterCodeSnippets(word: string): CodeSnippet[] {
+        return this._snippets.filter((snippet: CodeSnippet): boolean =>
+            snippet.name.startsWith(word.toLowerCase()));
```
testforstephen: Do we need fuzzy search?

src/yaml-support/yaml-snippet.ts
```diff
@@ -0,0 +1,55 @@
...
+            item.documentation = snippet.description;
+            return item;
+        });
+    }
+
+    // load yaml code snippets from ../../snippets folder
+    private _loadCodeSnippets(): void {
+        const snippetRoot = path.join(__dirname, '../../../snippets');
+        this._snippets  = fs.readdirSync(snippetRoot)
+            .filter((filename: string): boolean => filename.endsWith('-snippet.yaml'))
```
testforstephen: why need to filter `-snippet.yaml` files?

src/yaml-support/yaml-snippet.ts
andxu: vscode will do further fuzzy search on the results of this function, we can comment out the filter logic but in the future, there will be logic for deciding the list of snippets suitable at the given position, we'd better keep the search logic since later we will do the filtering against the position.

src/yaml-support/yaml-snippet.ts
andxu: shall we load arbitrary yaml files under snippets?

bnookala: @andxu might as well! if somebody chooses to add more snippets in the future, they wouldn't need to hunt through this code to identify why their file didn't load.

src/yaml-support/yaml-snippet.ts
```diff
@@ -0,0 +1,55 @@
...
+            .map((filename: string): CodeSnippet =>
+                <CodeSnippet>yaml.safeLoad(fs.readFileSync(path.join(snippetRoot, filename), 'utf-8')));
+    }
+
+    // filter all internal code snippets using the parameter word
+    private _filterCodeSnippets(word: string): CodeSnippet[] {
+        return this._snippets.filter((snippet: CodeSnippet): boolean =>
+            snippet.name.startsWith(word.toLowerCase()));
+
```
bnookala: remove this extra line

itowlson:
@andxu CI is failing on (gasp) a missing semicolon.  It's not a compiler error but we do have it in our lint rules - have you enabled tslint in VS Code?

src/extension.ts
```diff
@@ -202,7 +204,7 @@ export async function activate(context) : Promise<extensionapi.ExtensionAPI> {
     subscriptions.forEach((element) => {
         context.subscriptions.push(element);
     }, this);
-
+
```
itowlson: Is this diff indicating there's a spurious space on the blank line?  I can't quite tell.

src/yaml-support/yaml-snippet.ts
```diff
@@ -0,0 +1,55 @@
...
+        return this._filterCodeSnippets(word).map((snippet: CodeSnippet): vscode.CompletionItem =>  {
+            const item = new vscode.CompletionItem(snippet.label, vscode.CompletionItemKind.Snippet);
+            item.insertText = new vscode.SnippetString(snippet.body);
+            item.documentation = snippet.description;
+            return item;
+        });
+    }
+
+    // load yaml code snippets from ../../snippets folder
+    private _loadCodeSnippets(): void {
```
itowlson: We don't usually prefix private methods with `_` (although we often do for fields).

src/yaml-support/yaml-snippet.ts
```diff
@@ -0,0 +1,55 @@
...
+        const snippetRoot = path.join(__dirname, '../../../snippets');
+        this._snippets  = fs.readdirSync(snippetRoot)
+            .filter((filename: string): boolean => filename.endsWith('.yaml'))
+            .map((filename: string): CodeSnippet =>
+                <CodeSnippet>yaml.safeLoad(fs.readFileSync(path.join(snippetRoot, filename), 'utf-8')));
```
itowlson: Introduce helper method e.g. `readYamlCodeSnippet`.

snippets/deployment.yaml
```diff
@@ -0,0 +1,22 @@
+name: deployment
+label: Create Kubernetes Deployment
```
itowlson: Consider either "Kubernetes Deployment" or "Create a Deployment manifest".

snippets/deployment.yaml
```diff
@@ -0,0 +1,22 @@
+name: deployment
```
itowlson: How does the presence of these interact with the existing Helm snippets in `helm.json`?  They seem quite similar.

src/yaml-support/yaml-snippet.ts
```diff
@@ -0,0 +1,55 @@
...
+            const item = new vscode.CompletionItem(snippet.label, vscode.CompletionItemKind.Snippet);
+            item.insertText = new vscode.SnippetString(snippet.body);
+            item.documentation = snippet.description;
+            return item;
+        });
+    }
+
+    // load yaml code snippets from ../../snippets folder
+    private _loadCodeSnippets(): void {
+        const snippetRoot = path.join(__dirname, '../../../snippets');
```
itowlson: It feels a wee bit awkward to me that the YAML completion items are distinct files that share a directory with the JSON documents that contain snippet lists (and are automatically proferred as snippets).  I wonder if we should call the directory e.g. yaml-snippets?  What is the reason we can't just provide these in a JSON snippets file the way we do for the Helm snippets?  Is it worth documenting the reason?

andxu: No, they are separated yaml files, authoring code snippets itself is very easy to make mistakes and a lot of code snippets in one file is not suggested (vscode uses a large file snippet.json), I don't use snippet.json because I don't want to maintain the large file like:
```
"Deployment": {
    "prefix": "Deployment",
    "body": [
      "apiVersion: apps/v1beta1",
      "kind: Deployment",
      "metadata:",
      "  name: ${1:myapp}",
      "spec:",
      "  replicas: ${2|1,2,3|}",
      "  template:",
      "    metadata:",
      "      labels:",
      "        app: ${1:myapp}",
      "    spec:",
      "      containers:",
      "      - name: ${1:myapp}",
      "        image: ${5:<Image>}",
      "        ports:",
      "        - containerPort: ${6:<Port>}"
  ],
```
We have already made a folder named `rawsnippets` contains separated helm templates files(yaml&&helm format)

snippets/deployment.yaml
andxu: I think `Kubernetes Deployment` is better, during a arbitrary  yaml editing, Kubernetes will give user a definitive  hint about this snippet.

src/extension.ts
andxu: I will remove it.

src/yaml-support/yaml-snippet.ts
itowlson: So could we put these into a subdirectory of `rawsnippets` and use the same preprocessing tool as we do on the existing raw snippets?  (If we still have it... it might have been part of the old vs-helm repo.  If not, I can easily recreate it.)  I just find it a bit disconcerting that we will have two different ways of implementing snippets...

PR #120, opened 2018-03-20
Read the data from file  for the new configuration  certificate-authority instead of certificate-authority-data

andxu:


src/fs.ts
```diff
@@ -3,7 +3,7 @@ import * as sysfs from 'fs';
 export interface FS {
     existsSync(path : string | Buffer) : boolean;
     readFile(filename : string, encoding : string, callback : (err : NodeJS.ErrnoException, data : string) => void) : void;
-    readFileSync(filename : string, encoding : string) : string;
+    readFileSync(filename : string, encoding?: string) : string| Buffer;
```
itowlson: This signature feels hard to work with at a type level because the result will need a type assertion to be safe.  I suggest splitting out a separate member for reading raw buffers e.g. `readFileToBufferSync(filename : string) : Buffer` and leaving the existing function as it was.

src/kubeconfig.ts
```diff
@@ -39,12 +40,13 @@ export function readKubectlConfig() : Promise<KubeConfig> {
...
             const clientKeyData = currentUser['client-key-data'];
-
+            const clientCertDataFile = currentUser['client-certificate'];
+            const clientKeyFile = currentUser['client-key'];
             resolve({
                 endpoint: endpoint,
-                clientCertificateData: Buffer.from(clientCertData, 'base64'),
-                clientKeyData: Buffer.from(clientKeyData, 'base64'),
-                certificateAuthorityData: Buffer.from(cadata, 'base64')
+                clientCertificateData: clientCertDataFile ? <Buffer>fs.readFileSync(clientCertDataFile): Buffer.from(clientCertData, 'base64'),
```
itowlson: TS docs seem to now prefer `foo as Buffer` to `<Buffer>foo`, though if we have two distinct members in the interface then we can get rid of the type assertion altogether.

src/fs.ts
andxu: I advise to use a dedicate fs extra library which is compatible with node.js fs lib, see  https://github.com/jprichardson/node-fs-extra

PR #121, opened 2018-03-20
generate markdown document for kubernetes yaml 

andxu:


src/yaml-support/yaml-util.ts
```diff
@@ -122,4 +122,39 @@ function getStringValue(node, key: string, ignoreCase: StringComparison = String
             }
         }
     }
-}
\ No newline at end of file
+}
+
+// copied from explainer.ts, used by explainer.ts, yaml-schema.ts
```
itowlson: I'd avoid putting these in yaml-util as they are also used for JSON files (Explain works on JSON as well as YAML).  I think `explainer.ts` is still a reasonable place for them because that module is meant to be a reasonably pure 'transform Swagger to Markdown' module and reusable outside the hover provider; but if you're not comfortable with that then perhaps create a new module called something like `schema-formatting`?  In which case you could arguably rename these functions to e.g. `formatOne`, `formatComplex`, etc.

src/yaml-support/yaml-schema.ts
```diff
@@ -116,6 +138,23 @@ class KubernetesSchemaHolder {
             this._definitions[schema.id.toLowerCase()] = schema;
         }
     }
+
+    // get the markdown format of document for the current property and the type of current property
+    // see similar function chaseFieldPath in expaliner.ts
```
itowlson: Would it make sense to try to unify this with `chaseFieldPath` or something at that level, so we don't need to duplicate the logic, or export the low-level details of `explainOne` and `explainComplex`?  Seems like we should be able to have the explainer module have just one or two high-level entry points and hide all the complexity.  But perhaps that's for a later refactoring.

src/yaml-support/yaml-util.ts
itowlson: As these have been moved, you don't really need to describe the history -- that's what git logs are for.  (If you were duplicating them, then a TODO note to reunify the two copies would be useful - but I think you have kept a single implementation which is much better.)

src/yaml-support/yaml-schema.ts
andxu: explain command will be refactored for explain the user selection text or user input later

andxu:
using this PR instead https://github.com/Azure/vscode-kubernetes-tools/pull/132

PR #123, opened 2018-03-20
Meaningful error if user tries to create Helm chart when no folder open

### LINKED ISSUE #119: Command "Helm: Create chart" should pop out error when there is no open folder.
when there is no open folder, run command "Helm: Create chart", there is no error pop out.
But we can find error log in developer tools. 
image

Suggest to pop out an error when there is no open folder.

itowlson:
Fixes #119.

### DIFF
--- a/src/helm.exec.ts
+++ b/src/helm.exec.ts
@@ -6,6 +6,7 @@ import * as YAML from 'yamljs';
 import * as _ from 'lodash';
 import * as fs from "fs";
 import * as helm from './helm';
+import { showWorkspaceFolderPick } from './hostutils';
 
 export interface PickChartUIOptions {
     readonly warnIfNoCharts : boolean;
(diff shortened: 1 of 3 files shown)


src/helm.exec.ts
```diff
@@ -76,6 +76,10 @@ export function helmDepUp() {
 }
 
 export function helmCreate() {
+    if (!vscode.workspace.rootPath) {
```
testforstephen: The API `vscode.workspace.rootPath` is [deprecated](https://github.com/Microsoft/vscode/issues/39625) since vscode added the multiple workspace support feature. Need use `vscode.workspace.workspaceFolders` instead. And we have similar workspace picking up or empty check issue for many commands, maybe we could add an utility method for workspace check/pickup.

At the `extension.ts` file, i have implemented a method for that, suggest to move it to `extensionUtils.ts` file for sharing.

```
async function showWorkspaceFolderPick(): Promise<vscode.WorkspaceFolder> {
    if (!vscode.workspace.workspaceFolders) {
        vscode.window.showErrorMessage('This command requires an open folder.');
        return undefined;
    } else if (vscode.workspace.workspaceFolders.length === 1) {
        return vscode.workspace.workspaceFolders[0];
    }
    return await vscode.window.showWorkspaceFolderPick();
}
```

itowlson: Perfect, done and ready for re-review.

PR #124, opened 2018-03-20
After adding cluster, refresh Explorer to show it

### LINKED ISSUE #115: after configure cluster ,still need to click refresh button to get the new cluster
Repro steps:
1. click configure from Cluster
2. select the sb and the cluster you want to configure
3. wait configure complete

Expect: the new cluster appear in the cluster list
Actual: we need to manually refresh the list to see the new cluster

itowlson:
Fixes #115.

### DIFF
--- a/src/components/clusterprovider/azure/azureclusterprovider.ts
+++ b/src/components/clusterprovider/azure/azureclusterprovider.ts
@@ -1,3 +1,4 @@
+import * as vscode from 'vscode';
 import * as restify from 'restify';
 import * as portfinder from 'portfinder';
 import * as clusterproviderregistry from '../clusterproviderregistry';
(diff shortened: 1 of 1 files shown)


src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -446,3 +450,7 @@ function parseCluster(encoded: string) : azure.ClusterInfo {
         name: encoded.substr(delimiterPos + 1)
     };
 }
+
+async function refreshExplorer() : Promise<void> {
```
andxu: remove async and replace await with return

itowlson: Hmm, as this is conceptually `void`, I kind of find the imperative `await` more natural than a `return` (though I know it is indeed actually returning a Promise/Thenable object rather than being truly void).  Would prefer to leave it unless you feel strongly.

PR #129, opened 2018-03-21
Add a user setting switch to control whether to auto clean up the created debug deployment

testforstephen:

Sometimes the debugger cannot be launched successfully, and auto cleanup mechanism will cause the user to have no chance to show the logs of the pod to diagnose what happens.  Add a switch to disable the cleanup policy for failure diagnostics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Where `<your-image-prefix-here>` is something like `docker.io/brendanburns`.
        * `vs-kubernetes.namespace` - The namespace to use for all commands
        * `vs-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
        * `vs-kubernetes.draft-path` - File path to the draft binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
+       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. Default to `true`.
    * `vsdocker.imageUser` - Image prefix for docker images e.g. 'docker.io/brendanburns'
 
 ## Known Issues
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Where `<your-image-prefix-here>` is something like `docker.io/brendanburns`.
        * `vs-kubernetes.namespace` - The namespace to use for all commands
        * `vs-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
        * `vs-kubernetes.draft-path` - File path to the draft binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
-       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for `"Clean Up"` or not.
+       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created deployment and associated pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up if you wanted to view pod logs, etc.
    * `vsdocker.imageUser` - Image prefix for docker images e.g. 'docker.io/brendanburns'
 
 ## Known Issues
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Where `<your-image-prefix-here>` is something like `docker.io/brendanburns`.
        * `vs-kubernetes.namespace` - The namespace to use for all commands
        * `vs-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
        * `vs-kubernetes.draft-path` - File path to the draft binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
-       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. Default to `true`.
+       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for `"Clean Up"` or not.
    * `vsdocker.imageUser` - Image prefix for docker images e.g. 'docker.io/brendanburns'
 
 ## Known Issues
--- a/debug-on-kubernetes.md
+++ b/debug-on-kubernetes.md
@@ -12,7 +12,7 @@ One of the key features of VS Code Kubernetes Extension is its one-click debuggi
 
 ## 3. Extension Settings for debugging
    * `vs-kubernetes` - Parent for Kubernetes-related extension settings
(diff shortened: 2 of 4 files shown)


src/debug/debugSession.ts
```diff
@@ -270,20 +270,26 @@ export class DebugSession implements IDebugSession {
...
+                kubeChannel.showOutput(`Kubectl command failed: ${deleteResult.stderr}`);
+                return;
+            } else {
+                kubeChannel.showOutput(`Resource ${resourceId} is removed successfully.`);
+                vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
+            }
         } else {
-            kubeChannel.showOutput(`Resource ${resourceId} is removed successfully.`);
-            vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
+            kubeChannel.showOutput(`"vs-kubernetes.autoCleanupOnDebugTerminate" flag is disabled, won't clean up the resource "${resourceId}" automatically. Please do that manually.`);
```
itowlson: Suggestion: instead of this, put up a prompt saying something "Resource ${resourceId} will continue running on the cluster." with a "Clean Up" button.  If they choose Close then it leaves stuff there and they can party on it then clean it up by hand.  If they choose Clean Up then go into the standard clean up routine.

package.json
```diff
@@ -83,7 +83,18 @@
...
                             "default": null,
                             "description": "File path to a draft binary."
+                        },
+                        "vs-kubernetes.autoCleanupOnDebugTerminate": {
+                            "type": "boolean",
+                            "default": true,
+                            "description": "Once the debug session is terminated, automatically clean up the created Deployment and associated Pod by the command \"Kubernetes: Debug (Launch)\"."
                         }
+                    },
+                    "default": {
```
itowlson: What does this do?  How do we keep it in sync with the `default` value in each contributed config entry?

src/debug/debugSession.ts
itowlson: If you felt fancy you could add an 'Always Clean Up' button which added the config entry.  Or a 'Learn More' which linked to a relevant part of the debug documentation.

README.md
```diff
@@ -129,6 +129,7 @@ Where `<your-image-prefix-here>` is something like `docker.io/brendanburns`.
        * `vs-kubernetes.namespace` - The namespace to use for all commands
        * `vs-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
        * `vs-kubernetes.draft-path` - File path to the draft binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
+       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. Default to `true`.
```
itowlson: We should also add this to the debug documentation.

src/debug/debugSession.ts
testforstephen: Nice, i like the suggestion. Thanks.

package.json
testforstephen: > What does this do?

Because we wrap a parent level for `vs-kubernetes.xxx`, and vscode doesn't show the details in `DEFAULT SETTINGS` view. Adding an explicit `default` field for the parent level `vs-kubernetes` will help list each config items in `DEFAULT SETTINGS` view. See the snapshot below.
image

> How do we keep it in sync with the default value in each contributed config entry?

I did some experiment for this. The `default` value for the second level settings `vs-kubernetes.xxx` doesn't take any effect if i leave user setting to be empty, they all are `undefined` from the VS Code API `vscode.workspace.getConfiguration("vs-kubernetes")`. It seems only the first level `default` value has impact and shows up in `DEFAULT SETTINGS` view.

I'm going to remove the `default` value definition from each config entry, and put them in the parent level together.

README.md
```diff
@@ -129,6 +129,7 @@ Where `<your-image-prefix-here>` is something like `docker.io/brendanburns`.
        * `vs-kubernetes.namespace` - The namespace to use for all commands
        * `vs-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
        * `vs-kubernetes.draft-path` - File path to the draft binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
+       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for `"Clean Up"` or not.
```
itowlson: For last sentence suggest: "If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up in if you wanted to view pod logs, etc."

debug-on-kubernetes.md
```diff
@@ -10,8 +10,13 @@ One of the key features of VS Code Kubernetes Extension is its one-click debuggi
    * `Kubernetes: Debug (Launch)` - Run the current application as a Kubernetes Deployment and attach a debugging session to it (currently works only for Java/Node.js deployments)
    * `Kubernetes: Debug (Attach)` - Attach a debugging session to an existing Kubernetes Deployment (currently works only for Java deployments)
 
-## 3. How to use it for Java debugging
-### 3.1 Launch a Spring Boot application on Kubernetes and debug it
+## 3. Extension Settings for debugging
+   * `vs-kubernetes` - Parent for Kubernetes-related extension settings
+      * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for `"Clean Up"` or not.
```
itowlson: Same wording suggestion

package.json
```diff
@@ -71,19 +71,26 @@
...
+                        "vs-kubernetes.autoCleanupOnDebugTerminate": {
+                            "type": "boolean",
+                            "description": "Once the debug session is terminated, automatically clean up the created Deployment and associated Pod by the command \"Kubernetes: Debug (Launch)\"."
                         }
+                    },
+                    "default": {
+                        "vs-kubernetes.namespace": "",
+                        "vs-kubernetes.kubectl-path": "",
+                        "vs-kubernetes.draft-path": "",
+                        "vs-kubernetes.autoCleanupOnDebugTerminate": false
```
itowlson: I like the combination of defaulting to false and offering an Always Clean Up option.  That makes the option nicely discoverable and easy to get to the 'seamless' experience.

src/debug/debugSession.ts
```diff
@@ -270,11 +270,33 @@ export class DebugSession implements IDebugSession {
...
+            return await this.cleanupResource(resourceId);
+        }
+        const answer = await vscode.window.showWarningMessage(`Resource ${resourceId} will continue running on the cluster.`, "Clean Up", "Always Clean Up");
+        if (answer === "Clean Up") {
+            return await this.cleanupResource(resourceId);
+        } else if (answer === "Always Clean Up") {
+            // The object returned by VS Code API is readonly, clone it first.
+            const oldConfig = vscode.workspace.getConfiguration("vs-kubernetes");
+            const newConfig = {};
+            for (const key of Object.keys(oldConfig)) {
```
itowlson: Can this not be done using `Object.assign`?  (Asking because I need something similar, and was using `assign` - if this way is better then I should copy it!)

src/debug/debugSession.ts
```diff
@@ -270,11 +270,33 @@ export class DebugSession implements IDebugSession {
...
+                newConfig[key] = oldConfig[key];
+            }
+            newConfig["vs-kubernetes.autoCleanupOnDebugTerminate"] = true;
+            await vscode.workspace.getConfiguration().update("vs-kubernetes", newConfig);
```
itowlson: I believe this will set this option for the workspace, not the user.  Is that desired?

src/debug/debugSession.ts
testforstephen: Nice, use `Object.assign` is better.

src/debug/debugSession.ts
itowlson: If we feel user level is better I have code in progress for that on another branch: https://github.com/itowlson/vscode-kubernetes-tools/blob/98131ec800c3170054f935e070aae3221007e3ac/src/components/installer/installer.ts#L118.  Feel free to borrow/adapt!

testforstephen: Yes, it will be saved in `workspace settings.json`.  Use `update` api will save it in `Global` or `Workspace`, it's a single answer question. The ideal api is to check if workspace settings.json exists or not, if so, save in workspace, otherwise, global. But i didn't find such API. Let me check if there is other way to achieve this.

itowlson: Have a look at `inspect` - I think that does what you want.

testforstephen: Thanks to point that out. Have changed the update policy to make sure it will take effect under any situation.
* Always update global settings
* If workspaceFolder value exists, update it
* If workspace value exists, update it

PR #132, opened 2018-03-22
Generate detailed markdown document for kubernetes schema description

andxu:
Signed-off-by: andxu <andxu@microsoft.com>


src/explainer.ts
```diff
@@ -3,6 +3,7 @@
 import * as k8s from 'k8s';
 import * as pluralize from 'pluralize';
 import * as kubeconfig from './kubeconfig';
+import { formatComplex, formatOne, Typed, formatType } from "./yaml-support/schema-formatting";
```
itowlson: Again, this shouldn't mention `yaml`.  These functions are about formatting Swagger objects as Markdown, not about the file format the user is hovering over.  They are used as part of the explain function for JSON files too.  Why not leave them in the `explainer` module and have the YAML intellisense import that?  Or put the `schema-formatting` module at the top level rather than under `yaml-support`?

PR #133, opened 2018-03-22
Add Copy command to copy the resource id from tree explorer to clipboard

testforstephen:

Add `Copy` menu on Kubernetes Explorer. It will copy the resource id to the system clipboard. These menus are only visible for Kubernetes Resource (cluster/namespace/node/deployment/job/pod/service/configMap/secret).

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,11 @@
                     "group": "0@2"
                 },
                 {
+                    "command": "extension.vsKubernetesCopy",
+                    "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.cluster.inactive",
+                    "group": "1"
+                },
+                {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -375,6 +409,11 @@
                 "title": "Open Dashboard",
                 "category": "Kubernetes"
             },
+            {
+                "command": "extension.vsKubernetesCopy",
+                "title": "Copy",
```
itowlson: Should this perhaps be Copy Resource ID (or whatever k8s calls that kind/name string)?  Copy implies I am copying the resource itself...

testforstephen: In k8s schema, it called `Name`. So changed it to `Copy Name`

Issue #118, opened 2018-03-20
[Feature Request] Install dependencies automatically when installing Kubernetes extension


kyliel:
Could below dependencies be installed automatically if they are not there when installing Kubernetes extension at the first time? 

	docker
	kubectl
	helm
	draft

itowlson:
MVP is Linux.  OSX should be fairly similar I believe, though Docker install may be substantially different.

Installing Docker on Windows is a bit ugh (well, it's a bit ugh on Linux too but at least _scriptably_ ugh).  The command line tools should all be xcopy deployable, except that Windows doesn't have anything built in for unpacking `.tar.gz` files (though there appear to be NPM packages that do this, which we can certainly hope behave correctly on Windows).

If we exclude Docker from this requirement then it gets quite a bit easier...!

itowlson:
Well blister my barnacles, it looks like the npm tar module might just work...

squillace:
if docker is a blocker, let that be the only installation requirement on the part of the user. :-) help with what we can!

PR #136, opened 2018-03-26
Automatically install missing kubectl, Helm and Draft

itowlson:
This adds an option to the 'kubectl/helm/draft not found' error to install the extension's dependencies -- currently limited to `kubectl`, Helm and Draft.  (We'd like to do Docker and Azure CLI too, but that was a bit much for this PR.)

The experience is that if we detect a missing command line tool (via background e.g. populating the tree view or a command e.g. Helm Create), we pop an alert which now has the addition button "Install dependencies."  If the user clicks this button then we install kubectl, Helm and Draft into a `.vs-kubernetes/tools` directory under their home directory, and set VS Code config settings to point to these private copies.  We do **not** add the binaries to the system path, so if the user tries to run `kubectl` or `helm` at a terminal they will still get 'not found'.  Is this okay?  As a future consideration we should probably also support versioning: currently as long as we find the CLI tools we need, we don't check that they are up to date.

The experience still has some rough edges, and the code is quite untidy.  We should streamline the way we locate and run binaries, and also look at error reporting given that several things may be trying to locate `kubectl` at the same time.  However, I think it's usable, and I'd rather get people playing with it sooner rather than later.  (But would appreciate reviewers giving it a sanity check first!)

**BIG CAVEAT:** It has NOT been tested on Mac.  (Because I don't have access to a Mac.)  It Works On My Machine TM on both Windows and Linux (Ubuntu), though lacks a thorough test matrix.

@squillace @bacongobbler @technosophos or anyone else with a Mac: Could one of you fine people please test on Mac?  I don't have the facilities...

### DIFF
--- a/src/binutil.ts
+++ b/src/binutil.ts
@@ -6,6 +6,7 @@ export interface BinCheckContext {
     readonly host : Host;
     readonly fs : FS;
     readonly shell : Shell;
+    readonly installDependenciesCallback: () => void;
     binFound : boolean;
     binPath : string;
 }
@@ -48,31 +49,43 @@ export function execPath(shell : Shell, basePath : string) : string {
 
 type CheckPresentFailureReason = 'inferFailed' | 'configuredFileMissing';
 
-function alertNoBin(host : Host, binName : string, failureReason : CheckPresentFailureReason, message : string) : void {
+function alertNoBin(host : Host, binName : string, failureReason : CheckPresentFailureReason, message : string, installDependencies : () => void) : void {
     switch (failureReason) {
         case 'inferFailed':
-            host.showErrorMessage(message, 'Learn more').then(
+            host.showErrorMessage(message, 'Install dependencies', 'Learn more').then(
                 (str) => {
-                    if (str !== 'Learn more') {
-                        return;
+                    switch (str) {
+                        case 'Learn more':
+                            host.showInformationMessage(`Add ${binName} directory to path, or set "vs-kubernetes.${binName}-path" config to ${binName} binary.`);
+                            break;
+                        case 'Install dependencies':
+                            installDependencies();
+                            break;
                     }
 
-                    host.showInformationMessage(`Add ${binName} directory to path, or set "vs-kubernetes.${binName}-path" config to ${binName} binary.`);
                 }
             );
             break;
         case 'configuredFileMissing':
-            host.showErrorMessage(message);
+            host.showErrorMessage(message, 'Install dependencies').then(
+                (str) => {
+                    if (str === 'Install dependencies') {
+                        installDependencies();
+                    }
+                }
+            );
             break;
     }
 }
 
-export async function checkForBinary(context : BinCheckContext, bin : string, binName : string, inferFailedMessage : string, configuredFileMissingMessage : string) : Promise<boolean> {
+export async function checkForBinary(context : BinCheckContext, bin : string, binName : string, inferFailedMessage : string, configuredFileMissingMessage : string, alertOnFail : boolean) : Promise<boolean> {
     if (!bin) {
         const fb = await findBinary(context.shell, binName);
 
         if (fb.err || fb.output.length === 0) {
-            alertNoBin(context.host, binName, 'inferFailed', inferFailedMessage);
+            if (alertOnFail) {
+                alertNoBin(context.host, binName, 'inferFailed', inferFailedMessage, context.installDependenciesCallback);
+            }
             return false;
         }
 
@@ -86,7 +99,9 @@ export async function checkForBinary(context : BinCheckContext, bin : string, bi
     if (context.binFound) {
         context.binPath = bin;
     } else {
-        alertNoBin(context.host, binName, 'configuredFileMissing', configuredFileMissingMessage);
+        if (alertOnFail) {
+            alertNoBin(context.host, binName, 'configuredFileMissing', configuredFileMissingMessage, context.installDependenciesCallback);
+        }
     }
 
     return context.binFound;
--- a/src/components/installer/installer.ts
+++ b/src/components/installer/installer.ts
@@ -1,0 +1,177 @@
+'use strict';
+
+import * as download from 'download';
+import * as fs from 'fs';
+import mkdirp = require('mkdirp');
+import * as path from 'path';
+import * as tar from 'tar';
+import * as tmp from 'tmp';
+import * as vscode from 'vscode';
+import { Shell, Platform } from '../../shell';
+import { Errorable } from '../../wizard';
+import { exec } from 'child_process';
+
+export async function installKubectl(shell: Shell) : Promise<Errorable<void>> {
+    const tool = 'kubectl';
+    const binFile = (shell.isUnix()) ? 'kubectl' : 'kubectl.exe';
+    const os = platformUrlString(shell.platform());
+
+    const version = await getStableKubectlVersion();
+    if (!version.succeeded) {
+        return { succeeded: false, result: null, error: version.error };
+    }
+
+    const installFolder = getInstallFolder(shell, tool);
+    mkdirp.sync(installFolder);
+    
+    const kubectlUrl = `https://storage.googleapis.com/kubernetes-release/release/${version.result.trim()}/bin/${os}/amd64/${binFile}`;
+    const downloadFile = path.join(installFolder, binFile);
+    const downloadResult = await downloadTo(kubectlUrl, downloadFile);
+    if (!downloadResult.succeeded) {
+        return { succeeded: false, result: null, error: [`Failed to download kubectl: ${downloadResult.error[0]}`] };
+    }
+
+    if (shell.isUnix()) {
+        fs.chmodSync(downloadFile, '0777');
+    }
+
+    await addPathToConfig(`vs-kubernetes.${tool}-path`, downloadFile);
+    return { succeeded: true, result: null, error: [] };
+}
+
+async function getStableKubectlVersion() : Promise<Errorable<string>> {
+    const downloadResult = await downloadToTempFile('https://storage.googleapis.com/kubernetes-release/release/stable.txt');
+    if (!downloadResult.succeeded) {
+        return { succeeded: false, result: '', error: [`Failed to establish kubectl stable version: ${downloadResult.error[0]}`] };
+    }
+    const version = fs.readFileSync(downloadResult.result, 'utf-8');
+    fs.unlinkSync(downloadResult.result);
+    return { succeeded: true, result: version, error: [] };
+}
+
+export async function installHelm(shell: Shell) : Promise<Errorable<void>> {
+    const tool = 'helm';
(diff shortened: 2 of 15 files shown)


itowlson:
This relates to #118 but is not a complete fix.

src/components/installer/installer.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export async function installKubectl(shell: Shell) : Promise<Errorable<void>> {
+    const tool = 'kubectl';
+    const installFolder = getInstallFolder(shell, tool);
+    await mkdir(installFolder);
+    switch (shell.platform()) {
+        case Platform.Windows:
+        {
+            // Instruction say always to use this link for Windows, not to check the version first
+            const kubectlUrl = 'https://storage.googleapis.com/kubernetes-release/release/v1.9.0/bin/windows/amd64/kubectl.exe';
```
testforstephen: It seems this url will just download `v1.9.0`, but if you change it to the latest stable version `v1.9.6`, it will download the latest version.

src/components/installer/installer.ts
```diff
@@ -0,0 +1,215 @@
...
+        return { succeeded: false, result: '', error: [`Failed to establish kubectl stable version: ${downloadResult.error[0]}`] };
+    }
+    const version = fs.readFileSync(downloadResult.result, 'utf-8');
+    fs.unlinkSync(downloadResult.result);
+    return { succeeded: true, result: version, error: [] };
+}
+
+export async function installHelm(shell: Shell) : Promise<Errorable<void>> {
+    const tool = 'helm';
+    const executable = formatBin(tool, shell.platform());
```
testforstephen: Unused variable `executable`

bnookala:
I've been doing all this dev on a mac, so I'll take a look for you.

src/components/installer/installer.ts
```diff
@@ -0,0 +1,215 @@
...
+    const urlTemplate = 'https://storage.googleapis.com/kubernetes-helm/helm-v2.8.2-{os_placeholder}-amd64.tar.gz';
```
technosophos: At some point in the future, we might be able to use the GitHub API to find the latest release tag for the release. Definitely not necessary for this PR, though

src/components/installer/installer.ts
```diff
@@ -0,0 +1,215 @@
...
+    return await installToolFromTar(tool, urlTemplate, shell);
+}
+
+export async function installDraft(shell: Shell) : Promise<Errorable<void>> {
+    const tool = 'draft';
+    const urlTemplate = 'https://azuredraft.blob.core.windows.net/draft/draft-v0.11.0-{os_placeholder}-amd64.tar.gz';
```
bacongobbler: I'd personally hold off until we release v0.12.0 later this week, but it's your call. There is one BIG change to the architecture occurring in that release that might be worth holding off for a few days. :)

https://github.com/Azure/draft/pull/573

If you don't want to wait on that release dropping then I can follow up and PR a v0.12.0 version bump when it's released

itowlson: That makes sense, though I'll be away by then.  Depending on how the review goes I might merge this as is, and then do the much smaller change of updating the version while I'm on vacation.  Thanks for the heads up -- does this also affect the change I made to `Draft: Version` to pass the `-c` flag?  (I added that because installing Draft then immediately doing `Draft: Version` resulted in errors trying to get the server version.)

bacongobbler: yes. the -c flag on `draft version` was removed in v0.12.0 because there's only the client version:

```
><> draft version -c
Error: unknown shorthand flag: 'c' in -c
```

src/components/installer/installer.ts
```diff
@@ -0,0 +1,215 @@
+'use strict';
+
+import * as fs from 'fs';
+import * as http from 'http';
+import mkdirp = require('mkdirp');
```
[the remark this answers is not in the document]
itowlson: I don't think so.  `mkdirp` seems to use a form of `export` (the export is the function itself, not an object with a function property) that doesn't play nicely with TypeScript `import`.  `import ... = require('...');` is the recommended TypeScript way of acquiring such exports.

src/components/installer/installer.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    let newValue : any = {};
+    if (valueAtScope) {
+        newValue = Object.assign({}, valueAtScope);
+    }
+    newValue[configKey] = value;
+    await vscode.workspace.getConfiguration().update("vs-kubernetes", newValue, scope);
+}
+
+async function mkdir(path: string) : Promise<{}> {
```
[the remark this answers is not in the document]
itowlson: Nice find - thank you!

src/binutil.ts
```diff
@@ -48,31 +49,43 @@ export function execPath(shell : Shell, basePath : string) : string {
 
 type CheckPresentFailureReason = 'inferFailed' | 'configuredFileMissing';
 
-function alertNoBin(host : Host, binName : string, failureReason : CheckPresentFailureReason, message : string) : void {
+function alertNoBin(host : Host, binName : string, failureReason : CheckPresentFailureReason, message : string, installDependencies : () => void) : void {
```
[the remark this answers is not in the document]
itowlson: I like calling it `installDependencies` here because then I can write `installDependencies();` to invoke it, which I find more readable than `installDependenciesCallback();`.  But I think in the context structure it's better to identify it as a callback.  I take your point about the inconsistency though.

src/components/installer/installer.ts
itowlson: Hmm, I took the URL from the k8s installation documentation.  Agree that latest stable would be better.

src/components/installer/installer.ts
```diff
@@ -0,0 +1,215 @@
...
+    // add path to config
+    const executableFullPath = path.join(destinationFolder, executablePath);
+    await addPathToConfig(configKey, executableFullPath);
+
+    await fs.unlink(tarfile);
+
+    return { succeeded: true, result: null, error: [] };
+}
+
+async function download(sourceUrl: string) : Promise<Errorable<string>> {
```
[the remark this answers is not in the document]
itowlson: Good catch - that's definitely simpler to work with.  Thanks!

src/components/installer/installer.ts
```diff
@@ -0,0 +1,187 @@
...
+            }
+            fs.chmodSync(downloadFile, '0777');
+            addPathToConfig(`vs-kubernetes.${tool}-path`, downloadFile);
+            return { succeeded: true, result: null, error: [] };
+        }
+    }
+}
+
+async function getStableKubectlVersion() : Promise<Errorable<string>> {
+    const downloadResult = await downloadToTempFile('https://storage.googleapis.com/kubernetes-release/release/stable.txt');
```
bnookala: why not just make a http request against this to read the version?

itowlson: `<confession>` I was being lazy because I had already written the 'download to file' code and was struggling with the documentation to get the response body directly. `</confession>`

I've now ditched `request` in favour of `download` so will probably end up sticking with this approach though I agree it is entirely superfluous going via the filesystem!

src/components/installer/installer.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export async function installKubectl(shell: Shell) : Promise<Errorable<void>> {
+    const tool = 'kubectl';
+    const installFolder = getInstallFolder(shell, tool);
+    mkdirp.sync(installFolder);
+    switch (shell.platform()) {
+        case Platform.Windows:
+        {
+            // Instruction say always to use this link for Windows, not to check the version first
+            const kubectlUrl = 'https://storage.googleapis.com/kubernetes-release/release/v1.9.6/bin/windows/amd64/kubectl.exe';
```
testforstephen: See the document. https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl

The template `https://storage.googleapis.com/kubernetes-release/release/${version}/bin/${os}/amd64/kubectl.exe`  is also applied to windows. I think we need do that like what linux and mac did.

itowlson: Darn, you're right.  Thanks for spotting!

itowlson:
@bnookala Any word on whether it works on Mac?  If it does I think I can merge it.  (I did some more fixes and somewhat bungled the history, so please give it a quick spin on the latest commit if you have time.)

bacongobbler:
As someone new to testing dev releases of VSCode extensions, what's the best source of documentation to test this sucker?

itowlson:
@bacongobbler As far as I know the only way to test a PR is to add a remote for the source repo, and pull from that:

`git remote add itowlson https://github.com/itowlson/vscode-kubernetes-tools.git`
`git fetch itowlson`

then check out the `auto-install` branch.

Once you've got the right branch, to run and test it:

* `cd` to the working copy directory
* `npm install`
* Open the folder in Code (`code .`)
* Run the extension under the debugger (`F5`) - this launches a second instance of Code with the extension pseudo-installed from the local folder

bacongobbler:
Haven't manualls tested this yet, but the code and approach looks good to me from Draft's end!

bacongobbler:
When testing on a new system, I found this error when testing under debug, which appears unrelated to the changes applied here:

CTRL+P + "Helm: Lint"

> Extension `ms-kubernetes-tools.vscode-kubernetes-tools` failed to activate. Reason: unknown dependency `redhat.vscode-yaml`
> command 'extension.helmLint' not found

But this dependency wasn't required for linting (e.g. when creating a chart):

CTRL+P + "Helm: Create Chart"

> command 'extension.helmCreate' not found

I couldn't figure out how to install this extension to test this PR, but it appears that the right calls are being made :)

bacongobbler:
Ah nvm, I eventually found the extension by searching for "vscode-kubernetes-tools" and looking under the "Dependencies" toolbar. Just couldn't find the redhat yaml extension through searching for `redhat.vscode-yaml`. Noob problems :)

Now I got it working and things are working!

CTRL+P + "Draft: Version"

```
&version.Version{SemVer:"v0.12.0-rc3", GitCommit:"008eab68afbec3919f692f8c3747f6162abb704b", GitTreeState:"clean"}
```

Opening one of the example apps bootstrapped the app as expected. NICE!

2018-03-28-121731_3840x2160_scrot

bacongobbler:
However, this didn't appear to work for me on Arch Linux. When I removed the `draft` binary from my path, there was no action item in VSCode to download v0.12.0. Is this expected behaviour for Linux users?

2018-03-28-122612_3840x2160_scrot

squillace:
@itowlson @bnookala Looks good to me so far; I'll install on ubuntu tomorrow and see whether I can repro @bacongobbler's experience.

itowlson:
@bacongobbler @squillace Any word on the Ubuntu experience?  I grepped the code and any message displaying the "Learn more" option should now also be showing "Install dependencies".  I can't see any path that would give the prompt Matt encountered on Arch Linux.

PR #149, opened 2018-03-30
adding kubectl port-forward

### LINKED ISSUE #137: Add port-forwarding via tree view
Requested by `eugene_dev` via Marketplace review.

bnookala:
port-forward

Adds kubectl port forward as an invokable command and on the extension tree view. Fixes #137. Demo above. 

@squillace @akaroml @testforstephen


radu-matei:
A nice to have here would be to pre-populate the port input with all `containerPort` exposed by the pod - but this requires either parsing the `kubectl describe` command or using the Kubernetes API/TypeScript library directly.

bnookala:
@radu-matei excellent idea - let's label it for follow up? i'd like to get this in soon so i can re-use code for the dashboard.

src/components/kubectl/port-forward.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @param targetPort The target port to forward to.
+ * @param localPort (Optional) local port. If not provided, an unbound port is used.
+ */
+export async function portForwardToPod (podName: string, targetPort: number, localPort?: number) {
+    console.log(`port forwarding to pod ${podName} at port ${targetPort}`);
+
+    let usedPort;
+    if (!localPort) {
+        usedPort = await portFinder.getPortPromise({
+            port: 10000
```
radu-matei: Why is the local port always 10000?
I would expect:
- either the default `kubectl port-forward` experience, where if you only pass one port this will be the one used both locally and remote
- to have two steps in starting the port forwarding, one asking the local port, the other asking the remote port.

Either way, I would like to see the ability to specify the local port,

What do you think?

radu-matei: Why is the local port always 10000?
I would expect:

- either the default `kubectl port-forward` experience, where if you only pass one port this will be the one used both locally and remote
- to have two steps in starting the port forwarding, one asking the local port, the other asking the remote port.

Either way, I would like to see the ability to specify the local port.

What do you think?

bnookala: @radu-matei it's a badly named parameter… the port option tells the portfinder to search for an unbound port starting at the provided port, going up. as far as providing the local port through the UI, I'll change it to use the `localPort:remotePort` syntax.

bnookala:
@radu-matei Updated! Implemented your suggestion to ask for the local port. 
port-forward

PR #162, opened 2018-04-06
Add a drill-down for Pods created by a Deployment.

brendandburns:
One option would be to go:

Deployment -> ReplicaSet -> Pods

If we'd rather...


src/kubectlUtils.ts
```diff
@@ -101,6 +110,50 @@ export async function getNamespaces(kubectl: Kubectl): Promise<Namespace[]> {
...
+            name: item.metadata.name,
+            selector: item.spec.selector
+        };
+    });
+}
+
+export async function getPods(kubectl: Kubectl, selector: any): Promise<Pod[]> {
+    const currentNS = await currentNamespace(kubectl);
+
+    var labels = [];
```
bnookala: using `var` here (and once more below) is causing tslint to complain, and the travis build to fail. you can change the `var` declaration here to `const`, and the one below to a `let`.

src/kubectlUtils.ts
```diff
@@ -101,6 +110,50 @@ export async function getNamespaces(kubectl: Kubectl): Promise<Namespace[]> {
     });
 }
 
+export async function getDeployments(kubectl: Kubectl): Promise<Deployment[]> {
+    const currentNS = await currentNamespace(kubectl);
+
+    const shellResult = await kubectl.invokeAsync("get deployments -o json --namespace=" + currentNS);
```
bnookala: you can use some new fangled template strings here! 

ie:
```
kubectl.invokeAsync(`get deployments -o json --namespace=${currentNS}`)
```

src/kubectlUtils.ts
brendandburns: Done.

src/kubectlUtils.ts
brendandburns: Done.

brendandburns:

src/explorer.ts
```diff
@@ -213,3 +213,31 @@ class KubernetesNamespaceResource extends KubernetesResource {
...
+}
+
+class KubernetesDeploymentResource extends KubernetesResource {
+    constructor(readonly kind: kuberesources.ResourceKind, readonly id: string, readonly metadata?: any) {
+        super(kind, id, metadata);
+    }
+
+    getTreeItem(): vscode.TreeItem | Thenable<vscode.TreeItem> {
+        const treeItem = new vscode.TreeItem(this.id, vscode.TreeItemCollapsibleState.Collapsed);
+        treeItem.contextValue = `vsKubernetes.${this.id}`;
```
testforstephen: Can we change the contextValue to `vsKubernetes.resource.${this.kind.abbreviation}`? The new value you provided will lose the registered context menus for deployment nodes.

brendandburns:
@testforstephen comment addressed, please re-check.

Thanks!

PR #166, opened 2018-04-11
provide json content for kubernetes resources other than opening a new file

andxu:
Signed-off-by: andxu <andxu@microsoft.com>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -52,6 +52,7 @@ import * as clusterproviderregistry from './components/clusterprovider/clusterpr
 import * as azureclusterprovider from './components/clusterprovider/azure/azureclusterprovider';
 import { KubernetesCompletionProvider } from "./yaml-support/yaml-snippet";
 import { showWorkspaceFolderPick } from './hostutils';
+import { KubernetesDocumentProvider } from "./kube.documentProvider";
 
 
 let explainActive = false;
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -212,6 +213,34 @@ export async function activate(context) : Promise<extensionapi.ExtensionAPI> {
     }, this);
     await registerYamlSchemaSupport();
 
+
+      const provider: vscode.TextDocumentContentProvider = <vscode.TextDocumentContentProvider>{
```
bnookala: I'd prefer it if we can move this into its own file and extend the vscode.TextDocumentContentProvider interface as needed.

src/extension.ts
```diff
@@ -212,6 +213,34 @@ export async function activate(context) : Promise<extensionapi.ExtensionAPI> {
...
+        onDidChange: null,
+        provideTextDocumentContent: (uri: vscode.Uri, token: vscode.CancellationToken): Thenable<string | undefined> => {
+            if (uri.authority === 'loadkubernetescore') {
```
bnookala: early return if the authority is not `loadkubernetescore` ie:

```
// bad authority
if (uri.authority !== `loadkubernetescore`) {
    vscode.window.showErrorMessage('Unrecognized operation: ' + uri.authority);
    return;
}

// authority is ok
…
```

src/extension.ts
```diff
@@ -212,6 +213,34 @@ export async function activate(context) : Promise<extensionapi.ExtensionAPI> {
...
+                const value = querystring.parse(uri.query).value;
+                if (!value) {
+                    vscode.window.showErrorMessage(`Missing required value in uri: ${uri}`);
+                    return undefined;
+                }
+                return new Promise((resolve, reject) => {
+
+                    kubectl.invokeWithProgress(" -o json get " + value, `Loading ${value}...`, (result, stdout, stderr) => {
```
bnookala: can we make this async ie `const getData =  await kubectl.invokeAsyncWithProgress(command)`

andxu:
Update the PR.

src/kube.documentProvider.ts
```diff
@@ -0,0 +1,35 @@
...
+import * as querystring from "querystring";
+import { Kubectl } from "./kubectl";
+
+/// a kubernetes content provider provides json content for kubernetes resources
+export class KubernetesDocumentProvider implements vscode.TextDocumentContentProvider {
+    public constructor(private kubectl: Kubectl) {
+
+    }
+    provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken): vscode.ProviderResult<string> {
+        return new Promise<string>(async (resolve, reject) => {
```
bnookala: hmm - perhaps pull this anonymous function out into its own named async function on this class?

andxu: provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken): vscode.ProviderResult<string> will compile error if you add async because ProviderResult is not compatible with Promise

andxu: async provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken): Promise<string>  seems work.

andxu: Changed accordingly

bnookala:
Looks great! Thank you @andxu

PR #179, opened 2018-04-19
Add explorer api to allow third-party extension to mount tree nodes into explorer

testforstephen:
This PR will make Kubernetes Explorer Tree View to be extendable by other extensions.
In my current design, the third-party extension can add new children nodes at all levels of Kubernetes Explorer Tree View. I'm not sure whether this is too flexible. If anyone have concern about this, please let me know.

And See the sample code about how to extend the explorer on other extension side.
https://github.com/testforstephen/vscode-service-catalog/blob/master/src/extension.ts

https://github.com/testforstephen/vscode-service-catalog/blob/master/src/explorer.ts

@squillace @bnookala @andxu @akaroml

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,3 +27,48 @@ To implement a provider for the 'Create Cluster' and/or 'Add Existing Cluster' c
 * Your HTTP server must respond to this request with a suitable first page.  From this point on it is all in your hands, though typically your first page will gather some information, and have a link either to further information-gathering pages or to perform the action immediately.
 
 For an example of this, see the `components/clusterprovider/azure` directory.  If you have any questions or run into any problems, please post an issue - we'll be very happy to help.
+
+## Contribute a new ExplorerDataProvider to Kubernetes Explorer View
+
+To allow the third-party extensions to contribute more data nodes to the Kubernetes Explorer View, we have made the Kubernetes Explorer View to be extensible. The Kubernetes Explorer View will render the tree based on the registered ExplorerDataProviders. The current tree data is from a Built-in Data Provider implementation.
+
+To implement a new ExplorerDataProvider, your extension needs to do the following:
+
+* Add the activationEvent `onView:extension.vsKubernetesExplorer` to the `package.json`.
+* Copy the Explorer API interface definition file `src/explorer.api.ts` to your repository.
+* Get the VS Code Kubernetes extension's API object via the vscode.extensions.getExtension<T> and Extension<T>.activate() functions.
+```
+    const extension = vscode.extensions.getExtension("ms-kubernetes-tools.vscode-kubernetes-tools");
+    if (extension) {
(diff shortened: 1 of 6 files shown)


itowlson:
Can we have documentation on the API and its usage?  This would make it a lot easier to discuss and review the design, and would ensure that we refine internal types such as `KubernetesObject` if required to be easy for extenders to work with.

testforstephen:
The API is similar as clusterProviderRegistry API. Will update the usage at `CONTRIBUTING.md`.

testforstephen:
@itowlson I guess you may miss some background. The target is to allow standalone service-catalog extension to contribute a new Node "External Services" at Kubernetes Explorer Tree View.  The external services will list the provisioned instances by svcat cli tool. See the gif for preview.

testforstephen:
- Move `KubernetesObject` definition to `explorer.api.ts` for better reference from the external extension.
- Update CONTRIBUTING document about the API design idea and the usage gudiance.

@itowlson Could you help have a look?

squillace:
@itowlson and @testforstephen: The larger background here is the question we had about whether we should continue to extend the Kube extension or enable the Kube extension to itself be extended by others. In a series of conversations, Brendan and I agreed that the plan should be to open the kube extension (the intended goal of this PR) to layering by other extensions -- first, the service catalog extension of which bhargav built an experimental version to see what the UX might be like. 

We are assuming that people will use the base extensions, and then layer other extensions in on top, depending upon which they want to use, and we will also create an extension pack for the core group that we think create the best experience.

itowlson:
@squillace To me, 'open' means 'can be extended by third parties in a reliable, robust way, without hand-holding or ongoing involvement from the core extension devs'.  That means we are designing not just an extension but a platform, a framework which will need clear, stable, well-documented, versioned APIs with a commitment to backward compatibility.  That's fine, but it needs an approach that we haven't invested in to date - I would suggest, for example, that it means we need to approach APIs documentation-first, and we need to plan a versioning strategy, and we need careful separation between APIs and internal data structures.  There's a huge number of things other extensions could want to do on top of Kubernetes, of many different kinds - if we want to offer more than ad hoc point extensibility such as the "you can hook the Create Cluster wizard" and "you can add nodes to our tree" examples, we need to stop and think _hard_ about what we want to expose and how.

squillace:
@itowlson what, in general, are our limitation abilities? For example, if I said that in this case first I wanted to limit modifyability to a `External Services` node underneath Kubernetes and add a `Add External Service Reference` to the right-click on a /chart directory, could I do that? for that, what would be the required scope of extensibilitiness? I like that word.

itowlson:
Closing old PR - this worked well for the Service Catalog extension proof of concept and has provided useful learning, but in our next steps we are likely to take a different approach to API versioning, and will likely require a more general strategy for extending the explorer tree.

PR #180, opened 2018-04-20
Add a drill down for services...

brendandburns:
Builds on #162 and adds a similar drill-down for Service objects.


squillace:
Yeah, this is a good idea. Thanks! Testing it out....

squillace:
Works great for me. Nice addition.

image

squillace:
another note; it particular, I very much like the way that you pop open the services in the editing window when you open/select a Service in the left pane, but if I click a code file the service def vanishes, no prompt, nothing. Just gets out of my way.

I'd like all the openings/closings to work this way...

PR #181, opened 2018-04-21
Add browsing of ConfigMaps and Secrets to the tree explorer.

brendandburns:
@itowlson @bnookala

### DIFF
--- a/src/configMap.ts
+++ b/src/configMap.ts
@@ -1,0 +1,32 @@
+import * as vscode from 'vscode';
+import { fs } from './fs';
+import { shell } from './shell';
+import { host } from './host';
+import { create as kubectlCreate, Kubectl } from './kubectl';
+import { currentNamespace } from './kubectlUtils';
+import { KubernetesFileObject, KubernetesExplorer } from './explorer';
+
+export const uriScheme : string = "k8sviewfiledata";
+
+const kubectl = kubectlCreate(host, fs, shell);
+
+export class ConfigMapTextProvider implements vscode.TextDocumentContentProvider {
+    constructor(readonly kubectl: Kubectl) {}
+
+    provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken): vscode.ProviderResult<string> {
+        const parts = uri.path.split('/');
+        if (uri.authority == 'configmaps') {
+            return parts[1];
+        } else {
+            let b: Buffer = new Buffer(parts[1], 'base64');
+            return b.toString();
+        }
+    }
+}
+
+export function loadConfigMapData(obj: KubernetesFileObject) {
+    vscode.workspace.openTextDocument(
+        vscode.Uri.parse(`${uriScheme}://${obj.resource}/${obj.configData[obj.id]}/${obj.id}`)).then((doc) => {
+            vscode.window.showTextDocument(doc);
+        });
+}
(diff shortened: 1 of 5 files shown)


squillace:
@testforstephen as well.

testforstephen:
The ci failed because of errors from `gulp tslint` command. See the ci link https://travis-ci.org/Azure/vscode-kubernetes-tools/jobs/369408490

brendandburns:
Rebased.

src/extension.ts
```diff
@@ -19,8 +19,10 @@ import * as clipboard from 'clipboardy';
 
 // Internal dependencies
 import { host } from './host';
+import { loadConfigMapData } from './configMap'
 import * as explainer from './explainer';
 import { shell, ShellResult } from './shell';
+import * as configmaps from './configMap';
```
itowlson: Maybe I'm missing something obvious but I can't see `configMap.ts` and when I pull this branch I get missing file errors for it.  Where should I be finding it?

src/extension.ts
```diff
@@ -19,8 +19,10 @@ import * as clipboard from 'clipboardy';
 
 // Internal dependencies
 import { host } from './host';
+import { loadConfigMapData } from './configMap'
```
itowlson: We do an `import *` of this on line 25 so do we need to promote this one function to unqualified?

src/extension.ts
brendandburns: added.

src/extension.ts
brendandburns: fixed.

brendandburns:
Comments addressed, ptal.

Thanks!
--brendan

src/configMap.ts
```diff
@@ -0,0 +1,32 @@
...
+import { KubernetesFileObject, KubernetesExplorer } from './explorer';
+
+export const uriScheme : string = "k8sviewfiledata";
+
+const kubectl = kubectlCreate(host, fs, shell);
+
+export class ConfigMapTextProvider implements vscode.TextDocumentContentProvider {
+    constructor(readonly kubectl: Kubectl) {}
+
+    provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken): vscode.ProviderResult<string> {
```
itowlson: There seems to be a weird bug here with one of my configmaps.  This has three entries: the first a JSON document created from a file, the others simple strings created from literals.  (I don't yet know which of these details is relevant.)  If I click on the first (JSON file) _after_ clicking one of the others (simple literal) then VS Code hangs with 100% CPU usage.  Note that clicking on the first (JSON file) _before_ clicking on any of the others works fine; it is only the switch from one of the others to the first that causes the error.

I'm not sure what causes this.  For other configmaps I don't see the same issue when switching to the first entry, so I don't think it's related to position.  I don't know if configmaps remember the 'file'-ness of the source data but I thought not, so I assume it's not that.  So could it be the JSON formatting of the problem entry?

itowlson: Correction: clicking on the first one (the JSON file) _always_ causes the hang.  I was fooled because I was looking at the configmap JSON and mistook it for the configmap **entry** JSON.

brendandburns: I repro'd this... Not sure what's wrong but I will dig in... Thanks for catching it!

brendandburns: ok, figured it out, the URI was malformed, and that caused vscode to spiral out of control. Will file a separate bug on VS Code, but this is fixed here via bas64 encoding everything.

PR #187, opened 2018-04-26
Draft debug adapter should respect draft-path config setting

itowlson:
Hotfix for #160 for the case where `draft` is not on the system path but instead located via the `draft-path` config setting.

@radu-matei @squillace I am on Windows so can't test this (bring on 0.14) - can you folks take a look and see if it works for you?  Thanks!


itowlson:
I think this will be obsolete when we merge #206.

itowlson:
Closing as #206 covers this I believe.

PR #197, opened 2018-04-30
Fixed error when selecting service without selector in tree

### LINKED ISSUE #190: Cannot read property 'matchLabels' of undefined when clicking service "kubernetes" node in Explorer View
In Explorer, click the "Services -> kubernetes" node under the default namespace, it pops error message `Cannot read property 'matchLabels' of undefined`.

itowlson:
Kubernetes services are not required to have selectors, for example if the service is implemented by an external endpoint.  Clicking a service without a selector in the tree view previously caused an error "Cannot read property 'matchLabels' of undefined".  This fixes that error.

Fixes #190.


brendandburns:
LGTM.

PR #199, opened 2018-04-30
Document preequisities for running from source

### LINKED ISSUE #153: Extension fails because of unknown dependency
If I don't have the Red Hat extension, it will fail with the following message:

```
Extension 'ms-kubernetes-tools.vscode-kubernetes-tools' failed to activate. Reason: unknown dependency 'redhat.vscode-yaml'.
```

Should it prompt me to install it rather than fail?

itowlson:
It's not obvious that you need to install the RedHat YAML extension before running this extension from source (i.e. debugging it in the Extension Development Host).  This PR adds some documentation around that.

Fixes #153.  Well, sort of.


brendandburns:
LGTM.

PR #206, opened 2018-05-07
Support alternate kubeconfigs

itowlson:
We've had a few people who have multiple kubeconfig files and want to be able to swap between them.  This PR adds a setting (which can be at user or workspace level) to control which kubeconfig file is used.

Considerations for reviewers:

* Is a setting discoverable enough, or should users be able to swap using a hotkey?  Or should we even display multiple kubeconfigs in the tree view and allow users to switch between them?
* Are there any paths I have missed that spawn kubectl or anything else that uses kubeconfig?  I _think_ I have them all, but we have sprouted quite a few places that go round the back of the abstractions.  Please check areas that you know to be sure I haven't messed them up (e.g. @radu-matei's Draft debug stuff).
* In order to minimise the risk above, I've tried to encapsulate the way we invoke tools, and centralise associated logic, e.g. by setting up PATH in child processes rather than by invoking binaries by full path.  (This should have been a separate PR; sorry.)  Again, please check if any of the changes in areas you know look suspicious, or test them if possible, in scenarios both where the tool is on the PATH and where the tool is _not_ on PATH and its location is set through config.

### DIFF
--- a/src/kubectl.ts
+++ b/src/kubectl.ts
@@ -84,10 +84,8 @@ class KubectlImpl implements Kubectl {
                 }
             });
             this.context.host.onDidChangeConfiguration((change) => {
-                if (change.affectsConfiguration('vs-kubernetes')) {
-                    if (this.sharedTerminal) {
-                        this.sharedTerminal.dispose();
-                    }
+                if (change.affectsConfiguration('vs-kubernetes') && this.sharedTerminal) {
+                    this.sharedTerminal.dispose();
                 }
             });
         }
--- a/src/kubectl.ts
+++ b/src/kubectl.ts
@@ -163,11 +163,7 @@ async function spawnAsChild(context: Context, command: string[]): Promise<ChildP
 
 async function invokeInTerminal(context: Context, command: string, pipeTo: string | undefined, terminal: Terminal): Promise<void> {
     if (await checkPresent(context, 'command')) {
-        let bin = baseKubectlPath(context).trim();
-        if (bin.indexOf(" ") > -1 && !/^['"]/.test(bin)) {
-            bin = `"${bin}"`;
-        }
-        const kubectlCommand = `${bin} ${command}`;
+        const kubectlCommand = `kubectl ${command}`;
         const fullCommand = pipeTo ? `${kubectlCommand} | ${pipeTo}` : kubectlCommand;
         terminal.sendText(fullCommand);
         terminal.show();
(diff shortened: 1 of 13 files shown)


src/kubectl.ts
```diff
@@ -71,6 +83,13 @@ class KubectlImpl implements Kubectl {
                     disposable.dispose();
                 }
             });
+            this.context.host.onDidChangeConfiguration((change) => {
+                if (change.affectsConfiguration('vs-kubernetes')) {
```
bnookala: consolidate the two ifs into one check `if (change.affectsConfiguration('vs-kubernetes') && this.sharedTerminal) {`

src/shell.ts
```diff
@@ -109,7 +111,43 @@ function execCore(cmd: string, opts: any): Promise<ShellResult> {
...
+    }
+
+    return env;
+}
+
+function pathVariableName(env: any): string {
+    if (isWindows) {
+        for (const v of Object.keys(env)) {
+            if (v.toLowerCase() === "path") {
+                return v;
```
bnookala: if we're iterating through this environment dict to check if a key named `path` exists, why not just return `"path"` immediately?

itowlson: Because JavaScript is case-sensitive, but the Windows environment isn't.  So if I return `path`, but the Windows environment variable is called `Path` (as it is in PowerShell), then the `env` object ends up with two properties `{ …; Path: "whatever;it;inherited;from;parent"; path: "kubectl-dir;helm-dir;draft-dir"; … }`, and then when `shelljs` sets the child process environment variables, the second one clobbers the first one.  Hence, I need to find the property of the `env` object which matches `path` _ignoring case_.

Ask me how I found this out...  **grin**

bnookala: hah! could you perhaps use `process.env.SHELL` to identify if the parent shell is powershell vs. windows command prompt?

itowlson: I could... or I could just not take a dependency on specific shells (or on the presumption that these shells guarantee a particular casing of PATH), and use whatever casing of PATH happens to be present in this environment.  This is safer, and probably no more complex once you allow for all the switches in the check-the-shell approach.

itowlson: On the plus side this discussion did get me to spot an error in this code and now I really need to find the `tslint` option for "die, implicit truthiness and falseness, die."

PR #207, opened 2018-05-07
revert changes for provide json content for kubernetes resources other than opening a new file 

andxu:
because users may want to edit the generated json file and save it in current directory, we need to revert it.


andxu:
https://github.com/Azure/vscode-kubernetes-tools/issues/200#issuecomment-386955682

Issue #141, opened 2018-03-28
"Kubernetes log" can't work without a open yaml file.


songkaiape:
I think this command should behave like other command like "Kubernetes get".
When there is an open yaml file , just get the resource defined in the file. If there is no open yaml file, it should prompt available resource type to choose.

kyliel:
I can see resource type "pod" after I type "Kubernetes: logs". Could you please double confirm with latest version?

songkaiape:
@kyliel  
Minikube version :0.25.2
VScode: 1.22.1
OS: Ubuntu 16.04 &&MacOS
build: 3940 https://vscjavaci.cloudapp.net/job/vscode-k8s-tools.vsix/3940/  &&release build 0.17
I have checked this issue with private build 3940 and release build 0.1.7 .  This issue still repro. 
1. When there is no yaml file open, Executing "Kubernetes log"  will pop out an error "No active editor"
2. When the yaml file is not a pod kind, Executing "Kubernetes log" has no response.

![image](https://user-images.githubusercontent.com/4318793/38598056-71d430de-3d8d-11e8-8a72-2a7b014698a3.png)

songkaiape:
If there is an open file, kubernetes logs will work good. And if you close all editor ,it will throw "No active editor" error

itowlson:
The desired experience for this, per discussion a few weeks back captured in Kylie's email of 6 March, is:

```
Logs: go to terminal. Ask “which object do you want” – same
behavior as “Get”. As for what detailed resource type is supported,
read “kubctl logs”. 
1.	Kubernetes logs
2.	Show drop-down list for pods 
3.	Leave others to terminal commands
```

I don't recall the discussion well enough to fully parse this, but it sounds like we want to ignore any active file and just go straight to the prompt (can we remember why we weren't interested in the active file?).  This would duck the 'no active editor'/non-pod file issue of course.  The available resource types aren't well documented in `kubectl logs` but appear to be at minimum pod, deployment and job (service didn't work for me).

Can anyone remember what 'leave others to terminal commands' referred to?  Are the notes saying we _only_ want to list pods, not deployments or jobs?

We should probably handle the case where more than there is more than one container in the pod, though not sure how easily we can tell this if we support non-pod resource types.

itowlson:
Fixed via #208.

PR #208, opened 2018-05-08
Support container-level operations on multi-container pods

### LINKED ISSUE #203: Feature request: support multi-container pods
When running multi-container pods, the pod operations don't all work well. Starting a terminal in the pod defaults to one of the containers and happens to work if that's the one you need. Viewing logs doesn't work as it requires specifying a container name. You can repeat the failed command it executed with the container name option added to it, but it's not a great UX.

Multi-container pods are pretty frequently used having sidecars for db connections or things like Envoy. Could the pod have an additional drilldown per container?

itowlson:
The Terminal, Exec and Logs commands previously always operated only on the first container of a multi-container pod.  This PR prompts for the container if the pod contains multiple containers.  Note that Debug (Attach) and Sync already handled multiple containers (the former on a different code path) - I've rationalised the UI but have not consolidated the Debug (Attach) selection code.

Fixes #203, although the original post there asked for containers to be displayed in the tree view; we can add that too if people feel strongly about it.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -978,25 +978,22 @@ function parseName(line) {
     return line.split(' ')[0];
 }
 
-function findPod(callback) {
+function findPod(callback: (pod: PodMetadata) => void) {
     let editor = vscode.window.activeTextEditor;
-    if (!editor) {
-        vscode.window.showErrorMessage('No active editor!');
-        return null; // No open text editor
-    }
-
-    let text = editor.document.getText();
-    try {
-        let obj = yaml.safeLoad(text);
-        if (obj.kind && obj.kind === 'Pod' && obj.metadata) {
-            callback({
-                name: obj.metadata.name,
-                namespace: obj.metadata.namespace
-            });
-            return;
+    if (editor) {
+        let text = editor.document.getText();
+        try {
+            let obj = yaml.safeLoad(text);
+            if (obj.kind && obj.kind === 'Pod' && obj.metadata) {
+                callback({
+                    name: obj.metadata.name,
+                    namespace: obj.metadata.namespace
+                });
+                return;
+            }
+        } catch (ex) {
+            // pass
         }
-    } catch (ex) {
-        // pass
     }
 
     quickPickKindName(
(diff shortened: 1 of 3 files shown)


itowlson:
Fixes #141, as the changes to support multi-container pods also fixed the pod selection logic.

src/extension.ts
```diff
@@ -902,41 +902,50 @@ function promptKindName(resourceKinds: kuberesources.ResourceKind[], description
...
-                            });
+            quickPickKindNameFromKind(resourceKind, opts, handler);
+        }
+    });
+}
+
+function quickPickKindNameFromKind(resourceKind: kuberesources.ResourceKind, opts, handler) {
+    let kind = resourceKind.abbreviation;
+    kubectl.invoke("get " + kind, (code, stdout, stderr) => {
+        if (code === 0) {
```
bnookala: reverse this condition since the majority of work is done if the code === 0

ie: 
```
if (code !== 0) {
    vscode.window.showErrorMessage(stderr);
    return;
}

// everything else
```

src/extension.ts
```diff
@@ -902,41 +902,50 @@ function promptKindName(resourceKinds: kuberesources.ResourceKind[], description
...
+            let names = parseNamesFromKubectlLines(stdout);
+            if (names.length > 0) {
```
bnookala: same as above - reverse the condition ie:

```
if (names.length === 0) {
     vscode.window.showInformationMessage("…");
     return;
}
```

src/extension.ts
```diff
@@ -902,41 +902,50 @@ function promptKindName(resourceKinds: kuberesources.ResourceKind[], description
...
+                if (opts && opts.nameOptional) {
+                    names.push('(all)');
+                    vscode.window.showQuickPick(names).then((name) => {
+                        if (name) {
```
bnookala: reverse this condition :)

src/extension.ts
```diff
@@ -902,41 +902,50 @@ function promptKindName(resourceKinds: kuberesources.ResourceKind[], description
...
+                            handler(kindName);
                         }
-                    } else {
-                        vscode.window.showInformationMessage("No resources of type " + resourceKind.displayName + " in cluster");
-                    }
+                    });
                 } else {
-                    vscode.window.showErrorMessage(stderr);
+                    vscode.window.showQuickPick(names).then((name) => {
+                        if (name) {
```
bnookala: reverse the condition

src/extension.ts
```diff
@@ -965,27 +974,22 @@ function parseName(line) {
...
-    let text = editor.document.getText();
-    try {
-        let obj = yaml.safeLoad(text);
-        if (obj.kind !== 'Pod') {
-            return;
+    if (editor) {
+        let text = editor.document.getText();
+        try {
+            let obj = yaml.safeLoad(text);
+            if (obj.kind && obj.kind === 'Pod' && obj.metadata) {
```
bnookala: perhaps we can move this condition into its own function that returns a boolean so we can do something like :

```
if (isPod(obj)) { 
// etc.
}
```

src/extension.ts
itowlson: Hmm, not so keen on reversing this one.  The `showQuickPick(xs).then((x) => if (x) { ... })` idiom turns up in quite a few places, and flipping the condition seems to me like it won't simplify the reading.  What we should do is encapsulate the idiom so that callers can just be all happy-path about it and not worry about checking the cancellation path.

src/extension.ts
itowlson: Not keen on reversing this one either.

src/extension.ts
itowlson: Good call.  In my dreams we would get fancy and declare a `Pod` type and `isPod` would return `obj is Pod` for glorious type safety but life is short.

src/extension.ts
```diff
@@ -1145,9 +1198,13 @@ function execKubernetes() {
...
 async function terminalKubernetes(explorerNode?: explorer.ResourceNode) {
     if (explorerNode) {
-        // For those images (e.g. built from Busybox) where bash may not be installed by default, use sh instead.
-        const isBash = await isBashOnPod(explorerNode.id);
-        execTerminalOnPod(explorerNode.id, isBash ? 'bash' : 'sh');
+        const podMetadata = { name: explorerNode.id, namespace: undefined };  // TODO: namespaces
+        const container = await selectContainerForPod(podMetadata);
+        if (container) {
+            // For those images (e.g. built from Busybox) where bash may not be installed by default, use sh instead.
+            const isBash = await isBashOnContainer(podMetadata.name, podMetadata.namespace, container.name);
```
bnookala: nit: maybe we can make isBashOnContainer return the shell to run on the container, since we have some pretty regular usage of it.

src/extension.ts
```diff
@@ -1145,9 +1198,13 @@ function execKubernetes() {
 
 async function terminalKubernetes(explorerNode?: explorer.ResourceNode) {
     if (explorerNode) {
```
bnookala: invert the condition so we can return early w/ `execKubernetesCore`

itowlson: Disagree: although the 'not an explorer node' branch is shorter, it's not a true early exit scenario and the two branches have equal weight, so I prefer the positive if condition.

Issue #210, opened 2018-05-08
Duplicate file contents being loaded into editor v0.1.9


earlcoder:
If you open a cluster and click something like ingress->ingress_file , then click ingress_file again it will load duplicate contents in to the editor for as many times as you click it.
This happens to any editable file.

itowlson:
Fix merged (#211) - will be included in 0.1.10.  Thanks for flagging this!

earlcoder:
Thanks guys

PR #211, opened 2018-05-08
Fix copy of a resource being appended when resource re-selected

itowlson:
Consider the following scenario:

1. The user selects a resource in the tree view (or uses the `Load` command).
2. The extension creates an untitled document containing the resource JSON.
3. The user leaves the document open.
4. Later, the user again selects the same resource (or uses the `Load` command to load it again).

At the moment, this has a bug where the second time, the resource JSON is appended to the existing document.  This PR loads each copy of the resource into a new document.  Documents are named e.g. `deployment-foo.json`, `deployment-foo.1.json`, `deployment-foo.2.json`, etc.

A nice enhancement to this would be to compare the current JSON against the content of the existing document, and simply re-show that document if the resource had not changed.  This would avoid creating a lot of temp documents that the user has to close if they are simply browsing around in the tree.  (Alternatively, if the existing document was a temp document, I guess we could overwrite its contents with the new resource state...?)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -620,12 +620,9 @@ function loadKubernetesCore(value: string) {
             doc.uri.fsPath === path.join(vscode.workspace.rootPath || "", `${filename}.${extension}`)
         );
 
-        if (existingDocument) {
-            const existingDocContent = existingDocument.getText();
-            if (equalIgnoringWhitespace(existingDocContent, stdout)) {
-                vscode.window.showTextDocument(existingDocument);
-                return;
-            }
+        if (existingDocument && equalIgnoringWhitespace(existingDocument.getText(), stdout)) {
+            vscode.window.showTextDocument(existingDocument);
+            return;
         }
 
         const filepath = findUniqueName(filename, extension);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -614,7 +614,21 @@ function loadKubernetesCore(value: string) {
         }
 
         const filename = value.replace('/', '-');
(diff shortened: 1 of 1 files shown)


itowlson:
Fixes #210.

bnookala:
@itowlson I get why this is necessary... but would it be possible to do as you mentioned in your last paragraph and compare the current JSON of the temporary document, replacing it as necessary? I'd think a simple hash of the document contents would be a fairly reasonable approach to this. I really dislike the current behavior (as unavoidable as it is), and accidentally opening > 1 temporary document, and being prompted to save every single one would be even more annoying.

itowlson:
@bnookala Are you arguing for the suggestion of blatting the existing JSON in favour of the current state, or just for reactivating the document if it matches?  The latter seems safe; the former a bit trickier, and feel fraught with risks...!

bnookala:
@itowlson on second thought... just the latter :D

src/extension.ts
```diff
@@ -614,7 +614,21 @@ function loadKubernetesCore(value: string) {
...
 
         const filename = value.replace('/', '-');
-        const filepath = findUniqueName(filename, "json");
+        const extension = "json";
+
+        const existingDocument = vscode.workspace.textDocuments.find((doc) =>
+            doc.uri.fsPath === path.join(vscode.workspace.rootPath || "", `${filename}.${extension}`)
+        );
+
+        if (existingDocument) {
```
bnookala: condensable into one if block

PR #223, opened 2018-05-28
Kubernetes resource object types

itowlson:
This PR defines some interfaces for Kubernetes resources and collections, so that when we work with them (e.g. after loading them from JSON) we can work with them in a safer way.  It's motivated by having to hold 'what is this `any` representing... is it a Foo or some internal type containing interesting data from a Foo... what fields does it actually have... etc.' in my head.  The main additions are the resource interfaces themselves (which contain only the fields we're interested in, but that's okay, they don't need to be comprehensive) and the `kubectl.asJson` method which encapsulates the JSON parsing and type assertion (and should be used in preference to `invokeAsync` when we are requesting JSON for materialisation as an object).

These changes don't prevent maintainers from using `any` where appropriate - I just find that capturing resource shapes helps me to avoid errors and reason more efficiently.

Making these changes also revealed some cases where resource loading code wasn't dealing with the possibility of failure, e.g. `kubectl` errors; these have been fixed.

### DIFF
--- a/src/components/kubectl/dashboard.ts
+++ b/src/components/kubectl/dashboard.ts
@@ -12,7 +12,7 @@ import { shell } from '../../shell';
 import { create as kubectlCreate, Kubectl } from '../../kubectl';
 import { installDependencies } from '../../extension';
 import { Node, KubernetesCollection, Pod } from '../../kuberesources.objectmodel';
-import { failed } from '../../wizard';
+import { failed } from '../../errorable';
 
 
 const KUBE_DASHBOARD_URL = "http://localhost:8001/ui/";
--- a/src/components/kubectl/dashboard.ts
+++ b/src/components/kubectl/dashboard.ts
@@ -11,6 +11,7 @@ import { host } from '../../host';
 import { shell } from '../../shell';
 import { create as kubectlCreate, Kubectl } from '../../kubectl';
 import { installDependencies } from '../../extension';
+import { Node, KubernetesCollection, Pod } from '../../kuberesources.objectmodel';
 
 
 const KUBE_DASHBOARD_URL = "http://localhost:8001/ui/";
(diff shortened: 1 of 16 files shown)


itowlson:
I've also used this to move the `Errorable<T>` interface out into its own module, because it's increasingly not bound to wizard scenarios.

brendandburns:
LGTM.

src/components/kubectl/dashboard.ts
```diff
@@ -30,9 +32,11 @@ let terminal: vscode.Terminal;
  * @returns Boolean identifying if we think this is an AKS cluster.
  */
 async function isAKSCluster (): Promise<boolean> {
-    const nodes =  await kubectl.invokeAsync('get nodes -o json');
-    const nodesJson = JSON.parse(nodes.stdout);
-    const nodeItems = nodesJson.items;
+    const nodes = await kubectl.asJson<KubernetesCollection<Node>>('get nodes -o json');
```
bnookala: wondering if we should have `asJson` append the `-o json` so somebody down the line won't accidentally forget

src/kubectl.ts
```diff
@@ -210,16 +215,25 @@ function baseKubectlPath(context: Context): string {
...
         lines = lines.filter((l) => l.length > 0);
-        return lines;
+        return { succeeded: true, result: lines };
+
+    }
+    return { succeeded: false, error: [ shellResult.stderr ] };
+}
+
+async function asJson<T>(context: Context, command: string): Promise<Errorable<T>> {
+    const shellResult = await invokeAsync(context, command);
```
bnookala: yeah let's definitely just add the `-o json` somewhere here

src/components/kubectl/dashboard.ts
itowlson: Yeah... I thought about that, but then I thought "but what if someone already passed `-o json`" and then "what if someone messes up and passes `-o yaml`, what sort of confusing error message will they get then" and it felt like it might turn into a pit of edge cases.

And then just now I discovered that `kubectl` _doesn't care_ if you pass multiple output options, it just uses the last one, so PARTY TIME.

Except wait, what if the command has a trailing argument (a la `--`) so you can't stick extra args on the end... *sigh*

itowlson:
Discussed `-o json` with @bnookala on Slack, and decided to skip this for now as it introduces too much complexity.  But we'll need to watch out for this when reviewing code that calls `asJson` and if it's an issue then maybe we'll need to tackle it.

Issue #231, opened 2018-06-05
Bug: Port Forwarding Only Works for Default Namespace


shreyu86:
I have been using this extension and it is awesome productivity booster. 
I noticed that port forwarding currently does not propagate the namespace (selected namespace) context to `kubectl` [here](https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/components/kubectl/port-forward.ts#L45) and other places `portForwardToPod` is called no namespace is passed and hence the `default` namespace is used. 

Happy to work on a PR if you can point me towards the implementation you would like to see to plumb in the namespace.

itowlson:
cc @bnookala 

Thanks for the offer to tackle this!  If memory serves, using a non-default namespace should be changing the default namespace for `kubectl` (see `useNamespaceKubernetes` in extension.ts), so the code paths from right-click (`if (explorerNode)`) and 'select from drop-down' (lines 69ff.) _should_ work correctly.  If not then we do indeed need to propagate the namespace.  Let's confirm the issue occurs in these cases, and we can provide more guidance on the implementation of these cases if you want to work on it.

However there is definitely a problem if a pod manifest is open - in this case we infer the pod name from there, but do _not_ pick up the namespace.  Fixing this probably means updating the `tryFindKindNameFromEditor` function (referenced in line 179) which will have a knock-on effect throughout the extension - but this will fix similar incorrect behaviour throughout the extension.

Let us know if you plan to send a PR for this - otherwise we will tackle it, we just don't want to duplicate effort!  Happy to discuss other details of the implementation if the above pointers aren't enough.  Thanks again!

shreyu86:
Thanks for the response @itowlson, I can try fixing this bug:
1.  The `explorerNode` case was straightforward to [fix](https://github.com/shreyu86/vscode-kubernetes-tools/commit/fbc2a7f96a88886b6591938d8b94ad71abe2166a) I have tested this multiple times today. Let me know if I am in the right direction on this. 

2. In the case when pod manifest is open, is where I need some more pointers. Currently `tryFindKindNameFromEditor` returns `<object_kind>/<object_name>` there could be one implementation where it returns `<namespace>/<object_kind>/<object_name>` since other things (in `extension.ts`) rely on `tryFindKindNameFromEditor` wanted to get more feedback before implementing anything.

bnookala:
@itowlson @shreyu86 

[...]

PR #232, opened 2018-06-07
Fix for #231

shreyu86:
Based on the pointers and feedback on the issue here is an attempt to fix the problem. 
Please let me know if I have missed anything.

### DIFF
--- a/src/components/kubectl/port-forward.ts
+++ b/src/components/kubectl/port-forward.ts
@@ -25,6 +25,7 @@ interface PodFromDocument {
     readonly succeeded: true;
     readonly pod: string;
     readonly fromOpenDocument: true;
+    readonly namespace: string;
 }
 
 type PortForwardFindPodsResult = PodFromDocument | FindPodsResult;
@@ -64,7 +65,8 @@ export async function portForwardKubernetes (explorerNode?: any): Promise<void> 
             // The pod is described by the open document. Skip asking which pod to use and go straight to port-forward.
             const podSelection = portForwardablePods.pod;
             const portMapping = await promptForPort(podSelection);
-            portForwardToPod(podSelection, portMapping);
+            const namespace = portForwardablePods.namespace;
+            portForwardToPod(podSelection, portMapping, namespace);
             return;
         }
 
@@ -179,23 +181,20 @@ export function buildPortMapping (portString: string): PortMapping {
  */
 async function findPortForwardablePods (): Promise<PortForwardFindPodsResult> {
     let kindFromEditor = tryFindKindNameFromEditor();
-    let kind: string | undefined, podName: string | undefined;
 
     // Find the pod type from the open editor.
     if (kindFromEditor !== null) {
-        let parts = kindFromEditor.split('/');
-        kind = parts[0];
-        podName = parts[1];
 
         // Not a pod type, so not port-forwardable, fallback to looking
         // up all pods.
-        if (kind !== 'pods' && kind !== 'pod') {
+        if (kindFromEditor.resourceType !== 'pods' && kindFromEditor.resourceType !== 'pod') {
             return await findAllPods();
         }
 
         return {
             succeeded: true,
-            pod: podName,
+            pod: kindFromEditor.resourceName,
+            namespace: kindFromEditor.namespace,
             fromOpenDocument: true
         };
     }
--- a/src/components/kubectl/port-forward.ts
+++ b/src/components/kubectl/port-forward.ts
@@ -10,6 +10,7 @@ import { QuickPickOptions } from 'vscode';
 import * as portFinder from 'portfinder';
 import { Pod } from '../../kuberesources.objectmodel';
 import { succeeded } from '../../errorable';
+import * as kubectlUtils from '../../kubectlUtils';
 
 const kubectl = kubectlCreate(host, fs, shell, installDependencies);
 const PORT_FORWARD_TERMINAL = 'kubectl port-forward';
@@ -42,7 +43,8 @@ export async function portForwardKubernetes (explorerNode?: any): Promise<void> 
         // The port forward option only appears on pod level workloads in the tree view.
         const podName = explorerNode.id;
         const portMapping = await promptForPort(podName);
-        portForwardToPod(podName, portMapping);
+        const namespace = await kubectlUtils.currentNamespace(kubectl);
+        portForwardToPod(podName, portMapping, namespace);
         return;
     } else {
         let portForwardablePods: PortForwardFindPodsResult;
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -836,25 +836,35 @@ export function tryFindKindNameFromEditor(): string {
...
 
 export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
-    let kindName = tryFindKindNameFromEditor();
-    if (kindName === null) {
+    let kindObject = tryFindKindNameFromEditor();
+    if (kindObject === null) {
         promptKindName(resourceKinds, descriptionVerb, opts, handler);
     } else {
-        handler(kindName);
+        handler(kindObject.resourceType + '/' + kindObject.resourceName);
```
bnookala: let's stick to using template strings where we can:

```
`${kindObject.resourceType}/${kindObject.resourceName}`
```

src/extension.ts
```diff
@@ -1319,21 +1329,24 @@ const diffKubernetes = (callback) => {
...
+        let kindObject = null;
         let fileName = null;
 
         let fileFormat = "json";
 
         if (data) {
             fileFormat = (data.trim().length > 0 && data.trim()[0] == '{') ? "json" : "yaml";
-            kindName = findKindNameForText(data);
+            kindObject = findKindNameForText(data);
+            kindName = kindObject.resourceType + '/' + kindObject.resourceName;
```
bnookala: template string here

src/extension.ts
```diff
@@ -1319,21 +1329,24 @@ const diffKubernetes = (callback) => {
...
             fileName = path.join(os.tmpdir(), `local.${fileFormat}`);
             fs.writeFile(fileName, data, handleError);
         } else if (file) {
             if (!vscode.window.activeTextEditor) {
                 vscode.window.showErrorMessage('No active editor!');
                 return; // No open text editor
             }
-            kindName = tryFindKindNameFromEditor();
+            kindObject = tryFindKindNameFromEditor();
+            kindName = kindObject.resourceType + '/' + kindObject.resourceName;
```
bnookala: template string here

src/components/kubectl/port-forward.ts
```diff
@@ -62,7 +65,8 @@ export async function portForwardKubernetes (explorerNode?: any): Promise<void>
             // The pod is described by the open document. Skip asking which pod to use and go straight to port-forward.
             const podSelection = portForwardablePods.pod;
             const portMapping = await promptForPort(podSelection);
-            portForwardToPod(podSelection, portMapping);
+            const namespace = portForwardablePods.namespace;
```
bnookala: since this var is only used once, we can probably remove it, and replace the call below with the direct reference to `portForwardablePods.namespace`

shreyu86:
Thanks for the feedback @bnookala , quick question what should be the function in the case of [this condition](https://github.com/shreyu86/vscode-kubernetes-tools/blob/master/src/components/kubectl/port-forward.ts#L92)

bnookala:
@shreyu86 that's a last case condition if the user is trying to port-forward to a pod that is not being described by the current open document. In that case, we should assume that the namespace is  the `current` namespace as retrieved by a call to `await kubectlUtils.currentNamespace(kubectl);`

shreyu86:
Added namespace for the condition described in the above comment.

bnookala:
@shreyu86 looks great! thanks for the contribution. i'll wait for Ivan's sign off to merge this. unfortunately, he's traveling right now so won't be able to address this until monday.

src/extension.ts
```diff
@@ -1319,21 +1329,24 @@ const diffKubernetes = (callback) => {
...
         let kindName = null;
+        let kindObject = null;
         let fileName = null;
 
         let fileFormat = "json";
 
         if (data) {
             fileFormat = (data.trim().length > 0 && data.trim()[0] == '{') ? "json" : "yaml";
-            kindName = findKindNameForText(data);
+            kindObject = findKindNameForText(data);
```
itowlson: Do we need to handle the null case?  What do we do if the object contains a non-default namespace?  (Also applies around line 1348.)

src/extension.ts
```diff
@@ -836,25 +836,35 @@ export function tryFindKindNameFromEditor(): string {
     return findKindNameForText(text);
 }
 
-function findKindNameForText(text): string | null {
+interface FindKindNameForTextResult {
```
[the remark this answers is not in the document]
shreyu86: Changed name to `ResourceKindName`

src/extension.ts
```diff
@@ -836,25 +836,35 @@ export function tryFindKindNameFromEditor(): string {
...
+    readonly namespace: string;
+    readonly resourceType: string;
```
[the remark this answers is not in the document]
shreyu86: changed `resourceType` to `kind`

src/extension.ts
```diff
@@ -836,25 +836,35 @@ export function tryFindKindNameFromEditor(): string {
...
 
 export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
-    let kindName = tryFindKindNameFromEditor();
-    if (kindName === null) {
+    let kindObject = tryFindKindNameFromEditor();
+    if (kindObject === null) {
         promptKindName(resourceKinds, descriptionVerb, opts, handler);
     } else {
-        handler(kindName);
+        handler(`${kindObject.resourceType}/${kindObject.resourceName}`);
```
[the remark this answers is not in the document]
shreyu86: I agree with your comments, can perform some checks on how to best preserve the structure. Happy to address it in future PR.

src/extension.ts
shreyu86: Can you provide some pointers on how you would like to see the `null` case handled?

itowlson: I think this would probably be an early exit, possibly with an error message.  `null` implies that we weren't able to recognise the open document as a Kubernetes resource manifest, so we can't diff it; we should probably explain this to the user, but after that there's not much more we can do.  Does that sound right?  Open to feedback!

shreyu86: Have added early exit (`return`) with an error message `Cannot Parse Kubernetes Object` to address the `null` case

itowlson:
I found a couple of places in `diffKubernetes` that still required changes.  As these are off the main topic of the PR, and probably aren't of great interest, I sent you a PR containing them if you want to apply it to your branch rather than go around another batch of comments and make you do the work!  Let me know if you have any objections to the changes I sent you.  Thanks!

shreyu86:
Thanks a lot @itowlson for the PR, I have merged it.

itowlson:
Thanks heaps @shreyu86!

shreyu86:
Thanks a lot @itowlson and @bnookala for taking the time to review and merging this.

Issue #234, opened 2018-06-12
Diffs fail with resources loaded directly from cluster


itowlson:
If you load a resource directly from a cluster (via the `Load` command) and then try to diff it, even if you save it to the local file system first, it doesn't work.  It looks like the diff code converts the URI to a 'file path' and then back to a file-style URI, which the `vscode.diff` command promptly and justifiably fails to resolve.

The workaround is to save the JSON/YAML to the local file system, close the loaded doc and reopen it from the file system.

This is a consequence of the use of the virtual file system for loading documents, combined with diff code that doesn't handle non-disk URIs.  It doesn't seem like customers have hit it so far (though #218 may relate to it) and the workaround's not too painful, but I'd like to fix it!

itowlson:
Closed via #236.

PR #236, opened 2018-06-13
Fix limitations on Apply command

### LINKED ISSUE #218: Can not apply changes v0.1.9
After editing a deployment you can not apply your changes. When selecting apply a warning pops up saying "Could not find a valid API object"

screen shot 2018-05-17 at 12 28 48 pm

itowlson:
This PR fixes a couple of issues with the Apply command:

* You can't Apply a multi-resource YAML.  This is because Apply wants to show you a diff against the existing cluster configuration first, and we don't yet handle diffing multiple objects.  The fix for this is to cheat: we now put up a warning saying "hey we can't show you what's going to change but would you like to go ahead and apply it anyway?"

* You can't Apply changes made to a resource loaded via Load.  There is a workaround, to save the YAML, close it, open it from the file system and make the changes and apply them; but this PR enables the much simpler workflow of "Load > edit text > Apply".

This PR involved some slightly fiddly and hard to test changes to some pretty old code, so it might be best to consider it WIP for now - would very much appreciate eyes on it though, and suggestions for improving and unifying code paths.  It feels like it's still more fiddly than it ought to be.  If anyone fancies pulling it and trying to break it, that would be great too.

Fixes #218 (hopefully) and #234.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -44,7 +44,7 @@ import * as telemetry from './telemetry-helper';
 import * as extensionapi from './extension.api';
 import {dashboardKubernetes} from './components/kubectl/dashboard';
 import {portForwardKubernetes} from './components/kubectl/port-forward';
-import { Errorable, failed } from './errorable';
+import { Errorable, failed, succeeded } from './errorable';
 import { Git } from './components/git/git';
 import { DebugSession } from './debug/debugSession';
 import { getDebugProviderOfType, getSupportedDebuggerTypes } from './debug/providerRegistry';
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -529,14 +530,22 @@ function maybeRunKubernetesCommandForActiveWindow(command, progressMessage) {
...
+                        kubectlViaTempFile(command, editor.document.getText(), progressMessage, resultHandler);
+                    }
                 });
             }
         });
     } else {
-        const fullCommand = `${command} -f "${editor.document.fileName}"`;
-        console.log(fullCommand);
-        kubectl.invokeWithProgress(fullCommand, progressMessage, resultHandler);
+        if (editor.document.uri.scheme === 'file') {
```
brendandburns: Can we extract this as a function and then remove this code duplication by just calling it in both places?

brendandburns: Alternately, if we convert the promise into an await we might not even need to extract the code...

src/extension.ts
```diff
@@ -1304,18 +1330,73 @@ const deleteKubernetes = async (explorerNode?: explorer.ResourceNode) => {
...
+
+interface DiffResult {
+    readonly result: DiffResultKind;
+    readonly resourceName?: string;
+    readonly stderr?: string;
+    readonly reason?: string;
+}
+
+function confirmOperation(prompt: (msg: string, ...opts: string[]) => Thenable<string>, message: string, confirmVerb: string, operation: () => void) {
+    prompt(message, confirmVerb)
```
brendandburns: await?

brendandburns:
few comments but basically lgtm...

src/extension.ts
itowlson: I'll take the easy route for now but I think you're right - converting this function to async/await would probably simplify the flow a lot.

PR #237, opened 2018-06-13
WIP: Support Show/Follow for Kubernetes Container Logs

shreyu86:
- This is the initial version for the ability to follow/tail logs (#235). 
- Opening an PR for gather early feedback. 
- I have tried to keep it as DRY as possible.

Issue #235, opened 2018-06-12
Feature Request: Ability to Follow/Tail Logs


shreyu86:
Currently the extension prints all the log entries for selected pod, it would be great to be able to follow/tail logs.

bnookala:
Suggesting `Kubernetes: follow logs` (additional behavior) vs. `Kubernetes show logs` (existing behavior)

shreyu86:
Sounds good. I might attempt to get an PR out for this soon.

-Shreyas

shreyu86:
Oops sorry didn't see it was assigned to @bnookala already.

bnookala:
@shreyu86 all yours now :)

veonua:
the tail log is not implemented. so follow logs downloads dozens of MB of log

PR #239, opened 2018-06-13
WIP: Support Show/Follow for Kubernetes Container Logs

shreyu86:
- This is the initial version for the ability to follow/tail logs (#235). 
- Opening an PR for gather early feedback. 
- I have tried to keep it as DRY as possible. 
- Had to abandon #237 because of an rebase issue. (my git knowledge failed me 😅 sorry about that)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,8 @@ If you are building and running the extension from source, see [CONTRIBUTING.md]
 
    * `Kubernetes: Load` - Load a resource from the Kubernetes API and create a new editor window.
    * `Kubernetes: Get` - Get the status for a specific resource.
-   * `Kubernetes: Logs` - Get logs for a pod in an output window.
+   * `Kubernetes: Follow Logs` - Get logs for a pod in an output window.
+   * `Kubernetes: Show Logs` - Follow logs for a pod in an output window.
 
 ### Commands while viewing a Kubernetes file
 
--- a/package.json
+++ b/package.json
@@ -302,6 +302,11 @@
                     "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.pod"
                 },
                 {
+                    "command": "extension.vsKubernetesLoad",
+                    "group": "0",
+                    "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.configmap"
+                },
+                {
                     "command": "extension.vsKubernetesCopy",
                     "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.configmap",
                     "group": "1"
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
         "onCommand:extension.vsKubernetesLoad",
         "onCommand:extension.vsKubernetesGet",
         "onCommand:extension.vsKubernetesRun",
-        "onCommand:extension.vsKubernetesLogs",
+        "onCommand:extension.vsKubernetesShowLogs",
+        "onCommand:extension.vsKubernetesFollowLogs",
         "onCommand:extension.vsKubernetesExpose",
         "onCommand:extension.vsKubernetesDescribe",
         "onCommand:extension.vsKubernetesSync",
@@ -291,14 +292,14 @@
                     "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.pod"
                 },
                 {
-                    "command": "extension.vsKubernetesLogs",
+                    "command": "extension.vsKubernetesShowLogs",
                     "group": "3",
                     "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.pod"
                 },
                 {
-                    "command": "extension.vsKubernetesLoad",
-                    "group": "0",
-                    "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.configmap"
+                    "command": "extension.vsKubernetesFollowLogs",
+                    "group": "3",
+                    "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.pod"
                 },
                 {
                     "command": "extension.vsKubernetesCopy",
@@ -429,8 +430,13 @@
                 "category": "Kubernetes"
             },
             {
-                "command": "extension.vsKubernetesLogs",
-                "title": "Logs",
+                "command": "extension.vsKubernetesShowLogs",
+                "title": "Show Logs",
+                "category": "Kubernetes"
+            },
+            {
+                "command": "extension.vsKubernetesFollowLogs",
+                "title": "Follow Logs",
                 "category": "Kubernetes"
             },
             {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -127,7 +127,8 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
         registerCommand('extension.vsKubernetesLoad', loadKubernetes),
         registerCommand('extension.vsKubernetesGet', getKubernetes),
         registerCommand('extension.vsKubernetesRun', runKubernetes),
-        registerCommand('extension.vsKubernetesLogs', logsKubernetes),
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -291,14 +292,14 @@
                     "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.pod"
                 },
                 {
-                    "command": "extension.vsKubernetesLogs",
+                    "command": "extension.vsKubernetesShowLogs",
                     "group": "3",
                     "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.pod"
                 },
                 {
-                    "command": "extension.vsKubernetesLoad",
```
bnookala: I think this may have been accidentally deleted.

src/extension.ts
```diff
@@ -127,7 +127,8 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
         registerCommand('extension.vsKubernetesLoad', loadKubernetes),
         registerCommand('extension.vsKubernetesGet', getKubernetes),
         registerCommand('extension.vsKubernetesRun', runKubernetes),
-        registerCommand('extension.vsKubernetesLogs', logsKubernetes),
+        registerCommand('extension.vsKubernetesShowLogs', logsKubernetes),
+        registerCommand('extension.vsKubernetesFollowLogs', (obj) => {logsKubernetes(obj, true);}),
```
bnookala: hmm - let's try to give `obj` a better name here.

and also the type - looks like it's `explorer.ResourceNode`

src/extension.ts
```diff
@@ -1048,37 +1049,40 @@ async function selectPod(scope: PodSelectionScope, fallback: PodSelectionFallbac
...
     if (explorerNode) {
         const podSummary = { name: explorerNode.id, namespace: undefined };  // TODO: namespaces
         const container = await selectContainerForPod(podSummary);
         if (container) {
-            getLogsForContainer(podSummary.name, podSummary.namespace, container.name);
+            getLogsForContainer(podSummary.name, podSummary.namespace, container.name, follow);
         }
     } else {
-        findPod(getLogsForPod);
+        findPod((obj) => {getLogsForPod(obj, follow);});
```
bnookala: better name (and type) for `obj` here.

package.json
shreyu86: Sorry about that, I selected that line by mistake.

src/extension.ts
shreyu86: added meaningful name and type

src/extension.ts
shreyu86: added meaningful name and type

shreyu86:
Updated Readme to reflect the changes.

bnookala:
Looks like a clean, straightforward change, so I went ahead and merged. Thanks for the contribution!

README.md
```diff
@@ -74,7 +74,8 @@ If you are building and running the extension from source, see [CONTRIBUTING.md]
 
    * `Kubernetes: Load` - Load a resource from the Kubernetes API and create a new editor window.
    * `Kubernetes: Get` - Get the status for a specific resource.
-   * `Kubernetes: Logs` - Get logs for a pod in an output window.
+   * `Kubernetes: Follow Logs` - Get logs for a pod in an output window.
+   * `Kubernetes: Show Logs` - Follow logs for a pod in an output window.
```
itowlson: These descriptions seem to be the wrong way round?

src/extension.ts
```diff
@@ -127,7 +127,8 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
         registerCommand('extension.vsKubernetesLoad', loadKubernetes),
         registerCommand('extension.vsKubernetesGet', getKubernetes),
         registerCommand('extension.vsKubernetesRun', runKubernetes),
-        registerCommand('extension.vsKubernetesLogs', logsKubernetes),
+        registerCommand('extension.vsKubernetesShowLogs', logsKubernetes),
+        registerCommand('extension.vsKubernetesFollowLogs', (explorerNode: explorer.ResourceNode) => {logsKubernetes(explorerNode, true);}),
```
itowlson: I'm not keen on this style difference - it kind of disguises what the difference is between the two invocations (the unassuming second argument, rather than the big noisy ExplorerNode that appears three times in the text).  I'd suggest we either do both of them lambda style (with an explicit second argument in both cases), or create two functions (`showLogsKubernetes` and `followLogsKubernetes`) to be command handlers, both delegating to a core function.  (I prefer the latter but am okay with the former.)

src/extension.ts
```diff
@@ -1048,37 +1049,40 @@ async function selectPod(scope: PodSelectionScope, fallback: PodSelectionFallbac
...
     if (explorerNode) {
         const podSummary = { name: explorerNode.id, namespace: undefined };  // TODO: namespaces
         const container = await selectContainerForPod(podSummary);
         if (container) {
-            getLogsForContainer(podSummary.name, podSummary.namespace, container.name);
+            getLogsForContainer(podSummary.name, podSummary.namespace, container.name, follow);
         }
     } else {
-        findPod(getLogsForPod);
+        findPod((pod: PodSummary) => {getLogsForPod(pod, follow);});
```
itowlson: I don't know if we need the type declaration here - think TypeScript should infer it from `findPod` and `getLogsForPod`.  But if not then keep it.

src/extension.ts
```diff
@@ -1048,37 +1049,40 @@ async function selectPod(scope: PodSelectionScope, fallback: PodSelectionFallbac
...
+async function logsKubernetes(explorerNode?: explorer.ResourceNode, follow?: boolean) {
     if (explorerNode) {
         const podSummary = { name: explorerNode.id, namespace: undefined };  // TODO: namespaces
         const container = await selectContainerForPod(podSummary);
         if (container) {
-            getLogsForContainer(podSummary.name, podSummary.namespace, container.name);
+            getLogsForContainer(podSummary.name, podSummary.namespace, container.name, follow);
         }
     } else {
-        findPod(getLogsForPod);
```
itowlson: Do we need the braces around the lambda body, given that it's only one expression?

src/extension.ts
```diff
@@ -1048,37 +1049,40 @@ async function selectPod(scope: PodSelectionScope, fallback: PodSelectionFallbac
...
-function getLogsForContainer(podName: string, podNamespace: string | undefined, containerName: string | undefined) {
+function getLogsForContainer(podName: string, podNamespace: string | undefined, containerName: string | undefined, follow?: boolean) {
     let cmd = 'logs ' + podName;
     if (podNamespace && podNamespace.length > 0) {
         cmd += ' --namespace=' + podNamespace;
     }
     if (containerName) {
         cmd += ' --container=' + containerName;
     }
+    if (follow){
```
itowlson: Space before brace please

src/extension.ts
```diff
@@ -1048,37 +1049,40 @@ async function selectPod(scope: PodSelectionScope, fallback: PodSelectionFallbac
     return value.pod;
 }
 
-async function logsKubernetes(explorerNode?: explorer.ResourceNode) {
+async function logsKubernetes(explorerNode?: explorer.ResourceNode, follow?: boolean) {
```
itowlson: I'm a bit allergic to Booleans in this context, because the meaning is not obvious at point of call.  (One of my colleagues wrote http://www.nichesoftware.co.nz/2016/05/15/code-gardening-bools-are-evil.html which pretty much sums up my feelings.)  I'd suggest we declare an enum and make it mandatory.

shreyu86:
Fixed the documentation and linting issue in #240, happy to follow up on the other issues as well.

src/extension.ts
shreyu86: Thanks for the reference article, really helps to understand the refactor path.

itowlson:
@shreyu86 Thought you might appreciate this bit of user feedback on the log tailing feature: "I would say it's the best feature of the K8s extension".  Nice one!  https://twitter.com/codycraven/status/1046919046682140672?s=20

shreyu86:
Oh wow awesome great to hear the feedback.

PR #243, opened 2018-06-19
add gulp task to generate enum json for kubernetes schema.

andxu:


andxu:
See comments in https://github.com/Azure/vscode-kubernetes-tools/pull/228

andxu:
https://github.com/Azure/vscode-kubernetes-tools/issues/175

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+                        }
+                    }
+
+                } else {
+                    // ignore
+                }
+                i++;
+            }
+        }
+    }
```
itowlson: I feel like at this point we've ended up with a parsed representation of the relevant parts of the Go file, but I don't think I understand the representation, and I find it hard to work out from the regexes.  Breaking the entire preceding section out into a named function with some documentation of the returned representation would make it easier to follow, and would help readers to understand the next bit too.

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+    const json = JSON.parse(swagger_json);
+    const map = {};
+
+    _.each(Object.keys(typeList), key => {
```
itowlson: Any reason for using `lodash` rather than a `for ... in` or `for ... of` statement?

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+        let t = typeList[key];
+        _.each(t, field => {
```
itowlson: This line implies to me that `t` is a collection of fields?  So `typeList` is a map from what, type names to the their fields?

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+
+                } else {
+                    // ignore
+                }
+                i++;
+            }
+        }
+    }
+    const json = JSON.parse(swagger_json);
+    const map = {};
```
itowlson: A map of what?

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+        let t = typeList[key];
+        _.each(t, field => {
+            if (enumTypeList[field.type]) {
+                // console.log('Got:', key, field.name, enumTypeList[field.type].join('|'))
+                map[key + '!' + field.name] = {
+                    enums: enumTypeList[field.type]
+                }
+            }
+        });
+    });
```
itowlson: Again this feels like an end point where we have computed something new, and could profitable extract the block into something meaningfully named and with documentation about the returned map.

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+    let current;
+    const enums = {};
+    _.each(json.definitions, (definitions, key) => {
+        current = key.split('.').slice(-1)[0];
```
itowlson: Can't we `let current` here?  I thought the point of `let` was to allow block scoping.

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+        enums[key] = {};
+        _.each(definitions.properties, (propValue, propKey) => {
+            if (map[current + '!' + propKey]) {
+                if (!propValue.enum) {
+                    propValue.enum = map[current + '!' + propKey].enums;
+                    enums[key][propKey] = propValue.enum;
+                }
+            }
+        });
+        if (Object.keys(enums[key]).length === 0) {
```
itowlson: I feel this would be clearer if you built the object in the `each` block as a named variable (extracting to a function and returning the built object), then assigned it to `enums[key]` if the object was non-empty.

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
+const fs = require('fs');
+const _ = require('lodash');
+const download = require('download');
+
+const COMMENT_REGEX = /(\/\*([\s\S]*?)\*\/)|(\/\/(.*)$)/gm;
+
+function parse_protocol_buf(field) {
```
itowlson: Perhaps give an example of the kind of input this function handles?  That might make it easier for the reader to see what things like the regexes and the magic number array indexes are doing.

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
+const fs = require('fs');
```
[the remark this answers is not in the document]
andxu: sure

generate_kubernetes_enums.js
```diff
@@ -0,0 +1,145 @@
...
+    });
+    return enums;
+}
+
+function download_to_memory(_url) {
+    return download(_url).then(buf => buf.toString('utf-8'));
+}
+
+function generate_kubernetes_enums(targetFile) {
+    const type_go_url = 'https://raw.githubusercontent.com/kubernetes/kubernetes/master/staging/src/k8s.io/api/core/v1/types.go';
```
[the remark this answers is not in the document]
andxu: It is hard to determine the types.go for each versions, while the code is to generate enums, not the whole schema, so enums should not have any significant changes across different version, it is still acceptable to use the latest one for generating enums.json

itowlson:
This has been folded into #583 - closing this iteration.

PR #252, opened 2018-06-25
Add node expansion and listing of pods for nodes.

brendandburns:
@itowlson this mostly works, but selecting the pods doesn't work correctly. I see the virtual FS that does this, but I don't know how it is triggered, so I don't know how to pass the right namespace along...

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -7,6 +7,7 @@ import * as kubectlUtils from './kubectlUtils';
 import { Host } from './host';
 import * as kuberesources from './kuberesources';
 import { failed } from './errorable';
+import { filter } from 'minimatch';
 
 export function create(kubectl: Kubectl, host: Host): KubernetesExplorer {
     return new KubernetesExplorer(kubectl, host);
@@ -89,7 +90,7 @@ class KubernetesCluster implements KubernetesObject {
     getChildren(kubectl: Kubectl, host: Host): vscode.ProviderResult<KubernetesObject[]> {
         return [
             new KubernetesNamespaceFolder(),
-            new KubernetesResourceFolder(kuberesources.allKinds.node),
+            new KubernetesNodeFolder(),
             new KubernetesWorkloadFolder(),
             new KubernetesServiceFolder(),
             new KubernetesResourceFolder(kuberesources.allKinds.ingress),
@@ -195,6 +196,40 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
     }
 }
 
+class KubernetesNodeFolder extends KubernetesResourceFolder {
+    constructor() {
+        super(kuberesources.allKinds.node);
+    }
+
+    async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        const nodes = await kubectlUtils.getGlobalResources(kubectl, 'nodes');
+        return nodes.map((node) => new KubernetesNodeResource(node.name, node));
+    }
+}
+
+class KubernetesNodeResource extends KubernetesResource {
+    constructor(name: string, meta: any) {
+        super(kuberesources.allKinds.node, name, meta);
+    }
+
+    async getTreeItem(): Promise<vscode.TreeItem> {
+        const treeItem = await super.getTreeItem();
+        treeItem.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+        return treeItem;
+    }
+
+    async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        const pods = await kubectlUtils.getPods(kubectl, null, 'all');
+        const filteredPods = [];
+        pods.map((p) => {
+            if ('node/' + p.nodeName === this.resourceId) {
+                filteredPods.push(p);
+            }
+        });
+        return filteredPods.map((p) => new KubernetesResource(kuberesources.allKinds.pod, p.name, p));
+    }
+}
+
 class KubernetesNamespaceFolder extends KubernetesResourceFolder {
     constructor() {
         super(kuberesources.allKinds.namespace);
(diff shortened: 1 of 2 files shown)


src/kubectlUtils.ts
```diff
@@ -112,6 +115,19 @@ export async function getDataHolders(resource: string, kubectl: Kubectl): Promis
     return depList.result.items;
 }
 
+export async function getGlobalResources(kubectl: Kubectl, resource: string): Promise<KubernetesObject[]> {
+    const rsrcs = await kubectl.asJson<KubernetesCollection<Namespace>>(`get ${resource} -o json`);
```
itowlson: Collection of `Namespace`?  Should it be `KubernetesObject`?

src/kubectlUtils.ts
```diff
@@ -151,16 +167,18 @@ export async function getPodSelector(resource: string, kubectl: Kubectl): Promis
...
-export async function getPods(kubectl: Kubectl, selector: any): Promise<PodInfo[]> {
-    if (!selector) {
-        return [];
+export async function getPods(kubectl: Kubectl, selector: any, namespace: string = null): Promise<PodInfo[]> {
+    let ns = namespace;
+    if (!ns) {
+        ns = await currentNamespace(kubectl);
+    }
+    let nsFlag = `--namespace=${ns}`
+    if (ns === 'all') {
```
itowlson: Nit: Is `all` a legal name for a Kubernetes namespace?  It looks like this would do the wrong thing if someone was using a namespace with this name... unlikely as it may seem!

src/explorer.ts
```diff
@@ -195,6 +196,40 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+    async getTreeItem(): Promise<vscode.TreeItem> {
+        const treeItem = await super.getTreeItem();
+        treeItem.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+        return treeItem;
+    }
+
+    async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        const pods = await kubectlUtils.getPods(kubectl, null, 'all');
+        const filteredPods = [];
+        pods.map((p) => {
```
itowlson: Can this use `filter` instead of a `map` with a side-effectful `push`?  I think that would be more self-documenting...

src/kubectlUtils.ts
```diff
@@ -151,16 +167,18 @@ export async function getPodSelector(resource: string, kubectl: Kubectl): Promis
...
 }
 
-export async function getPods(kubectl: Kubectl, selector: any): Promise<PodInfo[]> {
-    if (!selector) {
-        return [];
+export async function getPods(kubectl: Kubectl, selector: any, namespace: string = null): Promise<PodInfo[]> {
+    let ns = namespace;
+    if (!ns) {
+        ns = await currentNamespace(kubectl);
+    }
```
itowlson: Would `const ns = namespace || await currentNamespace(kubectl);` do the same thing?  This seems to be the JS idiom for null coalescence.

src/explorer.ts
```diff
@@ -195,6 +196,40 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+            if ('node/' + p.nodeName === this.resourceId) {
+                filteredPods.push(p);
+            }
+        });
+        return filteredPods.map((p) => new KubernetesResource(kuberesources.allKinds.pod, p.name, p));
```
itowlson: The `get` command fetches pods from all namespaces, but the existing `KubernetesResource` class doesn't consider the namespace when loading the resource.  We'll need to add this I think...

itowlson:
@brendandburns You _shouldn't_ need to do anything around selection, because selection delegates to the `Load` command which does all the work (and knows how to form the URIs for the virtual file system).  Are the pods you tested on in the active namespace?  There is an limitation at the moment where we don't pass a namespace to the `Load` command and so it always issues a `get` _without namespace flag_ for the resource YAML/JSON... we will need to update the `Load` command and the VFS URIs to handle this.  I can plumb that in unless you fancy it...

src/kubectlUtils.ts
```diff
@@ -151,16 +167,18 @@ export async function getPodSelector(resource: string, kubectl: Kubectl): Promis
...
+    let nsFlag = `--namespace=${ns}`
```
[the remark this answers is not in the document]
brendandburns: done.

src/explorer.ts
```diff
@@ -195,6 +196,40 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+        const treeItem = await super.getTreeItem();
+        treeItem.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+        return treeItem;
+    }
+
+    async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        const pods = await kubectlUtils.getPods(kubectl, null, 'all');
+        const filteredPods = [];
+        pods.map((p) => {
+            if ('node/' + p.nodeName === this.resourceId) {
```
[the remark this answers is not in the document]

src/kubectlUtils.ts
brendandburns: Switched to `KubernetesResource`

src/kubectlUtils.ts
brendandburns: Do you have a better magic value?

src/explorer.ts

src/kubectlUtils.ts
brendandburns: done.

src/explorer.ts
brendandburns: Yeah, I think that is the bug I was seeing...

brendandburns:

Thanks!

src/kubectlUtils.ts
itowlson: The magic string doesn't need to be a valid namespace name does it?  So it could be something like `*all*` perhaps (assuming namespaces need to be DNS-valid?)?

brendandburns:
@itowlson It's up to you, if you can tell me how to fix it, I'm happy to add it to this PR, but it was kind of opaque to me...

bnookala:
@brendandburns let's file it as a separate issue, as Ivan had suggested.

Issue #244, opened 2018-06-20
Status Indicators for K8s Resources in Explorer View


shreyu86:
I have been using the extension daily and have found it extremely useful to interact with Kubernetes. Also using the `vscode-docker` extension for local development and container management. 
One of the features from `vscode-docker` is the visual indicators for showing status of running container (running/error) is neat and was thinking about seeing similar experience for pods and other relevant Kubernetes resources. 

After looking at the code and the importing the icons from `vscode-docker` was able to add visual indicators to pods in the explorer view and was wondering if other users would like to see such indicators. 

![image](https://user-images.githubusercontent.com/311217/41641322-c1d64c12-7419-11e8-9f49-c5e157698752.png)

Happy to put together PR if there is utility in such indicators on Kubernetes resources in the explorer view.

shreyu86:
@bnookala shall I try sending an POC PR for this feature?

bnookala:
@shreyu86 happy to take a look at your pull request!

PR #254, opened 2018-06-26
Status Indicators for K8s Pods and Jobs

shreyu86:
This PR implements the enhancement proposed in #244 towards pods and jobs. 
This change adds and indicator for a pod/job state (error/running/completed). 
This can be extended to include initialization state too. For this PR used the existing methods and added ability to pass the pod/job status to the explorer.


src/explorer.ts
```diff
@@ -189,7 +193,14 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
         if (this.kind === kuberesources.allKinds.pod ||
             this.kind == kuberesources.allKinds.secret ||
             this.kind == kuberesources.allKinds.configMap) {
-            treeItem.contextValue = `vsKubernetes.resource.${this.kind.abbreviation}`;
+			treeItem.contextValue = `vsKubernetes.resource.${this.kind.abbreviation}`;
+            if (this.kind === kuberesources.allKinds.pod && this.metadata.status != null) {
```
bnookala: move this logic into its own function ie. 

```
if (this.kind === kuberesources.allKinds.pod) {
     treeItem.iconPath = getIconForPodStatus(this.metadata.status);
}
```

it also appears that your editor is using tabs instead of spaces - I think we've settled on using 4 space tabs in the project.

src/explorer.ts
```diff
@@ -161,6 +160,10 @@ class KubernetesResourceFolder extends KubernetesFolder {
             return [new DummyObject("Error")];
         }
         return childrenLines.result.map((line) => {
+            if (this.kind.abbreviation==="pod"){
+            const bits = line.replace(/\s+/g,'|').split('|');
```
bnookala: I'm a little bit iffy about using regex to do data massaging for tabular formats... could you change the above

can you change the getChildren function to get data from kubectl as json? the Kubectl interface offers an `asJson` function. here's an example of its usage:

`const shellResult = await kubectl.asJson<any>("config view -o json");`

src/explorer.ts
shreyu86: Will fix this, does the project use an `.editorconfig` ?

src/explorer.ts
shreyu86: Sounds good! Will try `asJson`

src/explorer.ts
```diff
@@ -161,6 +160,10 @@ class KubernetesResourceFolder extends KubernetesFolder {
             return [new DummyObject("Error")];
         }
         return childrenLines.result.map((line) => {
+            if (this.kind.abbreviation==="pod"){
```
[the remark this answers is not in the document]
shreyu86: I had an issue with the formatting setup incorrectly, fixing all the formatting issues.

src/explorer.ts
```diff
@@ -161,6 +160,10 @@ class KubernetesResourceFolder extends KubernetesFolder {
...
+            const bits = line.replace(/\s+/g,'|').split('|');
+            return new KubernetesResource(this.kind, bits[0], {status: bits[2].toLowerCase()});
+            }
```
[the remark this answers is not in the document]
shreyu86: Once I fix formatting I will check this as well.

src/explorer.ts
```diff
@@ -172,6 +175,7 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
 
     constructor(readonly kind: kuberesources.ResourceKind, readonly id: string, readonly metadata?: any) {
         this.resourceId = kind.abbreviation + '/' + id;
+        this.metadata = metadata;
```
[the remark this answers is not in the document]
shreyu86: Oh cool!

src/explorer.ts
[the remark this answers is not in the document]
shreyu86: Current implementation includes pods (just pods and pods originating from from job execution) hence the title reference to both jobs and pods.

shreyu86:
@bnookala and @itowlson Thanks for the feedback, I will fix the formatting issues. What is the current formatting setup that this project uses, any extensions for formatting that this project development team use?

src/explorer.ts
shreyu86: thinking about how to switch to `asJson` and get the same data as with `asLines`. I can retrieve the `ID` but getting an consolidated pod status might require additional parsing and aggregation/sorting of the JSON output as with JSON the output contains individual `containerStatuses` in the pod. Let me know if you have any pointers regarding the approach.

src/explorer.ts
```diff
@@ -190,6 +201,9 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
             this.kind == kuberesources.allKinds.secret ||
             this.kind == kuberesources.allKinds.configMap) {
             treeItem.contextValue = `vsKubernetes.resource.${this.kind.abbreviation}`;
+            if (this.kind === kuberesources.allKinds.pod && this.metadata.status != null) {
```
bnookala: missed this - use the stricter equality check:

`this.metadata.status !== null`

src/explorer.ts
bnookala: @shreyu86 I think you'd want `status.phase` from the resulting json.

src/explorer.ts
```diff
@@ -126,7 +134,6 @@ class KubernetesWorkloadFolder extends KubernetesFolder {
     constructor() {
         super("workload", "Workloads");
     }
-
```
itowlson: Did a blank separator line accidentally get removed here?  Would prefer it to stay I think.

itowlson:
Something that occurred to me -- @shreyu86 is the icon artwork original with you?  If not, under what licensing terms is it distributed?  I want to make sure we're legit before the Microsoft legal eagles have me up in the Star Chamber...

src/explorer.ts
shreyu86: Sorry about that, fixed this.

src/explorer.ts
shreyu86: Thanks for catching that, fixed this.

src/explorer.ts
shreyu86: Tried to refactor with `status.phase` but an issue more of a `kubectl` one that I ran into was that `status.phase` (in my case `status.phase` was always `running` even if pod was crashing ) was that "The phase is not intended to be a comprehensive rollup of observations of Container or Pod state, nor is it intended to be a comprehensive state machine." and hence preferred the current output as it performs an rollup. What I might try to do is write a function to perform the rollup in an followup PR after doing some prototyping.

bnookala:
Thank you!

PR #263, opened 2018-06-28
Show Helm releases in cluster tree view

### LINKED ISSUE #245: Add Helm releases node to cluster tree view
Maps to the `helm list` command.

Note: test with large numbers of releases (we may need to use `--max` and `--offset`).

itowlson:
This adds a new top-level 'Helm Releases' node to clusters in the tree view.  At the moment there's not much you can do with this - viewing the node displays the `helm get` output as a text file, and that's it.

This is currently work in progress - it needs to be tested with large numbers of releases, and I would like feedback on whether to present the returned file as YAML (which it isn't, not really, but it colourises okay if you pretend it is) or as plain text.  Other feedback / suggestions welcome (but may be deferred to subsequent enhancements).

Fixes #245.

### DIFF
--- a/src/helm.exec.ts
+++ b/src/helm.exec.ts
@@ -6,6 +6,7 @@ import * as YAML from 'yamljs';
 import * as _ from 'lodash';
 import * as fs from "fs";
 import * as extension from './extension';
+import * as explorer from './explorer';
 import * as helm from './helm';
 import { showWorkspaceFolderPick } from './hostutils';
 import { shell as sh, ShellResult } from './shell';
@@ -140,6 +141,20 @@ export function helmDryRun() {
                 logger.log("⎈⎈⎈ INSTALL FAILED");
             }
         });
+    });
+}
+
+export function helmGet(resourceNode: explorer.ResourceNode) {
+    if (!resourceNode) {
+        return;
+    }
+    const releaseName = resourceNode.id.split(':')[1];
+    helmExec(`get ${releaseName}`, (code, stdout, stderr) => {
+        if (code !== 0) {
+            vscode.window.showErrorMessage(`Helm get failed: ${stderr}`);
+            return;
+        }
+        console.log(stdout);
     });
 }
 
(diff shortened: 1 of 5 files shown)


itowlson:
Huh, okay, so `helm list` actually returns releases in _all_ namespaces (whereas `helm install` does install into the current `kubectl` namespace), so I made _at least_ one wrong assumption there.  I kind of feel like our display should filter to the active namespace, otherwise it's going to be a bit misleading - but perhaps this would be confusing to Helm users...

itowlson:
After discussion with a Helm maintainer, the recommendation is to filter to the active namespace.

itowlson:
Seems to work okay with 500+ releases in the cluster.  I'm going to call that acceptable for our intended use case for now, and address paging if a customer hits a limit.

itowlson:
As ready to go as it's gonna get without some feedback.  @bnookala @technosophos @bacongobbler do you have time to take a look please?

src/helm.exec.ts
```diff
@@ -260,6 +279,57 @@ export function helmExec(args: string, fn) {
...
+    const bin = configuredBin ? `"${configuredBin}"` : "helm";
+    const cmd = `${bin} ${args}`;
+    return await sh.exec(cmd);
+}
+
+const HELM_PAGING_PREFIX = "next:";
+
+export async function helmListAll(namespace?: string): Promise<Errorable<string[]>> {
+    if (!ensureHelm(EnsureMode.Alert)) {
+        return { succeeded: false, error: [ 'Helm not installed' ] };
```
bnookala: Let's make this error consistent with the one from the explorer.

src/kuberesources.virtualfs.ts
```diff
@@ -65,6 +70,18 @@ export class KubernetesResourceVirtualFileSystemProvider implements FileSystemPr
         return sr.stdout;
     }
 
+    async execLoadResource(loadMode: string, ns: string | undefined, value: string, outputFormat: string): Promise<ShellResult> {
```
bnookala: call this `resourceAuthority` instead of `loadMode`

itowlson:
Done, please re-review - thanks!

PR #266, opened 2018-06-29
Display context name instead of cluster name in tree view

### LINKED ISSUE #178: The name of added ACS is different from the real name on portal
Vscode Kubernetes:
image
Portal:
image

itowlson:
ACS and AKS tend to provide fairly user-hostile cluster/context names.  Contexts can be renamed using `kubectl config rename-context`, but clusters can't.  So it is preferable to display context names in the tree view so that users can replace them with beautified ones.

I also took the opportunity to rename the tree to **Clusters**, because the Kubernetes label is already prepended by virtue of being on the new Activity Bar tab.

Fixes #178.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
             "kubernetesView": [
                 {
                     "id": "extension.vsKubernetesExplorer",
-                    "name": "Kubernetes"
+                    "name": "Clusters"
                 }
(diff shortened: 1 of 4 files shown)


shreyu86:
This feature will be awesome!

PR #269, opened 2018-06-30
Implementing Show/Follow events at Namespace level

kameshsampath:
Resolves Issue#268

This allows the user to show or follow Kubernetes events via `kubectl get events ` 

Signed-off-by: Kamesh Sampath <kamesh.sampath@hotmail.com>


kameshsampath:
@itowlson, I have to do a force push as I have stripped off all the formatting issues, just noticed they happened because of my user settings having the flag `"editor.formatOnSave": true,"files.autoSave": "true"`. I wonder if we could disable these forcibiliy via settings.json file of the sources to avoid any such issue from other contributors. WDYT ?

I am just trying to find out where other nodes in the tree where we can show the events, so that we can filter them by 
- kind 
- type 
- name 

Apart from this I have pushed changes with all your comments incorporated

kameshsampath:
with respect to the events filtering, the following command filters the events by kind=Pod

```kubectl get events -o=jsonpath='{range .items[?(@.involvedObject.kind == "Pod")]}{"\n"}{"message:"}{.message}{",source:"}{.source.component}{",count"}{.count}{",firstSeen:"}{.firstTimestamp}{",lastSeen"}{.lastTimestamp}{end}'```

This can be extended to other `properties` within the output of the `get events command `

The only thing am wondering about how to show them. one crude idea is to use `virtualdocuments` and decorate them as HTML.  But then another blocker here would be to think how to do `follow` ?

@itowlson  - thoughts ?

File: src/extension.ts [hunk 1]
```diff
@@ -1097,6 +1098,24 @@ async function getLogsForPod(pod: PodSummary, follow?: boolean) {
     }
 }
 
+async function getEvents(follow: kuberesources.EventDisplayMode, explorerNode?: explorer.ResourceNode) {
```
bnookala: extension.ts is getting too big, so we should componentize as much as possible. move this to `src/components/kubectl/events.ts`

src/extension.ts
```diff
@@ -1097,6 +1098,24 @@ async function getLogsForPod(pod: PodSummary, follow?: boolean) {
...
+    let currentNS;
```
bnookala: rename this var, since it's not always "current" if the explorerNode is provided.

src/extension.ts
```diff
@@ -1097,6 +1098,24 @@ async function getLogsForPod(pod: PodSummary, follow?: boolean) {
...
+
+    if (explorerNode) {
+        currentNS = explorerNode.id;
+    } else {
+        // defaulting to current namespace in kube config
+        currentNS = await kubectlUtils.currentNamespace(kubectl);
+    }
+
+    let cmd = ` get events --namespace ${currentNS}`;
+    if (kuberesources.EventDisplayMode.Follow === follow) {
```
bnookala: flip this condition.

src/extension.ts
bnookala: rename the `follow` variable, since it doesn't describe the other condition (show).

bnookala: flip the variables, explorerNode should come first since that's the namespace being operated on, with the enum second.

src/extension.ts
```diff
@@ -1097,6 +1098,24 @@ async function getLogsForPod(pod: PodSummary, follow?: boolean) {
...
+        cmd += ' -w';
+    }
+
+    kubectl.invokeInSharedTerminal(cmd);
```
itowlson: We probably don't want to use the shared terminal in `follow` mode, because it blocks the shared terminal from being used for other commands such as logs.  Follow mode should start a new terminal with a suitable descriptive name; see the `kubectl.invokeInNewTerminal` method.

This is also a bug in the existing `Follow Logs` command (and I appreciate you following the pattern; sorry we missed this on the logs code).  If you can address it here then we can backport your fixes for both the terminal blocking issue and the Boolean parameter issue onto the existing logs code.

itowlson:
@kameshsampath That filtering stuff looks a bit forbidding.  Let's leave it as namespace only for now; we can later see if it's worth adding per-resource events as a separate piece of work.  Thanks for looking into it!

src/extension.ts
bnookala: I'll file an issue for this

bnookala:
@itowlson I don't feel strongly either way. I think we should probably run a project level format just to be sure. I'd expect any future changes will be formatted "properly" as the editor config should help vscode point out those issues during editing.

src/extension.ts
kameshsampath: I was to ask about that , event we should think of moving lots of logical components out of extension.ts like `kuberrnetsCommands.ts`, `helmCommands.ts` etc., at first rather even now I find it hard to read the extension.ts :)

kameshsampath: the reason for that was we can allow it to be invoked from command palette as well where  we can make it to show events of current Namespace. with that logic behind the optional explorerNode cant be first argument

kameshsampath: thought to do it but somehow got slipped out of my mind

src/extension.ts
kameshsampath: what do you mean by flip here ?

kameshsampath:
@bnookla @itowlson - big +1 for editor config and have been using for long time, worth another issue ? so that we can discuss in more detail ?

src/kubeEvents.ts
```diff
@@ -0,0 +1,26 @@
...
+        eventsNS = await kubectlUtils.currentNamespace(kubectl);
+    }
+
+    let cmd = ` get events --namespace ${eventsNS}`;
+    if (EventDisplayMode.Follow === displayMode) {
+        cmd += ' -w';
+    }
+
+    kubectl.invokeInNewTerminal(cmd, 'Kubernetes Events');
+
```
bnookala: remove this newline

src/extension.ts
```diff
@@ -25,6 +25,7 @@ import * as configmaps from './configMap';
 import * as configureFromCluster from './configurefromcluster';
 import * as createCluster from './createcluster';
 import * as kuberesources from './kuberesources';
+import {EventDisplayMode,getEvents} from './kubeEvents';
```
bnookala: { EventDisplayMode, getEvents }

src/extension.ts
```diff
@@ -156,7 +157,8 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
         registerCommand('extension.vsKubernetesLoadConfigMapData', configmaps.loadConfigMapData),
         registerCommand('extension.vsKubernetesDeleteFile', (obj) => { deleteKubernetesConfigFile(kubectl, obj, treeProvider); }),
         registerCommand('extension.vsKubernetesAddFile', (obj) => { addKubernetesConfigFile(kubectl, obj, treeProvider); }),
-
+        registerCommand('extension.vsKubernetesShowEvents', (explorerNode: explorer.ResourceNode) => { getEvents(kubectl,EventDisplayMode.Show,explorerNode); }),
```
bnookala: spaces between arguments

src/extension.ts
```diff
@@ -156,7 +157,8 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
...
+        registerCommand('extension.vsKubernetesFollowEvents', (explorerNode: explorer.ResourceNode) => { getEvents(kubectl,EventDisplayMode.Follow,explorerNode); }),
```
bnookala: spaces between arguments

src/kubeEvents.ts
```diff
@@ -0,0 +1,25 @@
...
+export async function getEvents(kubectl: Kubectl, displayMode: EventDisplayMode, explorerNode?: explorer.ResourceNode) {
+    let eventsNS;
+
+    if (explorerNode) {
+        eventsNS = explorerNode.id;
+    } else {
+        eventsNS = await kubectlUtils.currentNamespace(kubectl);
+    }
+
+    let cmd = ` get events --namespace ${eventsNS}`;
```
itowlson: You don't need the leading space here - the `kubectl` methods handle this for you.

src/kubeEvents.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (EventDisplayMode.Follow === displayMode) {
+        cmd += ' -w';
+    }
+
+    kubectl.invokeInNewTerminal(cmd, 'Kubernetes Events');
```
itowlson: I'd suggest doing an `invokeInSharedTerminal` for Show (to avoid spawning a new terminal for every show) and `invokeInNewTerminal` for Follow (to avoid blocking the shared terminal).

kameshsampath: i get you now, sorry I should have handled that way.

kameshsampath:
@itowlson @bnookala - I did a rebase and squash of few commits to resolve some conflicts arising out of other merges. Sorry for any inconvenience here.

kameshsampath:
thanks @itowlson , @bnookala  if it looks good to you please go ahead and merge the same when you are at it .

src/kubeEvents.ts
```diff
@@ -0,0 +1,27 @@
+import { Kubectl } from './kubectl';
```
bnookala: this file needs to be at src/components/kubectl/events.ts

src/kubeEvents.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    if (explorerNode) {
+        eventsNS = explorerNode.id;
+    } else {
+        eventsNS = await kubectlUtils.currentNamespace(kubectl);
+    }
+
+    let cmd = `get events --namespace ${eventsNS}`;
+
+    if (EventDisplayMode.Follow === displayMode) {
```
bnookala: reverse the condition: `displayMode === EventDisplayMode.Follow`

kameshsampath: its a common practice to have non-nullable value in LHS, since this style is followed here I can reverse it

Issue #265, opened 2018-06-28
Kubernetes: set namespace command


bnookala:
We currently have a way to change the current namespace from the explorer. We should expose this as a way to change the namespace from the command palette:

`Kubernetes: Set Namespace`

itowlson:
@kameshsampath Thanks for the PR. Regarding formatting, we haven't historically had a standard but are gradually working one in via tslint and editorconfig.  Your formatting changes look good, but having them in the same PR as functional changes muddies the water a bit - would you have the time to send them as a separate PR please?

itowlson:
Fixed via #270.

PR #270, opened 2018-06-30
Implementation of Kubernetes: set namespace command

kameshsampath:
This implements and resolves Issue #265 , wherein now the Namespace command is available via the Context menu in explorer as well as via command palette.

Signed-off-by: Kamesh Sampath <kamesh.sampath@hotmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
         "onCommand:extension.vsKubernetesPortForward",
         "onCommand:extension.vsKubernetesDeleteFile",
         "onCommand:extension.vsKubernetesAddFile",
+        "onCommand:extension.vsKubernetesUseNamespace",
         "onCommand:extension.helmTemplate",
         "onCommand:extension.helmTemplatePreview",
         "onCommand:extension.helmLint",
@@ -418,7 +419,7 @@
                 },
                 {
                     "command": "extension.vsKubernetesUseNamespace",
-                    "when": "view == extension.vsKubernetesExplorer"
+                    "when": ""
                 },
                 {
                     "command": "extension.vsKubernetesCopy",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,6 +15,7 @@ import * as dockerfileParse from 'dockerfile-parse';
 import * as tmp from 'tmp';
 import * as uuid from 'uuid';
 import * as clipboard from 'clipboardy';
+import {pullAll} from 'lodash';
 
 // Internal dependencies
 import { host } from './host';
@@ -903,14 +904,8 @@ function promptKindName(resourceKinds: kuberesources.ResourceKind[], description
     let placeHolder: string = 'Empty string to be prompted';
     let prompt: string = "What resource do you want to " + descriptionVerb + "?";
     if (opts) {
-        // see if the options has 'placeHolder' string, if so use it instead of default one
-        if (opts.placeHolder) {
-            placeHolder = opts.placeHolder;
-        }
-        // see if the options has custom 'prompt', if so use it instead of default one
-        if (opts.prompt) {
-            prompt = opts.prompt;
-        }
+        placeHolder = opts.placeHolder || placeHolder;
+        prompt = opts.prompt || prompt;
     }
     vscode.window.showInputBox({ prompt, placeHolder}).then((resource) => {
         if (resource === '') {
@@ -947,6 +942,9 @@ function quickPickKindNameFromKind(resourceKind: kuberesources.ResourceKind, opt
         if (names.length === 0) {
             vscode.window.showInformationMessage("No resources of type " + resourceKind.displayName + " in cluster");
             return;
+        }
+        if(opts){
+            names = pullAll(names,opts.filterNames) || names;
         }
         if (opts && opts.nameOptional) {
             names.push('(all)');
@@ -1754,33 +1752,27 @@ async function useNamespaceKubernetes(explorerNode: explorer.KubernetesObject) {
             refreshExplorer();
         }
     } else {
+        const currentNS = await kubectlUtils.currentNamespace(kubectl);
         promptKindName([kuberesources.allKinds.namespace], undefined,
             {
                 prompt: 'What namespace do you want to use?',
-                placeHolder: 'Enter the namespace to switch to or press enter to select from available list'
+                placeHolder: 'Enter the namespace to switch to or press enter to select from available list',
+                filterNames: [currentNS]
             },
-            (resource) => {
+            async (resource) => {
                 if (resource) {
-                    const currentNS = kubectlUtils.currentNamespace(kubectl);
-                    currentNS.then((ns) => {
-                        let toSwitchNamespace = resource;
-                        // resource will be of format <kind>/<name>, when picked up from the quickpick
-                        if (toSwitchNamespace.lastIndexOf('/') != -1) {
-                            toSwitchNamespace = toSwitchNamespace.substring(toSwitchNamespace.lastIndexOf('/') + 1);
+                    let toSwitchNamespace = resource;
+                    // resource will be of format <kind>/<name>, when picked up from the quickpick
+                    if (toSwitchNamespace.lastIndexOf('/') != -1) {
+                        toSwitchNamespace = toSwitchNamespace.substring(toSwitchNamespace.lastIndexOf('/') + 1);
+                    }
+                    // Switch if an only if the currentNS and toSwitchNamespace are different
+                    if (toSwitchNamespace && currentNS !== toSwitchNamespace) {
+                        const promiseSwitchNS = await kubectlUtils.switchNamespace(kubectl, toSwitchNamespace);
+                        if (promiseSwitchNS) {
+                            refreshExplorer();
                         }
-                        // Switch if an only if the currentNS and toSwitchNamespace are different
-                        if (toSwitchNamespace && ns !== toSwitchNamespace) {
-                            const promiseSwitchNS = kubectlUtils.switchNamespace(kubectl, toSwitchNamespace);
-                            promiseSwitchNS.then((data) => {
-                                if (data) {
-                                    refreshExplorer();
-                                }
-                            });
-                        }
-                    }).catch((err) => {
-                        host.showErrorMessage(err);
-                    });
-
+                    }
                 }
             });
     }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -26,6 +26,7 @@ import * as configmaps from './configMap';
 import * as configureFromCluster from './configurefromcluster';
 import * as createCluster from './createcluster';
 import * as kuberesources from './kuberesources';
+import { useNamespaceKubernetes } from './kubeNamespace';
 import * as docker from './docker';
 import { kubeChannel } from './kubeChannel';
 import * as kubeconfig from './kubeconfig';
@@ -65,7 +66,7 @@ import { Container, isPod, isKubernetesResource, KubernetesCollection, Pod, Kube
 let explainActive = false;
 let swaggerSpecPromise = null;
 
-const kubectl = kubectlCreate(host, fs, shell, installDependencies);
+export const kubectl = kubectlCreate(host, fs, shell, installDependencies);
 const draft = draftCreate(host, fs, shell, installDependencies);
 const configureFromClusterUI = configureFromCluster.uiProvider();
 const createClusterUI = createCluster.uiProvider();
@@ -900,7 +901,7 @@ export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[]
     }
 }
 
-function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
+export function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
     let placeHolder: string = 'Empty string to be prompted';
     let prompt: string = "What resource do you want to " + descriptionVerb + "?";
     if (opts) {
@@ -1324,7 +1325,7 @@ async function syncKubernetes(): Promise<void> {
     }
 }
 
-async function refreshExplorer() {
+export async function refreshExplorer() {
     await vscode.commands.executeCommand("extension.vsKubernetesRefreshExplorer");
 }
 
@@ -1746,38 +1747,6 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject) 
     }
     if (await kubectlUtils.deleteCluster(kubectl, contextObj)) {
         refreshExplorer();
-    }
-}
-
-async function useNamespaceKubernetes(explorerNode: explorer.KubernetesObject) {
-    if (explorerNode) {
-        if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
-            refreshExplorer();
-        }
-    } else {
-        const currentNS = await kubectlUtils.currentNamespace(kubectl);
-        promptKindName([kuberesources.allKinds.namespace], undefined,
-            {
-                prompt: 'What namespace do you want to use?',
-                placeHolder: 'Enter the namespace to switch to or press enter to select from available list',
-                filterNames: [currentNS]
-            },
-            async (resource) => {
-                if (resource) {
-                    let toSwitchNamespace = resource;
-                    // resource will be of format <kind>/<name>, when picked up from the quickpick
-                    if (toSwitchNamespace.lastIndexOf('/') != -1) {
-                        toSwitchNamespace = toSwitchNamespace.substring(toSwitchNamespace.lastIndexOf('/') + 1);
-                    }
-                    // Switch if an only if the currentNS and toSwitchNamespace are different
-                    if (toSwitchNamespace && currentNS !== toSwitchNamespace) {
-                        const promiseSwitchNS = await kubectlUtils.switchNamespace(kubectl, toSwitchNamespace);
-                        if (promiseSwitchNS) {
-                            refreshExplorer();
-                        }
-                    }
-                }
-            });
     }
 }
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,7 +15,7 @@ import * as dockerfileParse from 'dockerfile-parse';
 import * as tmp from 'tmp';
 import * as uuid from 'uuid';
 import * as clipboard from 'clipboardy';
-import {pullAll} from 'lodash';
+import { pullAll } from 'lodash';
 
 // Internal dependencies
 import { host } from './host';
@@ -938,14 +938,17 @@ function quickPickKindNameFromKind(resourceKind: kuberesources.ResourceKind, opt
             vscode.window.showErrorMessage(stderr);
             return;
         }
+
         let names = parseNamesFromKubectlLines(stdout);
         if (names.length === 0) {
             vscode.window.showInformationMessage("No resources of type " + resourceKind.displayName + " in cluster");
             return;
         }
-        if(opts){
-            names = pullAll(names,opts.filterNames) || names;
+
+        if (opts) {
+            names = pullAll(names, opts.filterNames) || names;
         }
+
         if (opts && opts.nameOptional) {
             names.push('(all)');
             vscode.window.showQuickPick(names).then((name) => {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -900,7 +900,19 @@ export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[]
 }
 
 function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
-    vscode.window.showInputBox({ prompt: "What resource do you want to " + descriptionVerb + "?", placeHolder: 'Empty string to be prompted' }).then((resource) => {
+    let placeHolder: string = 'Empty string to be prompted';
+    let prompt: string = "What resource do you want to " + descriptionVerb + "?";
+    if (opts) {
+        // see if the options has 'placeHolder' string, if so use it instead of default one
+        if (opts.placeHolder) {
+            placeHolder = opts.placeHolder;
+        }
+        // see if the options has custom 'prompt', if so use it instead of default one
+        if (opts.prompt) {
+            prompt = opts.prompt;
+        }
+    }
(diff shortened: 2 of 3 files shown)


src/extension.ts
```diff
@@ -42,8 +42,8 @@ import { HelmTemplateCompletionProvider } from './helm.completionProvider';
 import { Reporter } from './telemetry';
 import * as telemetry from './telemetry-helper';
 import * as extensionapi from './extension.api';
-import {dashboardKubernetes} from './components/kubectl/dashboard';
-import {portForwardKubernetes} from './components/kubectl/port-forward';
+import { dashboardKubernetes } from './components/kubectl/dashboard';
```
itowlson: Can you separate out the tidying commits into a separate PR please?  I realise we will always get the odd bit of unrelated tidying in PRs, but there seems to be quite a lot in this one, and it obscures the functional changes.  Thanks!

itowlson: To be clear, the tidying changes are on point and are very welcome!

src/extension.ts
```diff
@@ -894,8 +894,8 @@ export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[]
     }
 }
 
-function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
-    vscode.window.showInputBox({ prompt: "What resource do you want to " + descriptionVerb + "?", placeHolder: 'Empty string to be prompted' }).then((resource) => {
+function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler, placeHolder: string = 'Empty string to be prompted') {
```
itowlson: I think this placement is problematic.  The `handler` is likely to be an inline code block representing the core of the calling function; the placeHolder could therefore end up separated by 10-20 lines from the call to `promptKindName`, so it is very hard for the reader to connect this random string back to the role it plays.  An alternative might be to add it to the `opts` record?

src/extension.ts
```diff
@@ -1731,9 +1736,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+                                refreshExplorer();
+                            }
+                        });
+                    }
+                }).catch((err) => {
+                    host.showErrorMessage(err);
+                });
+
+            }
+        }, 'Type the namespace name or press enter to select from available');
```
itowlson: Per previous comment... this placeholder is 20 lines of multi-indent-level code away from the function it's used in... definitely needs to move earlier in the args list or into the opts record.

src/extension.ts
```diff
@@ -1731,9 +1736,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+    if (explorerNode) {
+        if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
+            refreshExplorer();
+        }
+    } else {
+        promptKindName([kuberesources.allKinds.namespace], 'get', undefined, (resource) => {
+            if (resource) {
+                const currentNS = kubectlUtils.currentNamespace(kubectl);
+                currentNS.then((ns) => {
+                    //resource will be of format <kind>/<name>, hence get the substring
```
itowlson: I'm not sure if we're consistent about this but I think we usually put a space between `//` and the comment body.

src/extension.ts
```diff
@@ -1731,9 +1736,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+                    let toSwitchNamespace = '' + resource;
```
itowlson: Is this to coerce `resource` to a string?  Surely `promptKindName` produces a `string | undefined` so no coercion is needed?  (We should add type annotations to the `promptKindName.handler` parameter to have it checked though.)

src/extension.ts
itowlson: It sounds like you're expecting the user, after invoking Set Namespace, to type `ns/foo` rather than just `foo`.  That seems counterintuitive - the user expects to enter the _namespace_ to switch to, and will not expect to have to enter the `ns/` prefix.

src/extension.ts
```diff
@@ -1731,9 +1736,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
-async function useNamespaceKubernetes(explorerNode: explorer.KubernetesObject) {
-    if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
-        refreshExplorer();
+async function useNamespaceKubernetes(explorerNode?: explorer.KubernetesObject) {
+    if (explorerNode) {
+        if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
+            refreshExplorer();
+        }
+    } else {
+        promptKindName([kuberesources.allKinds.namespace], 'get', undefined, (resource) => {
```
itowlson: This will result in a prompt of "What resource do you want to get?"; suggest we should aim for e.g. "What namespace do you want to use?" or "Enter the namespace to switch to."

src/extension.ts
```diff
@@ -1731,9 +1736,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+            if (resource) {
+                const currentNS = kubectlUtils.currentNamespace(kubectl);
+                currentNS.then((ns) => {
+                    //resource will be of format <kind>/<name>, hence get the substring
+                    let toSwitchNamespace = '' + resource;
+                    toSwitchNamespace = toSwitchNamespace.substring(toSwitchNamespace.lastIndexOf('/') + 1);
+                    //Switch if an only if the currentNS and toSwitchNamespace are different
+                    if (toSwitchNamespace && ns !== toSwitchNamespace) {
+                        const promiseSwitchNS = kubectlUtils.switchNamespace(kubectl, toSwitchNamespace);
```
itowlson: Can you use `await` here?

src/extension.ts
```diff
@@ -1731,9 +1736,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+async function useNamespaceKubernetes(explorerNode?: explorer.KubernetesObject) {
+    if (explorerNode) {
+        if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
+            refreshExplorer();
+        }
+    } else {
+        promptKindName([kuberesources.allKinds.namespace], 'get', undefined, (resource) => {
+            if (resource) {
+                const currentNS = kubectlUtils.currentNamespace(kubectl);
+                currentNS.then((ns) => {
```
itowlson: Can you use `await` here?

src/extension.ts
kameshsampath: will knock it off, as said in my previous PR its my vscode settings that has caused autoformat

src/extension.ts
kameshsampath: not sure I got you here will need more clarity on this fo fix neatly :)

src/extension.ts
kameshsampath: the arrow functions does not allow me to use await - mostly I guess I dont know :D . Can we use await in arrow function ?

src/extension.ts
kameshsampath: the arrow functions does not allow me to use await - mostly I guess I dont know :D . Can we use await in arrow function ?

src/extension.ts
kameshsampath: ok

kameshsampath: no i don't expect that way here with currentNamespace function does return just NS alone and when the explorer.id it returned with `<resource>/name`

kameshsampath:
@itowlson  - just force pushed things, to make it easier for review, I have tried to address most of your comments

> I think this placement is problematic. The handler is likely to be an inline code block representing the core of the calling function; the placeHolder could therefore end up separated by 10-20 lines from the call to promptKindName, so it is very hard for the reader to connect this random string back to the role it plays. An alternative might be to add it to the opts record?

But I do have incorporated it partially via `opts` and using that for getting the custom placeholder and prompt. I guess this is what you meant the place where its called and place where its used ??

src/extension.ts
```diff
@@ -1732,8 +1744,40 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
 
 async function useNamespaceKubernetes(explorerNode: explorer.KubernetesObject) {
-    if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
-        refreshExplorer();
+    if (explorerNode) {
+        if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
+            refreshExplorer();
+        }
+    } else {
+        promptKindName([kuberesources.allKinds.namespace], undefined,
```
bnookala: We should pre-fill the namespaces a user may switch to here. Entering the namespace name is clunky.

src/extension.ts
```diff
@@ -1732,8 +1744,40 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+            {
+                prompt: 'What namespace do you want to use?',
+                placeHolder: 'Enter the namespace to switch to or press enter to select from available list'
+            },
+            (resource) => {
+                if (resource) {
```
bnookala: I think this logic around checking to see if the current namespace === selected namespace is un-necessary. We should just switch to the selected namespace and refresh the explorer regardless of whether or not it's the current namespace. 

You could, additionally, remove the current namespace from the list of namespaces a user may switch to avoid the current namespace === selected namespace condition entirely!

src/extension.ts
```diff
@@ -1732,8 +1744,40 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+                    const currentNS = kubectlUtils.currentNamespace(kubectl);
+                    currentNS.then((ns) => {
```
bnookala: we try not to use `.then` or `.catch` in the code base, and prefer to use the more idiomatic `async/await` keywords.

src/extension.ts
```diff
@@ -895,7 +895,19 @@ export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[]
 }
 
 function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
-    vscode.window.showInputBox({ prompt: "What resource do you want to " + descriptionVerb + "?", placeHolder: 'Empty string to be prompted' }).then((resource) => {
+    let placeHolder: string = 'Empty string to be prompted';
+    let prompt: string = "What resource do you want to " + descriptionVerb + "?";
+    if (opts) {
```
itowlson: Thanks, this is the sort of thing I was thinking of.  Sorry I explained it badly.

src/extension.ts
```diff
@@ -895,7 +895,19 @@ export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[]
...
+        // see if the options has 'placeHolder' string, if so use it instead of default one
+        if (opts.placeHolder) {
```
itowlson: Style suggestion: consider `placeholder = opts.placeHolder || placeholder;` (and similarly for prompt).

src/extension.ts
```diff
@@ -895,7 +895,19 @@ export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[]
 }
 
 function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
-    vscode.window.showInputBox({ prompt: "What resource do you want to " + descriptionVerb + "?", placeHolder: 'Empty string to be prompted' }).then((resource) => {
+    let placeHolder: string = 'Empty string to be prompted';
+    let prompt: string = "What resource do you want to " + descriptionVerb + "?";
+    if (opts) {
+        // see if the options has 'placeHolder' string, if so use it instead of default one
```
itowlson: I would drop this comment (and similarly for prompt) - the code is clear and I don't think it needs any additional explanation through the comment.

src/extension.ts
```diff
@@ -1732,8 +1744,40 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+                    const currentNS = kubectlUtils.currentNamespace(kubectl);
+                    currentNS.then((ns) => {
+                        let toSwitchNamespace = resource;
+                        // resource will be of format <kind>/<name>, when picked up from the quickpick
+                        if (toSwitchNamespace.lastIndexOf('/') != -1) {
+                            toSwitchNamespace = toSwitchNamespace.substring(toSwitchNamespace.lastIndexOf('/') + 1);
+                        }
+                        // Switch if an only if the currentNS and toSwitchNamespace are different
+                        if (toSwitchNamespace && ns !== toSwitchNamespace) {
+                            const promiseSwitchNS = kubectlUtils.switchNamespace(kubectl, toSwitchNamespace);
```
itowlson: Prefer using `await` here, e.g.:

```
try {
  if (await kubectlUtils.switchNamespace(...)) {
    refreshExplorer();
  }
} catch (err) {
  // ...
}
```

src/extension.ts
itowlson: You should just need to make the arrow function async, e.g.

```
promptKindName(…, async (resource) => {
  const ns = await kubectlUtils.currentNamespace(…);  // should work now
));

File: package.json [hunk 16]
```diff
@@ -386,7 +387,7 @@
                 },
                 {
                     "command": "extension.vsKubernetesUseNamespace",
-                    "when": "view == extension.vsKubernetesExplorer"
```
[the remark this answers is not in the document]
kameshsampath: yes this allows the command to be shown in Command Palette

src/extension.ts
kameshsampath: am working on to make the type-search-list (filter) pattern . WDYT?

src/extension.ts
kameshsampath: I wish to avoid unnecessary command when current == switch, the time and delay is trivial in local minikube or local kubernetes cluster, but if you are connected to remote cluster this will cause unnecessary delay. I wish to keep to avoid extra command- TBH thats transparent to the user.

src/extension.ts
kameshsampath: no worries! am glad i understood.

src/extension.ts
kameshsampath: fantastic will do.

src/extension.ts
kameshsampath: sorry, i added for my clarity. forgot to remove

src/extension.ts
kameshsampath: learnt how to use async for arrow functions, added it :)

src/extension.ts
kameshsampath: I was having that first but then some times it takes few seconds to load the namespaces list, hence thought to have typing allowed

src/extension.ts
```diff
@@ -15,6 +15,7 @@ import * as dockerfileParse from 'dockerfile-parse';
 import * as tmp from 'tmp';
 import * as uuid from 'uuid';
 import * as clipboard from 'clipboardy';
+import {pullAll} from 'lodash';
```
bnookala: `{ pullAll }`

src/extension.ts
```diff
@@ -900,7 +901,13 @@ export function findKindNameOrPrompt(resourceKinds: kuberesources.ResourceKind[]
 }
 
 function promptKindName(resourceKinds: kuberesources.ResourceKind[], descriptionVerb, opts, handler) {
-    vscode.window.showInputBox({ prompt: "What resource do you want to " + descriptionVerb + "?", placeHolder: 'Empty string to be prompted' }).then((resource) => {
+    let placeHolder: string = 'Empty string to be prompted';
+    let prompt: string = "What resource do you want to " + descriptionVerb + "?";
+    if (opts) {
+        placeHolder = opts.placeHolder || placeHolder;
+        prompt = opts.prompt || prompt;
+    }
```
bnookala: newline below

src/extension.ts
```diff
@@ -936,6 +943,9 @@ function quickPickKindNameFromKind(resourceKind: kuberesources.ResourceKind, opt
             vscode.window.showInformationMessage("No resources of type " + resourceKind.displayName + " in cluster");
             return;
         }
```
bnookala: newline below

src/extension.ts
```diff
@@ -936,6 +943,9 @@ function quickPickKindNameFromKind(resourceKind: kuberesources.ResourceKind, opt
...
+        if(opts){
+            names = pullAll(names,opts.filterNames) || names;
+        }
```
bnookala: newline below

src/extension.ts
```diff
@@ -936,6 +943,9 @@ function quickPickKindNameFromKind(resourceKind: kuberesources.ResourceKind, opt
             vscode.window.showInformationMessage("No resources of type " + resourceKind.displayName + " in cluster");
             return;
         }
+        if(opts){
```
bnookala: `if (opts) {`

src/extension.ts
```diff
@@ -936,6 +943,9 @@ function quickPickKindNameFromKind(resourceKind: kuberesources.ResourceKind, opt
...
+            names = pullAll(names,opts.filterNames) || names;
```
bnookala: space between , and arguments

kameshsampath:
all these formatting changes are side effects of me turning off the auto format on save ;) ..  if we can get these rules into tslint and/or .editorconfig that saves a lot of energy here

src/extension.ts
```diff
@@ -1737,8 +1747,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
 }
 
 async function useNamespaceKubernetes(explorerNode: explorer.KubernetesObject) {
```
bnookala: in the spirit of breaking down extension.ts, move this into it's own file under `src/components/kubectl/namespace.ts`

src/extension.ts
```diff
@@ -1737,8 +1747,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
-    if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
-        refreshExplorer();
+    if (explorerNode) {
+        if (await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
```
bnookala: you can combine these conditions since they're both expected to be truthy. Additionally, an early-return pattern is preferable here, since the majority of work is done in the "else" block:

```
if (explorerNode && await kubectlUtils.switchNamespace(kubectl, explorerNode.id)) {
    refreshExplorer();
    return;
}

const currentNs = await kubectlUtils.currentNamespace(kubectl);
...
```

src/extension.ts
```diff
@@ -1737,8 +1747,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+    } else {
+        const currentNS = await kubectlUtils.currentNamespace(kubectl);
+        promptKindName([kuberesources.allKinds.namespace], undefined,
+            {
+                prompt: 'What namespace do you want to use?',
+                placeHolder: 'Enter the namespace to switch to or press enter to select from available list',
+                filterNames: [currentNS]
+            },
+            async (resource) => {
+                if (resource) {
```
bnookala: use the early return pattern as described above to return early if no resource was provided.

src/extension.ts
```diff
@@ -1737,8 +1747,34 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+                    let toSwitchNamespace = resource;
+                    // resource will be of format <kind>/<name>, when picked up from the quickpick
+                    if (toSwitchNamespace.lastIndexOf('/') != -1) {
```
bnookala: use the stricter equality check: !==

bnookala:
@kameshsampath Do you have the tslint extension? We do actually have both .editorconfig and tsconfig: https://github.com/azure/vscode-kubernetes-tools/blob/master/tslint.json && https://github.com/azure/vscode-kubernetes-tools/blob/master/.editorconfig

src/extension.ts
kameshsampath: i guess thats fall under the else part and should return automatically there. We can skip early return here.

src/extension.ts
kameshsampath: yes done, and there are some access changes to extension.ts to allow access of them from other refactored files. Ideally I can access `kubectl` now in events file as well rather than having it as method  param - but that change will introduce a conflict will leave it for now as it is .WDYT?

bnookala: happy to leave it as is, and we can come back to it later.

kameshsampath: yes. thats what i did. as of now its own file and in future we can start working on the refactoring of each logical resource as separate PR for easy fixing them

src/extension.ts
bnookala: @kameshsampath I'm not sure I understand your reasoning.

kameshsampath: i meant to say for the early return in this case might not be needed as there no more lines post that which needs to be evaluated, since if the `resource` is not there then it skip the block and return automatically.

src/extension.ts
kameshsampath: for constants we can export it rather doing DI as that will have shorter method param list and easier to read and maintain.

src/extension.ts
bnookala: ah got it. early returns are more style than function. i prefer to use them since they don't use as much (unpleasant, in my opinion) indentation, providing a more clear picture to the reader as to what's going on in the truthy condition in functions which only perform one thing or the other. 

the following are exactly the same functionally:

```
function myFunction() {
    if (someTruthyValue) {
         something();
    } else {
         somethingElse();
    }
}
```

vs.
 
```
function myFunction {
    if (!someTruthyValue) {
        somethingElse();
        return;
    }

    something();
}
```

the truthy branch has no impact on the non-truthy branch.

src/extension.ts
bnookala: @kameshsampath let's stick with the DI for now, since that's how many other components expect it to be used:

https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/extension.ts#L151
https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/extension.ts#L155
https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/extension.ts#L157
https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/extension.ts#L158

kameshsampath: Agreed! pushed new changes with DI pattern

kameshsampath:
thanks @itowlson  @bnookala  for reviewing and accepting my first contribution :)

PR #271, opened 2018-07-02
Load correctly when tree item is not in active namespace

### LINKED ISSUE #267: Explorer: Load pods not in non-active namespaces
As @itowlson had noted, selecting a Pod in a non-default namespace from the Node tree in the explorer won't load the resource in VSCode. This is due to the way the `Load` command works (by using the current active namespace). This should be fixed such that the `Load` command is aware of the namespace resources are in, so it can load them.

itowlson:
Fixes #267.

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -39,7 +39,7 @@ export interface KubernetesObject {
 export interface ResourceNode {
     readonly id: string;
     readonly resourceId: string;
-    namespace(): string | null;
+    namespace: string | null;
 }
 
 export class KubernetesExplorer implements vscode.TreeDataProvider<KubernetesObject> {
@@ -189,7 +189,7 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
         this.resourceId = kind.abbreviation + '/' + id;
     }
 
-    namespace(): string | null {
+    get namespace(): string | null {
         return (this.metadata && this.metadata.namespace) ? this.metadata.namespace : null;
     }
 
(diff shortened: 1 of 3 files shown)


src/explorer.ts
```diff
@@ -188,6 +189,10 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
         this.resourceId = kind.abbreviation + '/' + id;
     }
 
+    namespace(): string | null {
```
bnookala: can we turn this into a getter? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

src/explorer.ts
```diff
@@ -188,6 +189,10 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+        return (this.metadata && this.metadata.namespace) ? this.metadata.namespace : null;
```
bnookala: is it even possible for an object to not have a namespace?

src/kuberesources.virtualfs.ts
```diff
@@ -51,7 +51,9 @@ export class KubernetesResourceVirtualFileSystemProvider implements FileSystemPr
     async loadResource(uri: Uri): Promise<string> {
         const outputFormat = workspace.getConfiguration('vs-kubernetes')['vs-kubernetes.outputFormat'];
         const value = querystring.parse(uri.query).value;
-        const sr = await this.kubectl.invokeAsyncWithProgress(`-o ${outputFormat} get ${value}`, `Loading ${value}...`);
+        const ns = querystring.parse(uri.query).ns;
```
bnookala: querystring.parse returns an object, so you should just be able to do this once and read from the object.

src/explorer.ts
itowlson: It _is_ possible per the spec, though shouldn't happen on a get... but in any case `metadata` is optional and untyped so I don't want to presume anything.  I want to get it clearer and more consistent what the tree view nodes contain, and that might obviate this check, but for the time being I feel safer with it in!

PR #274, opened 2018-07-03
Adding stricter tslint rules

bnookala:
Adding stricter tslint rules to make it easier to bring up code smells in dev, rather than in code review.


src/components/clusterprovider/azure/azure.ts
```diff
@@ -154,7 +154,7 @@ async function listLocations(context: Context): Promise<Errorable<Locations>> {
     const sr = await context.shell.exec(`az account list-locations --query ${query} -ojson`);
 
     return fromShellJson<Locations>(sr, (response) => {
-        let locations: any = {};
+        const locations: any = {};
```
itowlson: This is a scenario where I find the formal syntax confusing.  Which the `locations` _reference_ is constant, the object itself is going to be mutated.  Using `const` makes me think the object is constant, which it isn't.  I know from a language spec point of view the usage here is correct but I personally prefer `let` in these cases because it warns me to look for how the object changes over the course of its scope.  #lostbattles

src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -416,7 +416,7 @@ ${styles()}
 // Utility helpers
 
 function formatCluster(cluster: any): string {
-    return cluster.resourceGroup + '/' + cluster.name;
+    return `${cluster.resourceGroup}'/'${cluster.name}`;
```
itowlson: Are you sure this is right?  It looks like it will introduce spurious characters e.g. `myrg/mycluster` -> `myrg'/'mycluster` - can you check it still works correctly?

src/debug/debugProvider.ts
```diff
@@ -1,7 +1,5 @@
-import * as vscode from "vscode";
```
itowlson: Remove leading blank line?

src/debug/debugSession.ts
```diff
@@ -262,7 +262,8 @@ export class DebugSession implements IDebugSession {
     private async setupPortForward(podName: string, debugPort: number, appPort?: number): Promise<ProxyResult> {
         kubeChannel.showOutput(`Setting up port forwarding on pod ${podName}...`, "Set up port forwarding");
         const proxyResult = await this.createPortForward(this.kubectl, podName, debugPort, appPort);
-        kubeChannel.showOutput(`Created port-forward ${proxyResult.proxyDebugPort}:${debugPort} ${appPort ? proxyResult.proxyAppPort + ":" + appPort : ""}`);
+        const portStr = appPort ? `${proxyResult.proxyAppPort}:${appPort}` : "";
```
itowlson: `appPortStr` - there are too many ports in play for the unadorned name to be clear.

src/docker/dockerfileParser.ts
```diff
@@ -2,12 +2,11 @@ import * as fs from 'fs';
 import { CommandEntry, parse } from 'docker-file-parser';
 
 import { IDockerfile, IDockerParser } from "./parser";
-import { shell, ShellResult } from "../shell";
 
 class RawDockerfile {
     private readonly commandEntries: CommandEntry[];
 
-    constructor(private readonly dockerfilePath: string) {
+    constructor(readonly dockerfilePath: string) {
```
itowlson: This appears to make `this.dockerfilePath` public.  Is this necessary?

src/docker/dockerfileParser.ts
```diff
@@ -48,7 +47,7 @@ class RawDockerfile {
 class Dockerfile implements IDockerfile {
     private readonly dockerfile: RawDockerfile;
 
-    constructor(private readonly dockerfilePath: string) {
+    constructor(readonly dockerfilePath: string) {
```
itowlson: Does this need to be public?

src/extension.ts
```diff
@@ -130,7 +129,7 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
         registerCommand('extension.vsKubernetesGet', getKubernetes),
         registerCommand('extension.vsKubernetesRun', runKubernetes),
         registerCommand('extension.vsKubernetesShowLogs', logsKubernetes),
-        registerCommand('extension.vsKubernetesFollowLogs', (explorerNode: explorer.ResourceNode) => {logsKubernetes(explorerNode, true);}),
+        registerCommand('extension.vsKubernetesFollowLogs', (explorerNode: explorer.ResourceNode) => {logsKubernetes(explorerNode, true); }),
```
itowlson: Space before closing brace but not after opening brace looks asymmetrical.  If our lint rules allow this then is there something missing?

src/extension.ts
```diff
@@ -486,16 +483,16 @@ function initStatusBar() {
 //
 // @parameter command string The command to run
 function maybeRunKubernetesCommandForActiveWindow(command, progressMessage) {
-    let text, proc;
+    let text;
```
itowlson: Suggest we supply a type declaration while we're in here.

src/extension.ts
```diff
@@ -1087,7 +1084,7 @@ async function logsKubernetes(explorerNode?: explorer.ResourceNode, follow?: boo
             getLogsForContainer(podSummary.name, podSummary.namespace, container.name, follow);
         }
     } else {
-        findPod((pod: PodSummary) => {getLogsForPod(pod, follow);});
+        findPod((pod: PodSummary) => {getLogsForPod(pod, follow); });
```
itowlson: Unbalanced spacing again

src/extension.ts
```diff
@@ -1650,8 +1635,8 @@ const waitForRunningPod = (name, callback) => {
 };
 
 function exists(kind, name, handler) {
-    //eslint-disable-next-line no-unused-vars
-    kubectl.invoke('get ' + kind + ' ' + name, (result) => {
+    // eslint-disable-next-line no-unused-vars
```
itowlson: Can probably get rid of this eslint directive?

test/draft.test.ts
```diff
@@ -38,7 +33,7 @@ suite("draft tests", () => {
         suite("If draft is not on the path", () => {
 
             test("...and configuration is not present, then checkPresent reports an error", async () => {
-                let errors: string[] = [];
+                const errors: string[] = [];
```
itowlson: Ugh, this is particularly misleading because the `errors` collection gets things pushed onto it _via being passed as an argument_, so the mutation is very non-obvious.  (I guess the fix is to make the mutation more obvious somehow...)

src/components/clusterprovider/azure/azureclusterprovider.ts
bnookala: oh derp. good catch.

src/debug/debugProvider.ts
bnookala: 👍

src/debug/debugSession.ts
bnookala: 👍

src/extension.ts
bnookala: 👍

src/extension.ts
bnookala: 👍

src/extension.ts
bnookala: 👍

test/draft.test.ts
bnookala: "fixed" all of these in so far as i left them as `let` with a disable condition for the tslint rule.

src/components/clusterprovider/azure/azure.ts
bnookala: sounds good to me. fixed these!

bnookala:
@itowlson fixed the stuff you pointed out!

bnookala:
❤️

Issue #272, opened 2018-07-02
Pod status icon in tree view always shows stopped


itowlson:
I've noticed that in master the pod status always seems to be showing red for me.  If I do a `kubectl get pods` then they show as running but the icon in the tree view is red.  @shreyu86 do you have any idea what might be going on?  Wondering if something when awry when changing the column-based status parsing to JSON-based perhaps?  If you don't have time to look into it then let me know and I'll dig deeper.

shreyu86:
I will check this first thing if it is possible can you share the pod JSON output?

bnookala:
@itowlson to my knowledge, the parsing for the status is still using columns.

shreyu86:
@bnookala +1, if the columns is causing any issues might have some time this week to checkout the JSON route.

itowlson:
@shreyu86 Yes, that's right.  For some reason I convinced myself that the `Workloads > Pods` icons were also red, but trying now I can see I was wrong about that.  Sorry for the misleading bug report and thanks for diagnosing it correctly.  If you can fix the `Nodes` tree that would be great - if #271 would help as a foundation for that then we will hopefully merge that soon.  Thank you!

shreyu86:
@itowlson thanks for confirming. Nodes uses [slightly different logic](https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/explorer.ts#L239) than `Workload-> Pods` to list pods and filter them. Thinking about a way in which both could fetch info regarding the running state of the pod. Let me know if you have any pointers on this.

bnookala:
Fixed by #277 and #278.

PR #277, opened 2018-07-04
Fix for Pod status in Node Explorer View #272

shreyu86:
@itowlson this fixes the issue with the pod status in node explorer view. Wanted to open an PR for feedback to see if we should move to `JSON` based parsing of status as compared to column based parsing. Also wanted to get some clarification on use of [`getPods`](https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/explorer.ts#L258) vs [`kubectl.asLines`](https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/explorer.ts#L181) to get pod information as to which is the preferred method. Happy to refactor `getPods` to reflect pod status.

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -20,26 +20,11 @@ export function createKubernetesResource(kind: kuberesources.ResourceKind, id: s
 }
 
 function getIconForPodStatus(status: string): vscode.Uri {
-    if (status === "Running" || status === "Completed") {
+    if (status === "running" || status === "completed") {
         return vscode.Uri.file(path.join(__dirname, "../../images/runningPod.svg"));
     } else {
         return vscode.Uri.file(path.join(__dirname, "../../images/errorPod.svg"));
     }
-}
-
-function parsePodLine(podInfo: string[]): any[] {
-    const headers = podInfo.shift();
-    const parsedHeaders = headers.toLowerCase().replace(/\s+/g, '|').split('|');
-    const pods = [];
-    podInfo.map((line) => {
-        const podObject = {};
-        const bits = line.replace(/\s+/g, '|').split('|');
-        bits.map((columnValue, index) => {
-            podObject[parsedHeaders[index]]= columnValue;
-        });
-        pods.push(podObject);
-    });
-    return pods;
 }
 
 export interface KubernetesObject {
@@ -193,22 +178,18 @@ class KubernetesResourceFolder extends KubernetesFolder {
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
-        const childrenLines = await kubectl.asLines(`get ${this.kind.abbreviation}`);
-        if (failed(childrenLines)) {
-            host.showErrorMessage(childrenLines.error[0]);
+        const childrenObjects = await kubectl.fromLines(`get ${this.kind.abbreviation}`);
+        if (failed(childrenObjects)) {
+            host.showErrorMessage(childrenObjects.error[0]);
             return [new DummyObject("Error")];
         }
         if (this.kind.abbreviation === "pod") {
-            const parsedPods = parsePodLine(childrenLines.result);
-            return parsedPods.map((line) => {
-                return new KubernetesResource(this.kind, line.name, { status: line.status });
+            return childrenObjects.result.map((pod) => {
+                return new KubernetesResource(this.kind, pod.name, { status: pod.status.toLowerCase() });
             });
         }
-        childrenLines.result.shift();
-        // ignore headers for other resources
-        return childrenLines.result.map((line) => {
-            const bits = line.split(' ');
-            return new KubernetesResource(this.kind, bits[0]);
+        return childrenObjects.result.map((childrenObject) => {
+            return new KubernetesResource(this.kind, childrenObject.name);
         });
     }
 }
@@ -274,14 +255,13 @@ class KubernetesNodeResource extends KubernetesResource {
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
-        const pods = await kubectl.asLines(`get po --all-namespaces --field-selector spec.nodeName=${this.id}`);
+        const pods = await kubectl.fromLines(`get po --all-namespaces --field-selector spec.nodeName=${this.id}`);
         if (failed(pods)) {
             host.showErrorMessage(pods.error[0]);
             return [new DummyObject("Error")];
         }
-        const parsedPods = parsePodLine(pods.result);
-        return parsedPods.map((line) => {
-            return new KubernetesResource(kuberesources.allKinds.pod, line.name, { status: line.status, name: line.name, namespace: line.namespace, node: this.resourceId });
+        return pods.result.map((pod) => {
+            return new KubernetesResource(kuberesources.allKinds.pod, pod.name, { status: pod.status.toLowerCase(), name: pod.name, namespace: pod.namespace, node: this.resourceId });
         });
     }
 }
@@ -358,14 +338,13 @@ class KubernetesSelectorResource extends KubernetesResource {
             selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
         }
         const selectorString = selectors.join(",");
-        const pods = await kubectl.asLines(`get po -l ${selectorString}`);
+        const pods = await kubectl.fromLines(`get po -l ${selectorString}`);
         if (failed(pods)) {
             host.showErrorMessage(pods.error[0]);
             return [new DummyObject("Error")];
         }
-        const parsedPods = parsePodLine(pods.result);
-        return parsedPods.map((line) => {
-            return new KubernetesResource(kuberesources.allKinds.pod, line.name, { status: line.status });
+        return pods.result.map((pod) => {
+            return new KubernetesResource(kuberesources.allKinds.pod, pod.name, { status: pod.status.toLowerCase() });
         });
     }
 }
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -20,11 +20,26 @@ export function createKubernetesResource(kind: kuberesources.ResourceKind, id: s
 }
 
 function getIconForPodStatus(status: string): vscode.Uri {
-    if (status === "running" || status === "completed") {
+    if (status === "Running" || status === "Completed") {
         return vscode.Uri.file(path.join(__dirname, "../../images/runningPod.svg"));
     } else {
         return vscode.Uri.file(path.join(__dirname, "../../images/errorPod.svg"));
     }
+}
+
+function parsePodLine(podInfo: string[]): any[] {
+    const headers = podInfo.shift();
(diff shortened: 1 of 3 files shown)


itowlson:
Good questions.  The nice thing about `getPods` is it returns typed data, so we get some checking of code that uses it.  Though we would need to add the status to the returned `PodInfo` type, as it's not there at the moment; and we'd need to confirm that we can map the JSON `status.phase` to the `kubectl` RUNNING/COMPLETED statuses.

But `KubernetesResourceFolder` already uses `asLines`, and this is probably more efficient than pulling back the full JSON document.  So I'm pretty comfortable with changing `KubernetesNodeResource` to also use `asLines`.  However, if we do this, it would be good to centralise the pod line parsing logic - with this PR we semi-duplicate it (because of the different indexes?), and that could lead to maintenance woes in future...

cc @brendandburns who wrote the original code and may have input

shreyu86:
@itowlson Thanks for the feedback, +1 on centralizing the pod line parsing logic. The reason that this PR duplicates it because of the extra `namespace` column being returned because of the `--all-namespaces` flag.

itowlson:
Yeah, I tried running the `kubectl` command (because I was surprised by the different indexes) and saw that the different indexes were needed.  I'm not sure what the best solution is but here's a possible idea:

* Define a `PodLineLayout` interface with members `NamespaceIndex`, `NameIndex`, etc.
* Define two instances of this interface - `NamespacedPodLineLayout` with the normal indexes, and `AllNamespacesPodLineLayout` with the +1 indexes.
* Define a `parsePodLine` method that takes a line and an instance of `PodLineLayout`.
* Call `parsePodLine` from the two different places with the different layout instances.

If you want to get fancy you could even infer the layouts from the header line... (which `asLines` currently discards but we could change that)

An _even fancier_ alternative would be to create a new method that uses the header line to convert each line into an object, so instead of returning `string[]` it would return `any[]` and the objects would have fields `NAMESPACE`, `NAME` etc. and you could index into them using string literals e.g. `pod['NAMESPACE]`, `pod['NAME']`.  That would break the dependence on positional parsing altogether...

Or there are other ways.  Just some ideas to consider.

src/explorer.ts
```diff
@@ -20,13 +20,28 @@ export function createKubernetesResource(kind: kuberesources.ResourceKind, id: s
 }
 
 function getIconForPodStatus(status: string): vscode.Uri {
-    if (status === "running" || status === "completed") {
+    if (status === "Running" || status === "Completed") {
```
itowlson: I'd suggest using a case-insensitive check so we don't have problems with possible different casings when the function is used from new places.

shreyu86:
@itowlson made following changes: 
- modified `asLines` to retain header
- converted each line into an object using the header column in `parsePodLine` breaking dependence on positional parsing altogether.

src/explorer.ts
```diff
@@ -20,13 +20,28 @@ export function createKubernetesResource(kind: kuberesources.ResourceKind, id: s
...
         return vscode.Uri.file(path.join(__dirname, "../../images/runningPod.svg"));
     } else {
         return vscode.Uri.file(path.join(__dirname, "../../images/errorPod.svg"));
     }
 }
 
+function parsePodLine(podInfo: string[]): any[] {
```
itowlson: I think we can refine the return type a little, to `[key: string]: string` (which guarantees that it's a map from strings to strings, rather than involving arbitrary value data).

itowlson: This parses multiple lines, including the header, so should be called `parsePodLines` (plural).

src/explorer.ts
```diff
@@ -20,13 +20,28 @@ export function createKubernetesResource(kind: kuberesources.ResourceKind, id: s
...
+    const headers = podInfo.shift();
+    const parsedHeaders = headers.toLowerCase().replace(/\s+/g, '|').split('|');
+    const pods = [];
+    podInfo.map((line) => {
```
itowlson: This construct maps a `void` function over the string array, and uses the side effect of the `void` function pushing an object onto an object array.  Prefer `const pods = podInfo.map((line) => { ...; return podObject; });` - this says to map an object-returning function over the string array to produce an array of objects directly.

src/explorer.ts
```diff
@@ -20,13 +20,28 @@ export function createKubernetesResource(kind: kuberesources.ResourceKind, id: s
...
+        const podObject = {};
+        const bits = line.replace(/\s+/g, '|').split('|');
+        bits.map((columnValue, index) => {
```
itowlson: Can this be done with `forEach`?  `map` implies the function is performing a transformation and you are interested in the return values of the transformation, whereas this is being done purely for its side-effects (to populate the `podObject`).

src/explorer.ts
```diff
@@ -20,13 +20,28 @@ export function createKubernetesResource(kind: kuberesources.ResourceKind, id: s
...
+            podObject[parsedHeaders[index]]= columnValue;
```
itowlson: Space before `=`

src/explorer.ts
```diff
@@ -183,11 +198,13 @@ class KubernetesResourceFolder extends KubernetesFolder {
             host.showErrorMessage(childrenLines.error[0]);
             return [new DummyObject("Error")];
         }
+        if (this.kind.abbreviation === "pod") {
+            const parsedPods = parsePodLine(childrenLines.result);
+            return parsedPods.map((line) => {
```
itowlson: The range variable is now a pod, not a line.

src/explorer.ts
```diff
@@ -183,11 +198,13 @@ class KubernetesResourceFolder extends KubernetesFolder {
...
+                return new KubernetesResource(this.kind, line.name, { status: line.status });
```
itowlson: Interesting, I would have thought the header names were capitalised - does that not matter?

src/kubectl.ts
```diff
@@ -219,7 +219,7 @@ async function asLines(context: Context, command: string): Promise<Errorable<str
     const shellResult = await invokeAsync(context, command);
     if (shellResult.code === 0) {
         let lines = shellResult.stdout.split('\n');
-        lines.shift();
+        // lines.shift();
```
itowlson: Won't this break other consumers of `asLines` who expect the headers to have been stripped off?

src/kubectl.ts
```diff
@@ -219,7 +219,6 @@ async function asLines(context: Context, command: string): Promise<Errorable<str
     const shellResult = await invokeAsync(context, command);
     if (shellResult.code === 0) {
         let lines = shellResult.stdout.split('\n');
-        lines.shift();
```
itowlson: I think removing this line will break other consumers of `asLines`, which expect the header line to have been removed.

src/explorer.ts
itowlson: Looking at this function, I don't think it has anything to do with pods specifically - you have designed it to be completely generic.  So I think we should consider moving it into `kubectl` or `kubectlUtils` as a generic "get me a parsed representation of a tabular resource list" function.  In fact it would even work with Helm...

src/kubectl.ts
itowlson: I suggest we should either:

* Have two functions, `asLines` and `asLinesWithHeader` (both returning string arrays)
* Have two functions, `asLines` (unparsed) and `fromLines` (parsed), the former returning a string array and the latter an object array
* Have a `keepHeaderLine` argument (requires changing all calls to specify whether they want to keep the header line)
* Return a `{ headerLine: string; resourceLines: string[] }` instead of a `string` (requires changing all calls to discard the header line, which I know could be done with an `unshift` but this strategy gives us compiler checking)

I quite like the second option as it minimises changes elsewhere in the codebase, and as the parsed `fromLines` is probably going to be handy going forward.  But not quite sure, and naming is going to be tricky...

src/explorer.ts
shreyu86: Sounds good. Will change this.

src/explorer.ts
shreyu86: perfect, will do that.

shreyu86: sounds good.

src/explorer.ts
shreyu86: perfect, will change this.

src/explorer.ts
shreyu86: sounds good.

src/explorer.ts
shreyu86: fixing this.

src/explorer.ts
shreyu86: will make this change.

src/explorer.ts
shreyu86: I changed the header check to be case sensitive, will change this as the check for status has to be case insensitive.

src/kubectl.ts
shreyu86: Sounds good will go for second option.

src/explorer.ts
shreyu86: Cool, will move it to `kubectlUtils` if that is ok.

shreyu86:
@itowlson will address your comments soon. I have made one last change to make sure that `KubernetesSelectorResource`s listing of pods can take advantage of the changes. I am relatively new to TypeScript and appreciate your feedback, thanks again.

shreyu86:
@itowlson refactored code according to your feedback:

- Renamed `parsePodLine` as `parseLineOutput` and moved under `kubectlUtils` and replaced return `any[]` with `{ [key: string]: string }[]` 
- Added `fromLines` in `kubectl` to take advantage of `parseLineOutput` and swapped `asLines` (for pod status) in `explorer`

itowlson:
Thanks @shreyu86 for the quick fix!

src/kubectl.ts
```diff
@@ -227,6 +232,17 @@ async function asLines(context: Context, command: string): Promise<Errorable<str
     return { succeeded: false, error: [ shellResult.stderr ] };
 }
 
+async function fromLines(context: Context, command: string): Promise<Errorable<{ [key: string]: string }[]>> {
+    const shellResult = await invokeAsync(context, command);
+    if (shellResult.code === 0) {
+        let lines = shellResult.stdout.split('\n');
+        lines = lines.filter((l) => l.length > 0);
```
shreyu86: @itowlson I missed checking for empty response here, which will break `parseLineOutput`.
Thinking about following solution or open to other solutions: 
```typescript
let parsedOutput = [];
        if (lines.length > 0) {
            parsedOutput = parseLineOutput(lines);
        }
        return { succeeded: true, result: parsedOutput};
```

itowlson: Good catch.  You're right, if there are no resources then there is no header line.  I'd actually consider putting the fix into `kubectlUtils.parseLineOutput` -  you do `const headers = lineOutput.shift()` to get the header line, and checking `if (!headers)` should solve the problem nicely (and play better with the typechecker once Project Strict Nulls gets that far) - I am pretty sure `shift` returns `undefined` if there are no array elements...?

shreyu86: sounds good, testing this approach.

PR #279, opened 2018-07-06
Logs invoked in new terminal

### LINKED ISSUE #273: Kubernetes Follow Logs command should invoke in new terminal
The shared terminal might be running another task. Additionally, since it's possible for the command to be run a long running task in the shared terminal, which could block future tasks. We should change the follow logs command to use a new terminal.

bnookala:
Fixes #273.

Logs will be invoked in a new terminal when invoked from the command palette with the 'Follow Logs' command.

Additionally, fixes a TODO I had noticed in which Terminal invocation would only work for Pods in the current namespace.

### DIFF
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -1,0 +1,164 @@
+import * as vscode from 'vscode';
+import { Kubectl } from '../../kubectl';
+import { selectContainerForPod, PodSummary, quickPickKindName } from '../../extension';
+import { isPod } from '../../kuberesources.objectmodel';
+import * as kuberesources from '../../kuberesources';
+import { ResourceNode } from '../../explorer';
+import * as yaml from 'js-yaml';
+import * as kubectlUtils from '../../kubectlUtils';
+import { host } from '../../host';
+
+export enum LogsDisplayMode {
+    Show,
+    Follow
+}
+
+/**
+ * Fetches logs for a Pod. Handles use cases for fetching pods
+ * from an open document, or from the current namespace.
+ */
+export async function logsKubernetes(
+    kubectl: Kubectl,
+    explorerNode: ResourceNode | undefined,
+    displayMode: LogsDisplayMode
+) {
+    if (explorerNode) {
+        return await getLogsForExplorerPod(kubectl, explorerNode, displayMode);
+    }
+
+    return logsForPod(kubectl, displayMode);
+}
+
+/**
+ * Fetch logs from a Pod, when selected from the Explorer.
+ */
+async function getLogsForExplorerPod(
+    kubectl: Kubectl,
+    explorerNode: ResourceNode,
+    displayMode: LogsDisplayMode
+) {
+    const namespace = explorerNode.namespace;
+    const podSummary = { name: explorerNode.id, namespace: namespace };
+    const container = await selectContainerForPod(podSummary);
+
+    // No container. How'd we get here. How does this Pod even work. What is going on.
+    if (!container) {
+        return;
+    }
+
+    getLogsForContainer(
+        kubectl,
+        podSummary,
+        container.name,
+        displayMode
+    );
+
+    return;
+}
+
+/**
+ * Fetches logs for a pod. If there are more than one containers,
+ * prompts the user for which container to fetch logs for.
+ */
+async function getLogsForPod(kubectl: Kubectl, podSummary: PodSummary, displayMode: LogsDisplayMode) {
+    if (!podSummary) {
+        vscode.window.showErrorMessage('Can\'t find a pod!');
+        return;
+    }
+
+    const container = await selectContainerForPod(podSummary);
+    if (!container) {
+        return;
+    }
+
+    getLogsForContainer(kubectl, podSummary, container.name, displayMode);
+}
+
+/**
+ * Gets the logs for a container in a provided pod, in a provided namespace, in a provided container.
+ */
+function getLogsForContainer(
+    kubectl: Kubectl,
+    podSummary: PodSummary,
+    containerName: string | undefined,
+    displayMode: LogsDisplayMode
+) {
+    let cmd = `logs ${podSummary.name}`;
+
+    if (podSummary.namespace) {
+        cmd = `${cmd} --namespace=${podSummary.namespace}`;
+    }
+
+    if (containerName) {
+        cmd = `${cmd} --container=${containerName}`;
+    }
+
+    if (displayMode === LogsDisplayMode.Follow) {
+        cmd = `${cmd} -f`;
+        kubectl.invokeInNewTerminal(cmd, `${podSummary.name}-${containerName}`);
+        return;
+    }
+
+    kubectl.invokeInSharedTerminal(cmd);
+}
+
+/**
+ * Searches for a pod yaml spec from the open document
+ * or from the currently selected namespace.
+ */
+async function logsForPod(kubectl: Kubectl, displayMode: LogsDisplayMode): Promise<void> {
+    const editor = vscode.window.activeTextEditor;
+
+    if (editor) {
+        return await logsForPodFromOpenDocument(kubectl, editor, displayMode);
+    }
+
+    return await logsForPodFromCurrentNamespace(kubectl, displayMode);
+}
+
+/**
+ * Finds a Pod from the open editor.
+ */
+async function logsForPodFromOpenDocument(kubectl: Kubectl, editor: vscode.TextEditor, displayMode: LogsDisplayMode) {
+    const text = editor.document.getText();
+    try {
+        const obj: {} = yaml.safeLoad(text);
+        if (!isPod(obj)) {
+            // document doesn't desribe a pod.
+            return await logsForPodFromCurrentNamespace(kubectl, displayMode);
+        }
+
+        const podSummary = {
+            name: obj.metadata.name,
+            namespace: obj.metadata.namespace
+        };
+
+        getLogsForPod(kubectl, podSummary, displayMode);
+        return;
+    } catch (ex) {
+        // pass
+    }
+
+    return await logsForPodFromCurrentNamespace(kubectl, displayMode);
+}
+
+/**
+ * Alerts the user on pods available in the current namespace.
+ */
+async function logsForPodFromCurrentNamespace(kubectl: Kubectl, displayMode: LogsDisplayMode) {
+    const namespace = await kubectlUtils.currentNamespace(kubectl);
+
+    quickPickKindName(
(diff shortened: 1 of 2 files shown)


src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,184 @@
...
+    Show,
+    Follow
+}
+
+const FOLLOW_LOGS_TERMINAL = "Kubernetes: Follow Logs";
+
+/**
+ * Fetches logs for a Pod. Handles use cases for fetching pods
+ * from an open document, or from the current namespace.
+ * @param kubectl
```
itowlson: There's no point having JSDoc for parameters if it just states the parameter name, I feel?

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,184 @@
...
+        cmd = `${cmd} --namespace=${podSummary.namespace}`;
+    }
+
+    if (containerName) {
+        cmd = `${cmd} --container=${containerName}`;
+    }
+
+    if (displayMode === LogsDisplayMode.Follow) {
+        cmd = `${cmd} -f`;
+        kubectl.invokeInNewTerminal(cmd, FOLLOW_LOGS_TERMINAL);
```
itowlson: Suggest follow terminals be labelled as to which pod/container they're following - otherwise you could have several of them and not know which is which.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,184 @@
...
+    kubectl.invokeInSharedTerminal(cmd);
+}
+
+/**
+ * Searches for a pod yaml spec from the open document
+ * or from the currently selected namespace.
+ * @param kubectl
+ * @param displayMode
+ */
+async function findPod(kubectl: Kubectl, displayMode: LogsDisplayMode) {
```
itowlson: Consider declaring a return type.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,184 @@
...
+        if (!isPod(obj)) {
+            return;
+        }
+
+        const podSummary = {
+            name: obj.metadata.name,
+            namespace: obj.metadata.namespace
+        };
+
+        getLogsForPod(kubectl, podSummary, displayMode);
```
itowlson: Hmm, seems like we have a naming issue here - `findPodFromOpenDocument` implies it finds a pod, it says nothing about fetching logs.  Of course the naming issue is very likely inherited from `extension.ts` but it might be good to tidy this if we are doing a refactor.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,184 @@
...
+
+        return;
+    } catch (ex) {
+        host.showErrorMessage("Could not retrieve Pod from Active Document");
```
itowlson: "active document" (no capitals).

Actually, this sounds like we could retrieve a pod from the document, whereas I think it really means we couldn't retrieve _logs_ for the pod described by the active document... is that correct?  If so can we rephrase please?

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,184 @@
...
+        [kuberesources.allKinds.pod],
+        { nameOptional: false },
+        (pod) => {
+
+            const podSummary: PodSummary = {
+                name: pod.split('/')[1],
+                namespace: namespace // should figure out how to handle namespaces.
+            };
+
+            getLogsForPod(kubectl, podSummary, displayMode);
```
itowlson: Same comment about `findPod` function naming.  In the old code, `findPod` took a callback by the look of it; that callback now seems to have been hardwired into the function definitions.

src/extension.ts
```diff
@@ -150,15 +151,15 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
         registerCommand('extension.vsKubernetesUseContext', useContextKubernetes),
         registerCommand('extension.vsKubernetesClusterInfo', clusterInfoKubernetes),
         registerCommand('extension.vsKubernetesDeleteContext', deleteContextKubernetes),
-        registerCommand('extension.vsKubernetesUseNamespace', (obj) => { useNamespaceKubernetes(kubectl, obj); } ),
+        registerCommand('extension.vsKubernetesUseNamespace', (explorerNode: explorer.KubernetesObject) => { useNamespaceKubernetes(kubectl, explorerNode); } ),
```
itowlson: I get this but I'd prefer it to be in a separate PR as it's not related to logs (and yeah I know I do this sometimes as well)... just makes it clearer when reading the diffs in six months' time.

src/components/kubectl/logs.ts
bnookala: 👍

src/components/kubectl/logs.ts
bnookala: 👍

src/components/kubectl/logs.ts
bnookala: 👍

src/components/kubectl/logs.ts
bnookala: 👍

src/components/kubectl/logs.ts
bnookala: 👍

src/extension.ts
bnookala: considering it's not a drastic functional change, and it's fairly small, it should be okay to leave it here.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,161 @@
...
+async function getLogsForExplorerPod(
+    kubectl: Kubectl,
+    explorerNode: ResourceNode,
+    displayMode: LogsDisplayMode
+) {
+    const namespace = explorerNode.namespace;
+    const podSummary = { name: explorerNode.id, namespace: namespace };
+    const container = await selectContainerForPod(podSummary);
+
+    // No container. How'd we get here. How does this Pod even work. What is going on.
```
itowlson: _I_ think you should add "Inconceivable!!!" here but it is probably not worth stalling the PR just for a Princess Bride reference.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Finds a Pod from the open editor.
+ */
+async function logsForPodFromOpenDocument(kubectl: Kubectl, editor: vscode.TextEditor, displayMode: LogsDisplayMode) {
+    const text = editor.document.getText();
+    try {
+        const obj: {} = yaml.safeLoad(text);
+        if (!isPod(obj)) {
+            return;
```
itowlson: I think you have changed the logic here.  Previously if the open document was not a pod, it would fall through to the prompt case (same as if there was no active document).  It looks like now if the open document is not a pod it will silently do nothing.  The old behaviour was intentional and we should keep it.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,161 @@
...
+
+        const podSummary = {
+            name: obj.metadata.name,
+            namespace: obj.metadata.namespace
+        };
+
+        getLogsForPod(kubectl, podSummary, displayMode);
+        return;
+    } catch (ex) {
+        host.showErrorMessage("Could not retrieve logs for pod for active document.");
```
itowlson: This seems wrong.  If I have, say, an application code file open, and ask for logs, I'd like to be able to select the pod running the application rather than have to close the document or open the pod YAML first.  Again, the old behaviour was to swallow the YAML parse exception and fall through to the 'select a container' case.

src/components/kubectl/logs.ts
bnookala: fixed!

src/components/kubectl/logs.ts
itowlson: Actually I think this can happen if the pod contains multiple containers but the user cancels selection, so maybe we need to remove this comment - it might be misleading.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,164 @@
...
+ * Fetch logs from a Pod, when selected from the Explorer.
+ */
+async function getLogsForExplorerPod(
+    kubectl: Kubectl,
+    explorerNode: ResourceNode,
+    displayMode: LogsDisplayMode
+) {
+    const namespace = explorerNode.namespace;
+    const podSummary = { name: explorerNode.id, namespace: namespace };
+    const container = await selectContainerForPod(podSummary);
```
itowlson: From here onwards this appears to be the same as `getLogsForPod` - could we just call that method?

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * Finds a Pod from the open editor.
+ */
+async function logsForPodFromOpenDocument(kubectl: Kubectl, editor: vscode.TextEditor, displayMode: LogsDisplayMode) {
+    const text = editor.document.getText();
+    try {
+        const obj: {} = yaml.safeLoad(text);
+        if (!isPod(obj)) {
```
itowlson: Consider flipping the condition - that way you don't need to have two places where you fall through to `logsForPodFromCurrentNamespace`.

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,164 @@
...
+            // document doesn't desribe a pod.
+            return await logsForPodFromCurrentNamespace(kubectl, displayMode);
+        }
+
+        const podSummary = {
+            name: obj.metadata.name,
+            namespace: obj.metadata.namespace
+        };
+
+        getLogsForPod(kubectl, podSummary, displayMode);
```
itowlson: `await`?

src/components/kubectl/logs.ts
```diff
@@ -0,0 +1,164 @@
...
+        [kuberesources.allKinds.pod],
+        { nameOptional: false },
+        (pod) => {
+
+            const podSummary: PodSummary = {
+                name: pod.split('/')[1],
+                namespace: namespace // should figure out how to handle namespaces.
+            };
+
+            getLogsForPod(kubectl, podSummary, displayMode);
```
itowlson: `await`?

bnookala: fixed

src/components/kubectl/logs.ts
bnookala: fixed

src/components/kubectl/logs.ts
bnookala: fixed

src/components/kubectl/logs.ts
bnookala: fixed

PR #281, opened 2018-07-06
Fixing pod listing for services in explorer view

shreyu86:
While refactoring for #277 missed the where the `matchLabels` are empty (as current namespace is selected) for  `Services` -> `<Service-Name>` and hence pods under an service were not showing up and the extension threw an error. This PR fixes that. Sorry about breaking this functionality.

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -333,12 +333,16 @@ class KubernetesSelectorResource extends KubernetesResource {
         if (!this.selector) {
             return [];
         }
-        const selectors = [];
-        for (const sel in this.selector.matchLabels) {
-            selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
+        let selectorString = "";
+        if (this.selector.matchLabels) {
+            const selectors = [];
+            selectors.push("-l");
+            for (const sel in this.selector.matchLabels) {
+                selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
+            }
+            selectorString = selectors.join(",");
         }
-        const selectorString = selectors.join(",");
-        const pods = await kubectl.fromLines(`get po -l ${selectorString}`);
+        const pods = await kubectl.fromLines(`get po ${selectorString}`);
         if (failed(pods)) {
             host.showErrorMessage(pods.error[0]);
             return [new DummyObject("Error")];
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -339,7 +339,7 @@ class KubernetesSelectorResource extends KubernetesResource {
             for (const sel in this.selector.matchLabels) {
                 selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
             }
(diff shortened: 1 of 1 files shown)


src/explorer.ts
```diff
@@ -333,12 +333,16 @@ class KubernetesSelectorResource extends KubernetesResource {
...
-        for (const sel in this.selector.matchLabels) {
-            selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
+        let selectorString = "";
+        if (this.selector.matchLabels) {
+            const selectors = [];
+            selectors.push("-l");
+            for (const sel in this.selector.matchLabels) {
+                selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
+            }
+            selectorString = selectors.join(",");
```
itowlson: This looks like it will result in `-l,foo=bar" (with a comma) instead of `-l foo=bar`.  Can you confirm it's correct still correct if there are matchLabels present?

shreyu86: fixed this.

src/explorer.ts
```diff
@@ -333,12 +333,15 @@ class KubernetesSelectorResource extends KubernetesResource {
...
-        const selectors = [];
-        for (const sel in this.selector.matchLabels) {
-            selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
+        let selectorString = "";
+        if (this.selector.matchLabels) {
+            const selectors = [];
+            for (const sel in this.selector.matchLabels) {
+                selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
+            }
+            selectorString = "-l" + selectors.join(",");
```
itowlson: `-l ${selectors.join(",")}`?  Seems like we'd normally put a space between an option flag and its value, though maybe it isn't formally necessary...?

itowlson: Uh there were meant to be backticks around that but I guess Markdown ated them

shreyu86: fixing this.

src/explorer.ts
```diff
@@ -333,12 +333,15 @@ class KubernetesSelectorResource extends KubernetesResource {
         if (!this.selector) {
             return [];
         }
-        const selectors = [];
-        for (const sel in this.selector.matchLabels) {
-            selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
+        let selectorString = "";
+        if (this.selector.matchLabels) {
```
shreyu86: @itowlson, realized after testing this morning that had missed porting over the logic for [pod selection](https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/kubectlUtils.ts#L177) based on [labels](https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/kubectlUtils.ts#L190) and wanted to bring attention to that. I can send an PR later today to fix this. :( sorry about missing those conditions.

PR #283, opened 2018-07-07
Fixing missing selector conditions

shreyu86:
With the refactor for #272 I [missed couple](https://github.com/Azure/vscode-kubernetes-tools/pull/281#discussion_r200733572) of conditions selector in the `Explorer View`. 

As a consequence it broke the listing of Services and Deployments in explorer view causing `kubectl` to return all pods instead of the pods belonging only to the selected service. 

This PR converts code in `getPods` to a re-usable function `getResourceSelectorString` in `kubectlUtils` and fixes the explorer view in services returning only the pods belonging to the service. 
Sorry about missing one of the selection cases earlier. 😞

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -333,14 +333,7 @@ class KubernetesSelectorResource extends KubernetesResource {
         if (!this.selector) {
             return [];
         }
-        let selectorString = "";
-        if (this.selector.matchLabels) {
-            const selectors = [];
-            for (const sel in this.selector.matchLabels) {
-                selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
-            }
-            selectorString = `-l ${selectors.join(",")}`;
-        }
+        const selectorString = kubectlUtils.getResourceSelectorString(this.selector);
         const pods = await kubectl.fromLines(`get po ${selectorString}`);
         if (failed(pods)) {
             host.showErrorMessage(pods.error[0]);
--- a/src/kubectlUtils.ts
+++ b/src/kubectlUtils.ts
@@ -174,6 +174,22 @@ export async function getPods(kubectl: Kubectl, selector: any, namespace: string
     if (ns === 'all') {
         nsFlag = '--all-namespaces';
     }
+    const labelStr = getResourceSelectorString(selector);
+    const pods = await kubectl.asJson<KubernetesCollection<Pod>>(`get pods -o json ${nsFlag} ${labelStr}`);
+    if (failed(pods)) {
+        vscode.window.showErrorMessage(pods.error[0]);
+        return [];
+    }
+    return pods.result.items.map((item) => {
+        return {
+            name: item.metadata.name,
+            namespace: item.metadata.namespace,
+            nodeName: item.spec.nodeName
+        };
+    });
+}
+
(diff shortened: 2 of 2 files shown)


src/explorer.ts
```diff
@@ -333,14 +333,7 @@ class KubernetesSelectorResource extends KubernetesResource {
...
         }
-        let selectorString = "";
-        if (this.selector.matchLabels) {
-            const selectors = [];
-            for (const sel in this.selector.matchLabels) {
-                selectors.push(`${sel}=${this.selector.matchLabels[sel]}`);
-            }
-            selectorString = `-l ${selectors.join(",")}`;
-        }
+        const selectorString = kubectlUtils.getResourceSelectorString(this.selector);
```
itowlson: I'm getting lost in these changes I'm afraid.  It now looks like `getPods` isn't called from _anywhere_ so I'm not sure of the value of updating it.  And `KubernetesSelectorResource.getChildren` calling `kubectl.fromLines("get po...")` will return a _different schema_ from what `getPods` used to return.

It seems like we probably want to reset `KubernetesSelectorResource` and `getPods` to what they used to be except with an additional status field.  And perhaps we need to modify `KubernetesNodeResource.getChildren` to call `getPods` with the `all` namespace option.  That would keep us closest to how I think we were before we added the status GUI.

shreyu86:
@itowlson Thanks for your feedback. Sorry about causing confusion. `getPods` was updated as the selector logic was extracted into a new function for re-use in `KubernetesSelectorResource.getChildren`. Happy to revert changes per your comments. Which changes would you like to see backed out and re-implemented?

itowlson:
I saw why `getPods` was updated... the trouble is I don't see it being _called_ any more.

If we want to do the 'back out and restart' approach then I'd wind back to before you put the status display in, then re-approach knowing we've learned about nodes, deployments and services (which I think all go through `getPods` so we could maybe centralise the status collection there).

If you don't want to unpick that far -- and it may be tricky given what else has been changing in the Explorer module -- then I'd say your implementation of `getPods` looks pretty good now, we just maybe need to consider routing `KubernetesSelectorResource` and `KubernetesNodeResource` back through it instead of calling `kubectl get po` directly.  This is probably less work than unpicking, if we can get the changes clear in our heads!

shreyu86:
I like the _back out and restart_ option and the idea of centralizing fetch pod info with status through `getPods` to add the status indicators back. If we re-route `KubernetesSelectorResource` and `KubernetesNodeResource` through `getPods` in the current state it would mean adding status to each pod which might lead to a redundant call to get pod status.

itowlson:
Okay.  Shall I leave this with you, or do you want me to do the rollback?  If we're going to take this approach, I'd like to get at least the rollback done quite quickly, because we're overdue to release an update and I'd prefer not to do that with the existing service pods bug.

shreyu86:
@itowlson if you can rollback that would be perfect, I will get started on the fix then. Thanks a lot!

shreyu86:
@itowlson Is it ok if this PR is closed and a new issue is opened to re-implement the status indicators?

itowlson:
Yes, closing this one and looking forward to the new one!

PR #285, opened 2018-07-09
Roll back pod status changes as foundation for rework

itowlson:
We have had work in progress on displaying pod status, but due to other work happening in parallel (e.g. node pod expansion) the code has diverged in multiple directions and become hard to maintain, leading to various display bugs and regressions.

This PR resets the pod expansion code to roughly where it would have been if we had merged the node pod expansion work but deferred the pod status work.  It hopefully provides a more consistent and stable foundation to re-implement pod status indicators across the board.  It is not a fire-and-the-sword rollback as:

* It keeps the node pod expansion code (and other subsequent explorer fixes/features)
* It keeps the linting changes made to the explorer module
* It keeps the helper functions for line parsing, getting status icons, etc. - these may not be used in this interim stage but it saves us having to recreate them (as they were not problematic)
* It keeps the status icon files (as they were not problematic)

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -180,18 +180,14 @@ class KubernetesResourceFolder extends KubernetesFolder {
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
-        const childrenObjects = await kubectl.fromLines(`get ${this.kind.abbreviation}`);
-        if (failed(childrenObjects)) {
-            host.showErrorMessage(childrenObjects.error[0]);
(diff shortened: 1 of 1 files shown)


itowlson:
@shreyu86 I wasn't able to add you as a reviewer, but it would be great if you could check this PR, as you probably know this code pretty well by now!  Thanks!

shreyu86:
Thanks on creating the rollback @itowlson, I will review this soon.

shreyu86:
Looks good, thanks again @itowlson

PR #296, opened 2018-07-10
Use Context for Minikube clusters

### LINKED ISSUE #290: Don't have the option "Set as current cluster"
Build: 4651

1. Open the Kubernetes: cluster explorer
2. Right click the "minikube" cluster

Don't have the option "Set as current cluster"
image

But right click other cluster have the option "Set as current cluster"

bnookala:
Fixes #290 & adds the minikube logo.


src/explorer.ts
```diff
@@ -32,7 +35,7 @@ export interface KubernetesObject {
     readonly id: string;
     readonly metadata?: any;
     getChildren(kubectl: Kubectl, host: Host): vscode.ProviderResult<KubernetesObject[]>;
-    getTreeItem(): vscode.TreeItem | Thenable<vscode.TreeItem>;
+    getTreeItem(type: string | undefined): vscode.TreeItem | Thenable<vscode.TreeItem>;
```
itowlson: I'm not keen on this - the only place the `type` information can come from is surely the object itself, so why would it need to pass it to its own member function?  Suggest that things that are interested in `type` implement it as a field and inspect it from their `getTreeItem` implementations...?

src/explorer.ts
```diff
@@ -117,23 +120,25 @@ class KubernetesContext implements KubernetesObject {
         ];
     }
 
-    getTreeItem(): vscode.TreeItem | Thenable<vscode.TreeItem> {
+    getTreeItem(type: string | undefined): vscode.TreeItem | Thenable<vscode.TreeItem> {
         const treeItem = new vscode.TreeItem(this.id, vscode.TreeItemCollapsibleState.Collapsed);
-        treeItem.contextValue = "vsKubernetes.cluster";
+        treeItem.contextValue = type ? type : KUBERNETES_CLUSTER;
```
itowlson: This definitely feels like it should be based on a `this.type` field - the type is an attribute of the object, not a transient parameter that might vary from method call to method call.

bnookala:
@itowlson good points all around. I've changed the implementation to align with your request.

src/explorer.ts
```diff
@@ -70,12 +73,13 @@ export class KubernetesExplorer implements vscode.TreeDataProvider<KubernetesObj
 
     private async getClusters(): Promise<KubernetesObject[]> {
         const contexts = await kubectlUtils.getContexts(this.kubectl);
-        return contexts.map((context): KubernetesContext => {
+        return contexts.map((context): ClusterContext => {
```
itowlson: This type declaration seems wrong (and seems like it was wrong before).  `kubectlUtils.getContexts` returns `kubectlUtils.KubectlContext[]` - a model object not an explorer object.  (But funnily enough it seems like TypeScript is ignoring the type declaration and inferring `context` as `KubectlContext`... weird...)

src/explorer.ts
```diff
@@ -101,8 +105,9 @@ class DummyObject implements KubernetesObject {
     }
 }
 
-class KubernetesContext implements KubernetesObject {
-    constructor(readonly id: string, readonly metadata?: kubectlUtils.KubectlContext) {
+class ClusterContext implements KubernetesObject {
```
itowlson: I don't get the value in renaming this.  If we're going to rename it, I'd suggest something like `ContextNode` or `ContextObject` or `KubernetesContextObject`.

src/explorer.ts
```diff
@@ -119,21 +124,29 @@ class KubernetesContext implements KubernetesObject {
 
     getTreeItem(): vscode.TreeItem | Thenable<vscode.TreeItem> {
         const treeItem = new vscode.TreeItem(this.id, vscode.TreeItemCollapsibleState.Collapsed);
-        treeItem.contextValue = "vsKubernetes.cluster";
-        treeItem.iconPath = vscode.Uri.file(path.join(__dirname, "../../images/k8s-logo.png"));
+        treeItem.contextValue = this.clusterType;
+
+        let iconPath: vscode.Uri;
```
itowlson: Consider breaking this block out into a separate function that returns the icon path, so we can make this `const` - a la `const iconPath = getIconPathForClusterType();`

src/explorer.ts
bnookala: It seems like it's inferring the correct type for me…

src/explorer.ts
bnookala: fixed.

src/explorer.ts
bnookala: fixed.

PR #302, opened 2018-07-12
If HOME is defined on Windows, then use that rather than USERPROFILE

### LINKED ISSUE #301: Error connecting to Azure / AKS
I'm having trouble getting the extension to work against my Azure / AKS clusters. Specifically, the Add Cluster wizard fails to recognize my Azure logon on the AZ CLI, resulting in the following error

image

I've checked that I am logged in to the azure cli and can access my resources already.

* Azure CLI version is `2.0.41`
* Operating System is Windows 10
* Code version is `1.25.0`

Is there a way to enable logging on the executed commands to check why `az` is failing for it?

itowlson:
The Azure CLI and kubectl both use the HOME environment variable, falling back on USERPROFILE only if HOME is not defined.  We had a bug where, when we spawned them, if we were on Windows, we always set HOME to the value of USERPROFILE.  So if a user had a custom HOME, spawning would break.

Fixes #301.

### DIFF
--- a/src/components/download/download.ts
+++ b/src/components/download/download.ts
@@ -1,0 +1,39 @@
+const home = process.env['HOME'];
+
+import * as download_core from 'download';  // NEVER do a naked import of this as it corrupts $HOME on Windows - use wrapper functions from this module (or export the download function from this module if you really need the flexibility of the core implementation)
+import * as path from 'path';
+import * as stream from 'stream';
+import * as tmp from 'tmp';
+
+import { succeeded, Errorable } from '../../errorable';
+
+// Fix download module corrupting HOME environment variable on Windows
+// See https://github.com/Azure/vscode-kubernetes-tools/pull/302#issuecomment-404678781
+// and https://github.com/kevva/npm-conf/issues/13
+if (home) {
+    process.env['HOME'] = home;
+}
+
+type DownloadFunc =
+    (url: string, destination?: string, options?: any)
+         => Promise<Buffer> & stream.Duplex; // Stream has additional events - see https://www.npmjs.com/package/download
+
+const download: DownloadFunc = download_core;
+
+export async function toTempFile(sourceUrl: string): Promise<Errorable<string>> {
+    const tempFileObj = tmp.fileSync({ prefix: "vsk-autoinstall-" });
+    const downloadResult = await to(sourceUrl, tempFileObj.name);
+    if (succeeded(downloadResult)) {
+        return { succeeded: true, result: tempFileObj.name };
+    }
+    return { succeeded: false, error: downloadResult.error };
+}
+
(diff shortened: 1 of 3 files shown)


itowlson:
@tomasr If you want to give this a try in your 'proper' environment (before I asked you to muddy the waters) then your feedback would be very welcome!

tomasr:
@itowlson Well, this is interesting. I'm running your patch in the debugger, and somehow, buy the time that line gets hit, `$HOME` exists, but it has been overwritten to match `$USERPROFILE`. Wonder if it's some other extension that is messing this up....

itowlson:
@tomasr That's _really_ strange.  I don't _think_ we blat $HOME anywhere else but I'll investigate...

itowlson:
It almost looks like the integrated terminal process and the VS Code process (or the process that runs extensions) have different environment blocks (at least for the HOME variables).  That surely can't be right...

itowlson:
I've verified that in a 'dummy' extension, `process.env['HOME']` has the value set in the parent shell... but at the very top of `activate` in our extension, `process.env['HOME']` has been reset to the USERPROFILE value.  So if we're doing it -- and if both of us are seeing it then it _must_ be our extension, not something else -- then it must be _before_ `activate` runs...

itowlson:
`$HOME` gets corrupted during the import of the `./components/installer/installer.ts` module.

Score another one for the maxim of always blaming my own code first!

itowlson:
Spoke to soon.  `installer` imports the NPM `download` module, and it looks like _that_ import is corrupting $HOME.

itowlson:
I can reproduce the `download` naughtiness in a dummy extension, even with the latest version of `download`.  It's not obvious from the `download` sources how it's corrupting HOME so it could be a second-level dependency.  Digging.

itowlson:
So... `installer` imports `download` imports `caw` imports `get-proxy` imports `npm-conf`... which [sets `process.env.HOME` to the result of `os.homedir()`](https://github.com/kevva/npm-conf/blob/master/lib/defaults.js#L21)... which on Windows is _always_ the user profile directory, regardless of the HOME environment variable.

Fixing this should be... interesting.

tomasr:
Oh that's.... beyond silly. Thanks for looking into it, @itowlson. There's probably not much you can do about it. Let's just chalk it up to node being.... node.

itowlson:
@tomasr I've patched around the download module naughtiness and it still seems to be working for me - it's pretty ugly, but at least for our usage, I don't think `download` depends critically on $HOME being aliased to $USERPROFILE.  Would you mind giving it another try please?  #nosurrender

If this doesn't work, though, then yeah, we may have to chalk it up to node being node and document it as a known issue - thanks for being understanding about it.  #maybealittlesurrender

tomasr:
@itowlson Worked perfectly. Many thanks, and sorry for making you spend so much time on this!

PR #308, opened 2018-07-13
Pod status Indicator re-implementation

shreyu86:
Implements pod status indicators in explorer view.
Changes:
* `getPods` uses `fromLines` as compared to the JSON pod output perviously. 
* `get pods -o wide` to include node information so as to not break `PodInfo` 
* `getChildren` in `KubernetesResourceFolder` checks if the resource being explored is pod and uses `getPods`

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -201,6 +201,12 @@ class KubernetesResourceFolder extends KubernetesFolder {
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        if (this.kind === kuberesources.allKinds.pod) {
+            const pods = await kubectlUtils.getPods(kubectl, null, null);
+            return pods.map((pod) => {
+                return new KubernetesResource(this.kind, pod.name, pod);
+            });
+        }
         const childrenLines = await kubectl.asLines(`get ${this.kind.abbreviation}`);
         if (failed(childrenLines)) {
             host.showErrorMessage(childrenLines.error[0]);
@@ -240,6 +246,9 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
             this.kind === kuberesources.allKinds.secret ||
             this.kind === kuberesources.allKinds.configMap) {
             treeItem.contextValue = `vsKubernetes.resource.${this.kind.abbreviation}`;
+            if (this.kind === kuberesources.allKinds.pod && this.metadata.status !== null) {
+                treeItem.iconPath = getIconForPodStatus(this.metadata.status.toLowerCase());
+            }
         }
         if (this.namespace) {
             treeItem.tooltip = `Namespace: ${this.namespace}`;  // TODO: show only if in non-current namespace?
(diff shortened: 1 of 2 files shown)


src/explorer.ts
```diff
@@ -240,6 +246,9 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
             this.kind === kuberesources.allKinds.secret ||
             this.kind === kuberesources.allKinds.configMap) {
             treeItem.contextValue = `vsKubernetes.resource.${this.kind.abbreviation}`;
+            if (this.kind === kuberesources.allKinds.pod && this.metadata.status !== null) {
```
itowlson: For extra safety, consider `if (this.kind === … && this.metadata && this.metadata.status)`

src/kubectlUtils.ts
```diff
@@ -189,16 +189,18 @@ export async function getPods(kubectl: Kubectl, selector: any, namespace: string
         labelStr = "--selector=" + labels.join(",");
     }
 
-    const pods = await kubectl.asJson<KubernetesCollection<Pod>>(`get pods -o json ${nsFlag} ${labelStr}`);
+    const pods = await kubectl.fromLines(`get pods -o wide ${nsFlag} ${labelStr}`);
```
itowlson: What do we gain by changing this?  `asJson` should still be able to give us a status I think (via `spec.status`?) - is there a reason to prefer switching to `fromLines`?  (To be clear, I think the change is harmless, but I have an aversion to change for change's sake.)

src/kubectlUtils.ts
```diff
@@ -189,16 +189,18 @@ export async function getPods(kubectl: Kubectl, selector: any, namespace: string
...
+
+    return pods.result.map((item) => {
         return {
-            name: item.metadata.name,
-            namespace: item.metadata.namespace,
-            nodeName: item.spec.nodeName
+            name: item.name,
+            namespace: item.namespace || ns,
+            nodeName: item.node,
+            status: item.status
```
itowlson: I think we should change the definition of `PodInfo` to include the `status` member.

src/explorer.ts
shreyu86: refactored the code to reflect this change.

src/kubectlUtils.ts
shreyu86: refactored `PodInfo` to include `status`

src/kubectlUtils.ts
shreyu86: The specific reason to change this to `fromLines` was to get aggregated pod status which is very handy to add the status indicator icons. `asJson` returns `spec.status` which would say running even if one of the pods has in a crash loop state. It would be possible to get the same aggregated status from `asJson` but it involves aggregating the status manually by implementing something similar to `kubectl` [printer](https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L547), I did go down this route but found the solution to be convoluted hence I preferred `fromLines`

PR #310, opened 2018-07-15
Improve startup speed

### LINKED ISSUE #307: Startup is slow
The extension takes a long time to start up, especially under the debugger.  It should be faster, perhaps even bigly faster.

itowlson:
Some of our dependencies take a very long time to load.  This PR defers loading these dependencies until they are needed`*`.

Fixes #307.

~~`*` _conditions apply_... specifically, the `restify` dependency is not yet fully deferred, as we currently need to start HTTP servers during activation in order to parcel out ports to the various providers.  I'll keep noodling away on this and push an update if I can see a reliable way to do this.~~  The `restify` dependency is now _even more deferred_.

### DIFF
--- a/src/components/clusterprovider/azure/azureclusterprovider.ts
+++ b/src/components/clusterprovider/azure/azureclusterprovider.ts
@@ -1,4 +1,4 @@
-import * as restify from 'restify';
+import restify = require('restify');
 import * as portfinder from 'portfinder';
 import * as clusterproviderregistry from '../clusterproviderregistry';
 import * as azure from './azure';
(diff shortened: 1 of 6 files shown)


radu-matei:
Did a bit of digging on TypeScript dynamic loading, and it turns out you can have something like this - example for the `installer`, as it was the easiest to approach:

- remove the initial import of `installer` at the beginning of `extension.ts`:

```
//import { installHelm, installDraft, installKubectl } from './components/installer/installer';
```

- then in the `installDependencies` function use the `import()` function to load the module, then use it:
```
export async function installDependencies() {
    const installer = await import('./components/installer/installer');
    // TODO: gosh our binchecking is untidy
    const gotKubectl = await kubectl.checkPresent('silent');
    const gotHelm = helmexec.ensureHelm(helmexec.EnsureMode.Silent);
    const gotDraft = await draft.checkPresent(DraftCheckPresentMode.Silent);

    // TODO: parallelise

    await installDependency("kubectl", gotKubectl, installer.installKubectl);
    await installDependency("Helm", gotHelm, installer.installHelm);
    await installDependency("Draft", gotDraft, installer.installDraft);

    kubeChannel.showOutput("Done");
}
```

Did not extensively test it, and it might not work in our case, but wanted to bring this up as a solution for dynamic module loading in TypeScript that is worth investigating.

itowlson:
@radu-matei I think you're right - I recall something about that going by on the TS announcements feed.  That might be a cleaner solution than lazy loading the dependencies within the modules themselves...

Issue #298, opened 2018-07-11
Pod Listing error due to kubectl client/server version/semver mismatch


bnookala:
I'll let @itowlson fill in the details here, since he noted it first! We should handle this situation more gracefully.

itowlson:
If the version of kubectl used by the extension is too far ahead of the server version then some operations return errors.  For example, I had kubectl 1.11 (latest stable I think) and k8s version 1.7 (long established ACS cluster), and expanding pod lists in the tree view gave an error (server error Not Acceptable or Not Found, can't remember which).  This is a kubectl issue for sure, but it would be great if we could either:

* Figure out the right version of kubectl to use for the active cluster, and offer to install it if necessary; or
* Detect the version mismatch and give a meaningful error rather than the bald API error

I think k8s promises that kubectl will remain compatible with up to two minor versions back, but not beyond that.

PR #312, opened 2018-07-16
Fix kubectl versioning woes

### LINKED ISSUE #306: Inconsistent choice of kubectl executable
If a user has configured a location for the `kubectl` executable (which we do automatically if they do an Install Dependencies), _and_ there is a version of `kubectl` on the path, then we are inconsistent about which we use.  Specifically, the configured on is used for 'silent' invocations, such as when populating the tree view; but in a spawned terminal, the one on the path wins out.  This can lead to different behaviour if the two are different versions - e.g. if the configured version is far ahead of the server but the path version is in sync, then the tree view may display NotAcceptable errors which the user cannot reproduce or diagnose in the terminal.

itowlson:
This PR addresses two versioning issues with `kubectl`:

1. Terminals could use a different version of `kubectl` from "behind-the-scenes" calls, if the user had set a location for `kubectl` in their VS Code configuration.  This occurred because "behind-the-scenes" would directly invoke the program at the configured location, whereas terminals would append the configured location to the path (so that the user could issue follow-up commands without having to type out the configured path) -- so if `'kubectl` was _also_ on the path then the path version would take precedence.  This meant the kubectl terminal was consistent with a system terminal, but made it hard to diagnose/reproduce the behaviour of "behind-the-scenes" commands.  Terminals now _prepend_ the configured location to the path, so that the configured location takes precedence.

2. Kubernetes only guarantees compatibility with at most one minor version of skew between client and server.  For example, `kubectl get pods` in v1.11 fails with server version v1.7.  This would lead to enigmatic errors, for example when expanding the tree (an Error node and toast saying something about being Not Acceptable).  We now check `kubectl` failures and, if the versions are too far out of sync, display an additional warning to this effect.  (This may be too aggressive at the moment, e.g. might not want to warn more than once per cluster, for both performance and user annoyance reasons; but `kubectl` failures should be relatively rare and a versioning issue is likely to be non-obvious when it hits.)

Fixes #306 and #298.

### DIFF
--- a/src/components/kubectl/compatibility.ts
+++ b/src/components/kubectl/compatibility.ts
@@ -1,4 +1,4 @@
-import { Kubectl } from '../../kubectl';
+import { ShellResult } from '../../shell';
 
 interface CompatibilityGuaranteed {
     readonly guaranteed: true;
@@ -17,8 +17,8 @@ export function isGuaranteedCompatible(c: Compatibility): c is CompatibilityGuar
     return c.guaranteed;
 }
 
-export async function check(kubectl: Kubectl): Promise<Compatibility> {
-    const sr = await kubectl.invokeAsync('version --short');
+export async function check(kubectlInvokeAsync: (cmd: string) => Promise<ShellResult>): Promise<Compatibility> {
+    const sr = await kubectlInvokeAsync('version --short');
     if (sr.code !== 0) {
         return {
             guaranteed: false,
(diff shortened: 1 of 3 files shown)


src/components/kubectl/compatibility.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (sr.code !== 0) {
+        return {
+            guaranteed: false,
+            didCheck: false,
+            clientVersion: '',
+            serverVersion: ''
+        };
+    }
+
+    const lines = sr.stdout.split('\n').map((l) => l.trim()).filter((l) => l.length > 0);
```
bnookala: you can use the json output from `kubectl version -o json`, if that makes life any easier here…

itowlson: TIL! Thank you!

PR #313, opened 2018-07-17
Install Minikube if not present

### LINKED ISSUE #300: Extension should offer to install Minikube, when no binary is found.
Currently, we assume minikube is installed when invoking minikube commands. We should offer to install minikube when it's not present.

itowlson:
This adds Minikube to the 'Install dependencies' prompt, and adds missing dependency detection to the Minikube 'create cluster' wizard.  As dependencies are installed in a private location rather than on the path, this also requires some hardening of the Minikube detection and launch logic which brings it more in line with kubectl and Draft.

Fixes #300.

### DIFF
--- a/src/components/clusterprovider/minikube/minikube.ts
+++ b/src/components/clusterprovider/minikube/minikube.ts
@@ -1,10 +1,80 @@
 'use strict';
 
 import * as vscode from 'vscode';
-import { shell, ShellResult } from '../../../shell';
+import { Shell, ShellResult } from '../../../shell';
+import { Host } from '../../../host';
+import { FS } from '../../../fs';
+import * as binutil from '../../../binutil';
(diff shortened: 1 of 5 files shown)


itowlson:
There is an issue (slightly wider than downloading missing Minikube) that Minikube requires kubectl to be on the path.  This will not be the case if the user has used 'Install dependencies' to get kubectl, so we may need to inject that into the path (we have the technology for this in our terminal code, which may in fact be a better way to address some of the pathing stuff in the launchers).  We also need to see if we can detect and report other startup errors, e.g. if Virtual Box is not present - this currently stalls the Minikube process at a 'send crash information?' prompt.  I don't know if `minikube start` has a non-interactive mode - it's not obvious from the help...

Issue #309, opened 2018-07-13
Status Indicators for Helm Releases


shreyu86:
It will be nice to see status indicators (Failed/Deployed) for Helm Releases.
Example of WIP Approach:
<img width="203" alt="screen shot 2018-07-13 at 11 51 22 am" src="https://user-images.githubusercontent.com/311217/42708766-1970d5c4-8693-11e8-8c42-a6c98719b2a5.png">

shreyu86:
Closed via. #318

PR #318, opened 2018-07-19
Adding status icons for Helm release status

shreyu86:
Adding status indicators to helm releases (deployed vs other status) per: #309 

- Changed `parseLineOutput` to accept an regex which describes how to parse different column based outputs. `kubectl` and `helm` use different spacing logic for output.
- Output from `parseLineOutput` now `trim`'s the strings 
- Added helm status icons (original SVG from [`vscode-icons`](https://github.com/vscode-icons/vscode-icons/blob/master/icons/file_type_helm.svg))
- `HelmReleaseResource` has an additional member `status: string`


src/explorer.ts
```diff
@@ -416,23 +424,25 @@ export class KubernetesFileObject implements KubernetesObject {
 
 class HelmReleaseResource implements KubernetesObject {
     readonly id: string;
+    readonly status: string;
```
itowlson: This doesn't appear to be used?

src/explorer.ts
```diff
@@ -416,23 +424,25 @@ export class KubernetesFileObject implements KubernetesObject {
...
 
-    constructor(readonly name: string) {
-        this.id = "helmrelease:" + name;
+    constructor(readonly helmRelease: any) {
```
itowlson: I am trying to smother `any`s as close to birth as possible; can we pass in a name and status, or a typed interface, please?

src/helm.exec.ts
```diff
@@ -297,7 +298,7 @@ export async function helmListAll(namespace?: string): Promise<Errorable<string[
         return { succeeded: false, error: [ "Helm client is not installed" ] };
     }
 
-    const releases: string[] = [];
+    const releases = [];
```
itowlson: I suggest leaving the type annotation in: when I switch on `strictNullChecks`, the compiler will start inferring the naked literal as type `never[]` and will complain.  Even without that, declaring the type at this point ensures that adding the wrong kind of thing to it will typecheck.

src/helm.exec.ts
```diff
@@ -321,9 +322,8 @@ export async function helmListAll(namespace?: string): Promise<Errorable<string[
             }
         }
         if (lines.length > 0) {
-            lines.shift();  // remove the header line
-            const names = lines.map((l) => l.split('\t')[0].trim());
-            releases.push(...names);
+            const hemlReleases = parseLineOutput(lines, helm.HELM_OUTPUT_PARSING_REGEX);
```
itowlson: typo: `he*ml*Releases`

src/helm.ts
```diff
@@ -1,6 +1,7 @@
 export const PREVIEW_SCHEME = 'helm-template-preview';
 export const PREVIEW_URI = PREVIEW_SCHEME + '://preview';
 export const INSPECT_SCHEME = 'helm-inspect-values';
+export const HELM_OUTPUT_PARSING_REGEX = /\t+/g;
```
itowlson: Do we need the separate regexes?  I would have thought that `\s+` would still work for tab-delimited output... or oh, can Helm output include spaces within a column?

src/kubectl.ts
```diff
@@ -237,7 +237,7 @@ async function fromLines(context: Context, command: string): Promise<Errorable<{
     if (shellResult.code === 0) {
         let lines = shellResult.stdout.split('\n');
         lines = lines.filter((l) => l.length > 0);
-        const parsedOutput = parseLineOutput(lines);
+        const parsedOutput = parseLineOutput(lines, POD_OUTPUT_PARSING_REGEX);
```
itowlson: `fromLines` shouldn't be pod-specific.  Would it be correct to say `KUBECTL_OUTPUT_PARSING_REGEX`? (or `KUBECTL_OUTPUT_COLUMN_SEPARATOR`)?

src/kubectlUtils.ts
```diff
@@ -5,6 +5,8 @@ import { sleep } from "./sleep";
 import { ObjectMeta, KubernetesCollection, DataResource, Namespace, Pod, KubernetesResource } from './kuberesources.objectmodel';
 import { failed } from "./errorable";
 
+export const POD_OUTPUT_PARSING_REGEX = /\s+/g;
```
itowlson: I think I would prefer `KUBECTL_OUTPUT_COLUMN_SEPARATOR` here.  "Parsing regex" is a bit vague as to what it is.

src/kubectlUtils.ts
```diff
@@ -339,19 +341,20 @@ export async function getResourceAsJson<T extends KubernetesResource | Kubernete
  * Parse column based output which is seperated by whitespace(s) from kubectl or similar sources
  * for example, kubectl get po
  * @param lineOutput raw output with headers from kubectl or similar sources
+ * @param parsingRegex regex to parse the columns
  * @return array of objects with key as column header and value
  */
-export function parseLineOutput(lineOutput: string[]): { [key: string]: string }[] {
+export function parseLineOutput(lineOutput: string[], parsingRegex: RegExp): { [key: string]: string }[] {
```
itowlson: As this is now used for both kubectl and Helm, should we move it to a, say, `outputUtils` or `parsing` file?

src/kubectlUtils.ts
```diff
@@ -339,19 +341,20 @@ export async function getResourceAsJson<T extends KubernetesResource | Kubernete
  * Parse column based output which is seperated by whitespace(s) from kubectl or similar sources
  * for example, kubectl get po
  * @param lineOutput raw output with headers from kubectl or similar sources
+ * @param parsingRegex regex to parse the columns
```
itowlson: I would say something like "a regex for the column separators" - `parsing` doesn't tell you *what* it parses.

src/kubectl.ts
```diff
@@ -5,7 +5,7 @@ import { FS } from './fs';
 import { Shell, ShellHandler, ShellResult } from './shell';
 import * as binutil from './binutil';
 import { Errorable } from './errorable';
-import { parseLineOutput } from './kubectlUtils';
+import { parseLineOutput, POD_OUTPUT_PARSING_REGEX } from './kubectlUtils';
```
itowlson: The regex appears to be used only in `kubectl`, not in `kubectlUtils` - why not declare it here, rather than declaring it there and having to import?

src/explorer.ts
shreyu86: This is used at: `treeItem.iconPath = getIconForHelmRelease(this.helmRelease.status.toLowerCase());`

src/explorer.ts
shreyu86: Fixing this.

src/helm.ts
shreyu86: The `UPDATED` column in helm output contains a timestamp which contains spaces hence the regex with tabs is needed here as compared to `kubectl` output.

src/kubectl.ts
shreyu86: Fixed

src/kubectlUtils.ts
shreyu86: Added new description

shreyu86:
@itowlson thanks for the feedback, I have made changes.

src/explorer.ts
shreyu86: Fixed

src/outputUtils.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Parse column based output which is seperated by whitespace(s) from kubectl or similar sources
+ * for example, kubectl get po
+ * @param lineOutput raw output with headers from kubectl or similar sources
+ * @param parsingRegex a regex for the column separators
```
itowlson: I'd prefer to change the name of this argument to something like `columnSeparator` though it's not a showstopper.

shreyu86: Changed name.

itowlson:
Thanks @shreyu86!

PR #321, opened 2018-07-25
Fix error listing charts in Helm repo if repo URL contained a path

itowlson:
If a Helm repo URL contained a path, e.g. `https://azure.github.io/brigade`, we were blatting that path with `index.yaml` instead of appending `index.yaml` to the end of the path.  This prevented us from listing the charts in the repo.  This PR fixes that issue, and also adds hover detail for errors should there be any more lurking in there.  WHICH THERE ARE NOT.

### DIFF
--- a/src/helm.repoExplorer.ts
+++ b/src/helm.repoExplorer.ts
@@ -44,7 +44,7 @@ export class HelmRepoExplorer implements vscode.TreeDataProvider<HelmObject> {
     private async getHelmRepos(): Promise<HelmObject[]> {
         const repos = await listHelmRepos();
         if (failed(repos)) {
-            return [ new HelmError('Unable to list Helm repos') ];
+            return [ new HelmError('Unable to list Helm repos', repos.error[0]) ];
         }
         return repos.result;
     }
@@ -55,10 +55,12 @@ export class HelmRepoExplorer implements vscode.TreeDataProvider<HelmObject> {
 }
 
 class HelmError implements HelmObject {
-    constructor(private readonly text: string) {}
+    constructor(private readonly text: string, private readonly detail: string) {}
 
     getTreeItem(): vscode.TreeItem {
-        return new vscode.TreeItem(this.text);
+        const treeItem = new vscode.TreeItem(this.text);
+        treeItem.tooltip = this.detail;
+        return treeItem;
     }
 
     async getChildren(): Promise<HelmObject[]> {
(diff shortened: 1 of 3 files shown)


technosophos:
Last night I realized that if we wanted this feature to work more or less in parity with `helm search`, then the better thing to do here would be to load the cached indices from `$HELM_HOME/repositories/cache/*`. A refresh operation would then do a `helm repo update` to fetch the most recent version of the indices.

There is one case in which the current behavior (in this PR) could potentially cause an issue. Since the current version fetches the index directly, but `helm install` uses the cached index, it is possible that someone could see a version of the chart in the tree viewer that they cannot install with `helm install` (until they do a `helm repo update`).

I wouldn't call this a showstopper, though, which is why I LGTMed this PR. It's more  aquestion of a longer term decision about how we want to present data to the user. Do we privilege being most up-to-date, or causing least possible confusion when people are switching tools?

itowlson:
That's a good question.  I have to fetch the index because, as far as I could tell, there is no way to list the charts in a repo using the Helm CLI (when I asked, @bacongobbler pointed me at `index.yaml`, but that might have been because I phrased the question badly).  So I am not sure if Helm itself gives me a way to get a view that's consistent with commands that use the cached index.  If the structure of the cache is documented and contractual, then I guess I could pull it from there, but otherwise I don't want to rely on implementation details...

bjornmagnusson:
There is no cli command for it, but one can use `helm search` as a workaround. E.g. to list charts in the `stable` repo one can use `helm search stable` or perhaps `helm search stable/`. It´s not beautiful, but it does work.

itowlson:
Ah, thanks @bjornmagnusson.  `helm search stable` works, and is presumably in sync with the cache, so perhaps we should use that...  cc @technosophos

itowlson:
Updated to use `helm search` - please re-review.

src/helm.repoExplorer.ts
```diff
@@ -160,31 +160,19 @@ async function listHelmRepos(): Promise<Errorable<HelmRepo[]>> {
...
-                const charts = await listHelmRepoChartsCore(repoName, indexUrl);
-                return { succeeded: true, result: charts };
-            } catch (e2) {
-                return { succeeded: false, error: [ "" + e2 ] };
-            } finally {
-                d.dispose();
-            }
-        }
-        return { succeeded: false, error: [ "" + e ] };
+    const sr = await helm.helmExecAsync(`search ${repoName} -l`);
```
technosophos: I would recommend adding a `/` to the end of repoName. Otherwise, it is possible that the regexp matcher make return a greedier result set than you might expect.

For example, say I had a repo named `foo` and one named `foobar`. `helm search foo` would return results for both repos, while `helm search foo/` will only return those for repo `foo`.

PR #322, opened 2018-07-26
add support for helm package

mostafahussein:
This PR adds the ability to package helm chart through vscode

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
         "onCommand:extension.helmDepUp",
         "onCommand:extension.helmInspectValues",
         "onCommand:extension.helmGet",
-	    "onCommand:extension.helmPackage",
+	"onCommand:extension.helmPackage",
         "onCommand:extension.draftVersion",
         "onCommand:extension.draftCreate",
         "onCommand:extension.draftUp",
@@ -744,11 +744,11 @@
                 "category": "Helm"
             },
             {
-		        "command": "extension.helmPackage",
-		        "title": "Package",
-		        "description": "Package a chart directory into a versioned chart archive file.",
-		        "category": "Helm"
-	        }
+		"command": "extension.helmPackage",
+		"title": "Package",
+		"description": "Package a chart directory into a versioned chart archive file.",
+		"category": "Helm"
+	    }
         ],
         "languages": [
             {
--- a/package.json
+++ b/package.json
@@ -741,7 +741,13 @@
                 "title": "Get Release",
                 "description": "Get a Helm release from the cluster",
                 "category": "Helm"
-            }
+            },
+            {
+				"command": "extension.helmPackage",
+				"title": "Package",
+				"description": "Run the Helm package on this chart.",
+				"category": "Helm"
+			}
         ],
         "languages": [
             {
--- a/package.json
+++ b/package.json
@@ -745,7 +745,7 @@
             {
 				"command": "extension.helmPackage",
 				"title": "Package",
-				"description": "Run the Helm package on this chart.",
+				"description": "Package a chart directory into a versioned chart archive file.",
 				"category": "Helm"
 			}
         ],
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
         "onCommand:extension.helmDepUp",
         "onCommand:extension.helmInspectValues",
         "onCommand:extension.helmGet",
+	"onCommand:extension.helmPackage",
         "onCommand:extension.draftVersion",
         "onCommand:extension.draftCreate",
         "onCommand:extension.draftUp",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -741,7 +741,13 @@
                 "title": "Get Release",
                 "description": "Get a Helm release from the cluster",
                 "category": "Helm"
-            }
+            },
+            {
+				"command": "extension.helmPackage",
+				"title": "Package",
+				"description": "Run the Helm package on this chart.",
```
itowlson: Suggest following the Helm CLI text, e.g. "Package the chart into a versioned chart archive file"

src/helm.exec.ts
```diff
@@ -173,6 +173,26 @@ export function helmGet(resourceNode: explorer.ResourceNode) {
     });
 }
 
+// helmPackage runs the Helm package on a chart within your project.
+export function helmPackage() {
+    pickChart((path) => {
+        vscode.window.showInputBox({
```
itowlson: This might be better done with `showOpenDialog`?  Then the user can enter a path manually or select one using the GUI.

src/helm.exec.ts
```diff
@@ -173,6 +173,26 @@ export function helmGet(resourceNode: explorer.ResourceNode) {
...
+            prompt: "package path",
+            placeHolder: "package path"
+        }).then((ppath) => {
+            const packagepath = ppath;
```
itowlson: I think you need to check `if (ppath)` - if the user cancels `showInputBox` then the continuation is still called but with an `undefined` argument.

itowlson: Not sure of the value of capturing `ppath` into a separate constant - could we not use it directly?  (In which case we would want to use the full name `packagePath`.)

itowlson: Prefer camelCasing - `packagePath`.

package.json
```diff
@@ -741,7 +741,13 @@
                 "title": "Get Release",
                 "description": "Get a Helm release from the cluster",
                 "category": "Helm"
-            }
+            },
+            {
+				"command": "extension.helmPackage",
```
itowlson: Probably a good idea to add this to the activation events too (lines 25-67).

itowlson: I wonder if we should also add this to the file explorer context menu (a la Inspect Values) but I'm not sure whether there is a good criterion for which nodes to show it on -- happy to defer.

src/helm.exec.ts
mostafahussein: For me as a user opening a dialog would be a better option!
After taking a quick look about `showOpenDialog` and after testing it, it supports only selection through GUI. Will it be enough for you as a user or you need the ability to enter the path manually ? Also will we have to use both `showOpenDialog` and `showInputBox` in this case or this is a better way to do it ?

src/helm.exec.ts
mostafahussein: `packagePath` is needed to define a custom path where you want to save the chart packages.

src/helm.exec.ts
itowlson: I'm sorry, I explained myself badly.  As you rightly say, `showOpenDialog` always shows a GUI.  But if the user really wants to type or paste a path then they can do that in the text box on the dialog (at least on Windows).  So although it is not quite as seamless as `showInputBox` in this case, it still suffices, and is easier for what I think is the common case where the user wants to browse.  So no, we don't need to also use `showInputBox` - `showOpenDialog` should serve both audiences.  Hope that makes more sense now!

src/helm.exec.ts
itowlson: I understand that, I was trying to say that I don't think we need separate `ppath` and `packagePath` variables.  Can we not just write `....then((packagePath) => { … });`?

mostafahussein:
@itowlson I have pushed the changes, please review it again!

package.json
mostafahussein: Can you explain more please ?

package.json
```diff
@@ -62,6 +62,7 @@
         "onCommand:extension.helmDepUp",
         "onCommand:extension.helmInspectValues",
         "onCommand:extension.helmGet",
+	    "onCommand:extension.helmPackage",
```
itowlson: Can you check the indentation on this please?  It looks a bit off but I'm not sure if that's GitHub preview.

mostafahussein: it seems to be an issue with Github preview as the file is intended as excepted as you can see in [here](https://github.com/mostafahussein/vscode-kubernetes-tools/blob/feature-1/package.json#L65) however it looks different in this preview. Also I tried to use the github editor itself online but still looks different in the preview

mostafahussein: Also through VSCode itself looks indented correclty

itowlson: No worries, I wonder if it's some tabs vs spaces nonsense, but let's not get hung up on it; thanks for taking a look.  I'll get the PR merged and if it doesn't look right to some people then we can sort it in one of our formatting passes.  Many thanks for the contribution!

itowlson: For what it's worth, it was indeed mixed tabs/spaces (which is why it showed up off in the GH preview but looked fine in VS Code).  Normally the linter flags this stuff but I guess it doesn't know how to check `package.json`.  Anyway all sorted now - thanks for your patience!

PR #325, opened 2018-07-30
Helm repo explorer context menu

### LINKED ISSUE #323: Feature: Context menu for chart explorer
With the new chart repository explorer view, here are a few things that I think would be useful to have in a context menu hanging off of a chart version:

- `install`: Install that chart
- `dependencies`: Run `helm dep list` on the chart
- `inspect`: Run a `helm inspect` (possibly could be a submenu with `helm inspect values` and `helm inspect chart`. NB: Inspecting _values_ is a fairly common task for people during dev. Inspecting chart YAML is less common
- `fetch`: Do a `helm fetch` (into the local project?) and then unpack the chart for editing. I'm thinking of this as a way to start with an existing chart and start writing/forking your own.

itowlson:
This provides a context menu for charts and chart versions in the new Helm repo explorer tree view.  Commands implemented so far:

* `Install`
* `Fetch`
* `Inspect Chart`
* `Inspect Values`
* `Show Dependencies`

~~The original issue also asks for `Dependencies` but there's a question mark over how to implement that - will push an update and remove the WIP tag when that's resolved (or parked).~~

Fixes #323.

### DIFF
--- a/src/helm.documentProvider.ts
+++ b/src/helm.documentProvider.ts
@@ -7,6 +7,7 @@ import { escape as htmlEscape } from 'lodash';
 
 import * as helm from './helm';
 import * as logger from './logger';
+import { failed } from './errorable';
 
 interface HelmDocumentResult {
     readonly title: string;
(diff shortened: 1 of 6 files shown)


itowlson:
Question mark over Dependencies command is now removed - have at it.

technosophos:
This is so awesome. I have been manually testing it this morning. 

Now that you are scanning the local cache, the list comes up very quickly. Refreshes are fast enough that I suspected (until I went and looked at the code) that it was not working.

On the context menu, it's very much inline with what I was hoping for.

- Fetch does exactly what I would expect. It might be nice to add a follow-on "fetch into..." feature that would allow me to specify where to put it. (Not sure what the idiom is for that in VS Code)
- It was really convenient to be able to take a quick look at the chart with the `inspect` options.
- One thing confused me about the `dependencies` frame:

screen shot 2018-08-03 at 09 01 34

What does "missing" mean in this case? The dependencies are typically bundled into the chart, so if the dependency is "missing" from inside that chart, it's a major bug. But if "missing" merely means not already installed in the cluster, that is not usually important for a Helm chart. It might make sense to just remove that text.

There are some interesting tidbits of information inside of a `requirements.yaml` file that we could surface in future versions (like whether or not a dependency can be turned off by using a condition). I'll file a separate issue for that, though, since it's outside the scope of this PR.

src/helm.documentProvider.ts
```diff
@@ -134,4 +144,33 @@ export class HelmTemplatePreviewDocumentProvider implements vscode.TextDocumentC
...
+        const version = uri.query;
+        const dependencies = await exec.helmDependenciesCore(chartId, version);
+        if (failed(dependencies)) {
+            return `<p>${dependencies.error[0]}</p>`;
+        }
+
+        const list =  dependencies.result
+                                  .map(this.formatDependency)
+                                  .join('<br>');
+        return `<p>${chartId} depends on:</p><ul>${list}</ul>`;
```
technosophos: I am guessing, based on what I see as output, that this is mis-formatting text:

```
$ helm dep list ./wordpress
NAME   	VERSION	REPOSITORY                                       	STATUS
mariadb	4.x.x  	https://kubernetes-charts.storage.googleapis.com/	unpacked
```

technosophos:
Other than the missing status field (or maybe I'm still misunderstanding something), this all looks great to me, and I am really excited to see this viewer make its way in.

PR #327, opened 2018-07-31
Add Delete now to Pod context menu

### LINKED ISSUE #326: Add `Delete now` entry to Pod context menu
Please add a `Delete now` entry to Pod context menu, here &mdash;

image

Sometimes a pod isn't easily convinced to exit or the grace period is set too high, this comes in handy to convince the pod to give up in an instant (`kubectl delete pod pod_name --now` has grace period = 1).

ubikusss:
Resolves #326 

I see some bugs in the diffs but should be fine. 

Addressed the problem mentioned in the issue.


src/extension.ts
```diff
@@ -1298,6 +1299,29 @@ const deleteKubernetes = async (explorerNode?: explorer.ResourceNode) => {
     }
 };
 
+const deleteKubernetesNow = async (explorerNode?: explorer.ResourceNode) => {
```
bnookala: Thanks for the contribution! I noted that this code is almost the same as the function `deleteKubernetes`, with the exception of the kubectl invocation adding `--now`. We should be able to simplify this so as to avoid duplicate code. Take a look at lines 166, and 167 of `extension.ts`, where we create an anonymous function, and call `getEvents`, passing it kubectl, display mode, and the explorer node.

In a similar fashion, you can modify the `deleteKubernetes` function to accept a `DeleteMode` enum (containing `Now`, and `Graceful`), adding the `--now` flag to the kubectl invocation if the if deletion mode is `DeleteMode.Now`.

src/extension.ts
```diff
@@ -130,6 +130,7 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
             maybeRunKubernetesCommandForActiveWindow.bind(this, 'create', "Kubernetes Creating...")
         ),
         registerCommand('extension.vsKubernetesDelete', deleteKubernetes),
+        registerCommand('extension.vsKubernetesDeleteNow', deleteKubernetesNow),
```
itowlson: Our naming convention for these functions is a bit weird (and we should probably make it more natural) - we typically use the command name followed by `Kubernetes` - so I'd suggest `deleteNowKubernetes`.

package.json
```diff
@@ -334,20 +335,25 @@
                     "when": "view == extension.vsKubernetesExplorer && viewItem == vsKubernetes.resource.pod"
                 },
                 {
-                    "command": "extension.vsKubernetesTerminal",
+                    "command": "extension.vsKubernetesDeleteNow",
```
itowlson: I'm a bit wary of having both `Delete` and `Delete Now` in the context menu.  It feels like we're starting to overload the menu with possible options for one particular command.  I have a couple of alternative suggestions:

1. Have `Delete Now` in the palette only, so it takes more effort to select as befits a more destructive command.

2. As we already display a confirmation for the `Delete` command, add a "now" option to that confirmation.  E.g. 'pod/foo-123456 will be deleted. [Delete Gracefully] [Delete Now] [Cancel]'.  But perhaps that makes it too easy to click Delete Now when you meant Delete Gracefully, I'm not sure.

package-lock.json
```diff
@@ -9,8 +9,8 @@
             "resolved": "https://registry.npmjs.org/@types/bunyan/-/bunyan-1.8.4.tgz",
             "integrity": "sha1-acEa3HtQU41F+2jZrjnQYrlDLzg=",
             "requires": {
-                "@types/events": "1.1.0",
-                "@types/node": "6.0.96"
+                "@types/events": "*",
```
itowlson: I'm not sure why this PR includes changes to `package-lock.json`.  We're not changing any dependencies or version info in `package.json` so we shouldn't be revving versions in the lock file.  Can you back these out please?

package.json
itowlson: Is there a reason to offer Delete Now only for pods and not for other resources?

src/extension.ts
ubikusss: Thank you! Will do!

ubikusss:
Summary for reviews: 
- Made the change for duplicated code
- package-lock.json - backed, but there are still conflicts since it changed on master.
- Naming convention: kept it to be in sync with the other commands. 
- UI change: I think it's worth it :)
- Delete now only for pods: not sure if it makes sense for other resources.

package.json
snobu: As the issue owner, let me try to add some value :)

The `--now` flag seems to only affect Pods, it does not propagate to Deployments/Services (according to https://github.com/kubernetes/kubernetes/issues/25055).

snobu: I'm fine with `2.)`, however, that overloaded confirmation window will now be surprisingly different to unsuspecting users, which may or may not create even greater confusion. I'd say keep it simple(r)?

src/extension.ts
```diff
@@ -1275,14 +1280,20 @@ async function reportDeleteResult(resourceId: string, shellResult: ShellResult)
     refreshExplorer();
 }
 
-const deleteKubernetes = async (explorerNode?: explorer.ResourceNode) => {
+async function deleteKubernetes(delMode: KubernetesDeleteMode, explorerNode?: explorer.ResourceNode) {
+
+    let delModeString: string = '';
```
itowlson: Minor: consider using a ternary expression or helper function so that `delModeString` can be `const`.

itowlson: Minor: suggest `deleteModeArg` rather than `String` to clarify that this is an argument going into `kubectl` (`String` just tells us its type).

itowlson: Minor: I don't think we need the type annotation here - `string` will be inferred I believe.

Issue #305, opened 2018-07-13
Add ability to customize minikube start command


detiber:
The minikube start command is currently hardcoded, being able to override it would help in cases where VirtualBox is not installed or users would prefer to customize the minikube installation (such as using an alternative container runtime).

itowlson:
cc @brendandburns

detiber:
I found that this limitation can be worked around using the minikube config, running `minikube config set vm-driver kvm2` allows this extension to be able to create a minikube cluster.

itowlson:
Fixed by #330.

PR #330, opened 2018-07-31
Add an option for VM driver as well as additional flags.

### LINKED ISSUE #314: Adding possibility to use KVM driver when starting Minikube
On my Fedora machine, I prefer to use `minukube` with the KVM driver (with `--vm-driver=kvm` option) and not using VirtualBox.
This extension doesn't provide the possibility to chose a different vm driver.
It could be really useful having that. Maybe a configuration step when Minikube is selected in order to specify VirtualBox or the KVM driver (so that the command shell will be `minukube start --vm-driver=kvm`).

brendandburns:
Fixes #314 #305 

@itowlson

### DIFF
--- a/src/components/clusterprovider/minikube/minikubeclusterprovider.ts
+++ b/src/components/clusterprovider/minikube/minikubeclusterprovider.ts
@@ -146,7 +146,7 @@ async function promptForConfiguration(previousData: any, context: Context, actio
         </tr>
         <tr>
         <td>Additional Flags:</td>
-        <td style='text-align: right'><input name='addtionalflags' type='text' value='' /></td>
+        <td style='text-align: right'><input name='additionalflags' type='text' value='' /></td>
         </tr>
         </table>
         `
@@ -177,7 +177,7 @@ async function createCluster(previousData: any, context: Context): Promise<strin
 
     context.minikube.start({
         vmDriver: previousData.vmdriver,
-        additionalFlags: previousData.addtionalflags
+        additionalFlags: previousData.additionalflags
     } as MinikubeOptions);
 
     const title = createResult.result.succeeded ? 'Cluster creation has started' : `Error ${createResult.actionDescription}`;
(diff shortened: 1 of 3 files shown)


src/components/clusterprovider/minikube/minikubeclusterprovider.ts
```diff
@@ -130,7 +130,25 @@ async function promptForConfiguration(previousData: any, context: Context, actio
...
+           <option>xhyve</option>
+           <option>hyperv</option>
+           <option>hyperkit</option>
+           <option>kvm2</option>
+           <option>none</option>
+        </select></td>
+        </tr>
+        <tr>
+        <td>Additional Flags:</td>
+        <td style='text-align: right'><input name='addtionalflags' type='text' value='' /></td>
```
itowlson: typo: `addtional` -> `additional`

src/components/clusterprovider/minikube/minikubeclusterprovider.ts
```diff
@@ -157,7 +175,10 @@ async function createCluster(previousData: any, context: Context): Promise<strin
         result: runnable
     };
 
-    context.minikube.start();
+    context.minikube.start({
+        vmDriver: previousData.vmdriver,
+        additionalFlags: previousData.addtionalflags
```
itowlson: typo: `addtional` -> `additional`

itowlson:
Had a chat with @brendandburns and it looks like the desired behaviour is as follows:

* Add Existing Cluster - not relevant because if you have a Minikube cluster then it's automatically added to your kubeconfig.
* Create Cluster - gathers options and spins up a Minikube cluster, which is automatically added to kubeconfig.  If Minikube is already installed, then this should do nothing (unless you can define multiple Minikube 'clusters' e.g. with different configs).
* Start Minikube - runs Minikube with the already-gathered options.  This is what it does in the absence of options, so this is doing the right thing.

So we might need to remove Add Existing Cluster and tweak Create Cluster for the case where Minikube is already set up.  But otherwise what we have fits this model well.

itowlson:
@brendandburns Seemed a bit pointless bugging you to fix the typo so I used my maintainer powers on you - hope that was okay.

Issue #336, opened 2018-08-10
Support multiple kubeconfigs and switching between them


itowlson:
Some users have multiple kubeconfigs and want to be able to switch quickly and frequently between them.  We support having an override kubeconfig in the settings file (and this allows kubeconfig-per-project) but it would be good to allow users to change kubeconfigs more dynamically.

itowlson:
Fixed via #337.

PR #337, opened 2018-08-10
Support rapid switching between multiple kubeconfigs

itowlson:
Some users have multiple Kubernetes environments which they keep in separate kubeconfig files.  For working with relatively static kubeconfigs (e.g. single alternate kubeconfig, kubeconfig per project, etc.) we have the `vs-kubernetes.kubeconfig` setting which can be applied at global or workspace level.  But if the user wants to be able to swap more dynamically then they have to keep changing that setting and that is a bit niggly.

This PR adds a `Set Kubeconfig` command.  This command displays a list of 'known kubeconfigs' from which the user can choose.  The user can also choose to add a new known kubeconfig, which immediately becomes selected and is added to the list of known kubeconfigs.

In addition, the command supports passing a kubeconfig file path as an argument.  This means that a user can set up keybindings for their favourite kubeconfig files, e.g. in `keybindings.json` write:

```json
    {
        "key": "ctrl+shift+alt+k",
        "command": "extension.vsKubernetesUseKubeconfig",
        "args": "d:\\temp\\kubectl\\my.other.kubeconfig"
    }
```

The user can then instantly (well... fairly instantly) switch to that kubeconfig using their chosen key combo.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -67,6 +67,7 @@ import { DraftConfigurationProvider } from './draft/draftConfigurationProvider';
 import { installHelm, installDraft, installKubectl, installMinikube } from './components/installer/installer';
 import { KubernetesResourceVirtualFileSystemProvider, K8S_RESOURCE_SCHEME, KUBECTL_RESOURCE_AUTHORITY } from './kuberesources.virtualfs';
 import { Container, isKubernetesResource, KubernetesCollection, Pod, KubernetesResource } from './kuberesources.objectmodel';
+import { setActiveKubeconfig, getKnownKubeconfigs, addKnownKubeconfig } from './components/config/config';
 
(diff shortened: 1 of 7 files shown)


itowlson:
Fixes #336.

src/extension.ts
```diff
@@ -1671,6 +1673,50 @@ async function createClusterKubernetes() {
     vscode.commands.executeCommand('vscode.previewHtml', createCluster.operationUri(newId), 2, "Create Kubernetes Cluster");
 }
 
+const KNOWN_KUBECONFIGS_KEY = "vs-kubernetes.knownKubeconfigs";
```
bnookala: let's toss this config management stuff into components/config/config.ts

PR #340, opened 2018-08-13
Add persistent volumes, PV claims and storage classes to tree view

itowlson:
First cut of an update for storage-related resource types - it just adds them to the tree view with the same set of verbs as other resources.  Should there be other/different operations, and are any other resource types required?

Fixes #338.

PR #342, opened 2018-08-13
Support navigation on cluster resources

itowlson:
~~VS Code provides a Go To Definition command which finds the definition of the thing under the cursor and opens it in a new editor.  This PR implements Go To Definition for (certain) cluster resources - for example, you can navigate from a persistent volume to an associated persistent volume claim, or from a deployment that uses a config map to the config map itself.~~

VS Code provides a document linking feature, where you can click a hyperlink in a document to be taken to the target of the hyperlink.  This PR hyperlinks resource references within (certain) cluster resources - for example, you can navigate from a persistent volume to an associated persistent volume claim, or from a deployment that uses a config map to the config map itself.

NOTE: Because one of the motivations for this was navigating between storage resources, the PR includes the changesets for showing the storage resources in the tree (#340).  Once that is merged I'll rebase this PR to include only the Go To Definition bits.  In the meantime, **please ignore the diffs to `kuberesources.ts` and `explorer.ts` as they are covered by the earlier PR.**


itowlson:
Looking further at the API, perhaps this is a better fit for a document link than for Go To Definition...

itowlson:
I changed the design/implementation to use VS Code document links instead of the Go To Definition feature, because this seemed better aligned to the way it works.  If people prefer the Go To Definition UI, I can change it back!

src/kuberesources.linkprovider.ts
```diff
@@ -0,0 +1,167 @@
...
+import * as yl from './yaml-support/yaml-locator';
+
+export class KubernetesResourceLinkProvider implements vscode.DocumentLinkProvider {
+    provideDocumentLinks(document: vscode.TextDocument, token: vscode.CancellationToken): vscode.ProviderResult<vscode.DocumentLink[]> {
+        const sourceKind = k8sKind(document);
+        const yaml = yl.yamlLocator.getYamlDocuments(document);
+        const leaves: yl.YamlNode[] = getLeafNodes(yaml);
+        const links = leaves.map((l) => getLink(document, sourceKind, l))
+                            .filter((l) => !!l);
+        links.forEach((l) => { console.log(l.target.toString()); });
```
bnookala: extraneous logging?

PR #346, opened 2018-08-15
Warn if a pod or deployment does not specify CPU or memory limits

### LINKED ISSUE #344: Check for CPU/memory requests/limits for workload specs
A common way for workloads to be killed is to consume too much CPU or memory. Ideally k8s should provide sane defaults, but doesn't (yet). 

To get around this, we can assist users by alerting them that they don't have any requests/limits set on workloads.

itowlson:
We're finding that a common mistake that people make with Kubernetes is to deploy workloads with no limits on memory or CPU consumption.  This PR displays green wigglies and a 'Problem' indicator in VS Code when the user opens a YAML file for a pod or deployment which does not limit memory and CPU in this way.

This is an initial posting for review and feedback - I don't think things like wording are final.  We should probably sanity check things like the CPU limit as well...

Fixes #344.

### DIFF
--- a/src/components/lint/resourcelimits.ts
+++ b/src/components/lint/resourcelimits.ts
@@ -1,0 +1,117 @@
+import * as vscode from 'vscode';
+import * as yaml from 'js-yaml';
+
+import { Linter } from './linters';
+import { JsonALikeYamlDocumentSymbolProvider } from '../../yaml-support/jsonalike-symbol-provider';
+
+const jsonalikeYamlSymboliser = new JsonALikeYamlDocumentSymbolProvider();
+
+interface Syntax {
+    load(text: string): any;
+    symbolise(document: vscode.TextDocument): Promise<vscode.SymbolInformation[]>;
+}
+
+const jsonSyntax = {
+    load(text: string) { return JSON.parse(text); },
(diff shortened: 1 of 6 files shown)


src/components/lint/resourcelimits.ts
```diff
@@ -0,0 +1,61 @@
...
+        const warnings: vscode.Diagnostic[] = [];
+        const warnOn = (symbol: vscode.SymbolInformation, text: string) => {
+            warnings.push(warningOn(symbol, text));
+        };
+
+        for (const containersSymbol of containersSymbols) {
+            const imagesSymbols = childSymbols(symbols, containersSymbol, 'image');
+            const resourcesSymbols = childSymbols(symbols, containersSymbol, 'resources');
+            if (resourcesSymbols.length < imagesSymbols.length) {
+                warnOn(containersSymbol, 'One or more containers do not have resource limits - this could starve other processes');
```
bnookala: Is it possible for this text to render links? Might be useful to link to k8s documentation about resource starvation.

itowlson: Doesn't look like it - the type of the message is `text` and VS Code doesn't appear to autolink `http://` substrings.

PR #349, opened 2018-08-18
Adding insecure skip tls-verification

### LINKED ISSUE #303: Add "--insecure-skip-tls-verify" setting
With my current certificate setup, when I attempt to get any data from my cluster, I get the following error:

`Unable to connect to the server: x509: certificate is valid for <certificate_ip>, not <actual_ip>`

The `--insecure-skip-tls-verify=true` flag allows ignoring insecure certificates in `kubectl`. An equivalent setting (like `"vs-kubernetes.insecure-skip-tls-verify": false` as default) could be added to the extension to enable this configuration.

bnookala:
Fixes #303

### DIFF
--- a/src/kubectl.ts
+++ b/src/kubectl.ts
@@ -158,8 +158,9 @@ async function invokeWithProgress(context: Context, command: string, progressMes
 
 async function invokeAsync(context: Context, command: string, stdin?: string): Promise<ShellResult> {
     if (await checkPresent(context, 'command')) {
+        const kubectlArgs = getKubectlArgs(context);
         const bin = baseKubectlPath(context);
-        const cmd = `${bin} ${command}`;
+        const cmd = `${bin} ${command} ${kubectlArgs}`;
         const sr = await context.shell.exec(cmd, stdin);
         if (sr.code !== 0) {
             checkPossibleIncompatibility(context);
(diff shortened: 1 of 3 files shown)


src/kubectl.ts
```diff
@@ -272,3 +279,8 @@ function path(context: Context): string {
     const bin = baseKubectlPath(context);
     return binutil.execPath(context.shell, bin);
 }
+
+function getKubectlArgs(context: Context): string {
+    const skipTlsVerification: boolean = context.host.getConfiguration('vs-kubernetes')[`vs-kubernetes.skip-tls-verification`];
+    return `--insecure-skip-tls-verify=${skipTlsVerification}`;
```
itowlson: I'm wondering if rather than making this a one-off we should support custom flags, so that users could also inject certificates.  I guess the problem with that is that if we later add integrated support that could emit a particular flag then we'd have a problem where it could conflict with a custom flag.

itowlson: We often display our generated `kubectl` commands (e.g. when running in a terminal) so consider returning the empty string when the setting is false (to reduce noise in the displayed command).

itowlson:
Suggest we close this in favour of the documentation-based approach - are you okay with that?

PR #351, opened 2018-08-25
Add support for querying minikube status.

brendandburns:


src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -109,6 +133,41 @@ async function stopMinikube(context: Context): Promise<void> {
...
+                vscode.window.showInformationMessage(`Minikube is ${obj[0]}`);
+            }
+            return {
+                running: 'Stopped' !== obj[0],
+                cluster: obj[1],
+                kubectl: obj[2],
+            } as MinikubeInfo;
+        }
+        if (notification) {
+            vscode.window.showErrorMessage(`failed to get status: ${result.stderr}`);
```
itowlson: It would be good to clarify that this message refers to Minikube.  Also, I think we usually start messages with a capital letter.  E.g. `Failed to get Minikube status: ${…}` maybe?

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -109,6 +133,41 @@ async function stopMinikube(context: Context): Promise<void> {
...
+        } else {
+            throw new Error(`failed to get status: ${result.stderr}`);
+        }
+    } catch (err) {
+        if (notification) {
+            vscode.window.showErrorMessage(`minikube status failed: ${err}`);
```
itowlson: Again start with capital letter.  How does `Minikube status failed` differ from `Failed to get status` on line 159?  Seems like we should use a consistent form of wording, unless we are intentionally keeping them different e.g. for diagnostic purposes.

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -109,6 +133,41 @@ async function stopMinikube(context: Context): Promise<void> {
...
+        }
+        throw(err);
```
itowlson: `err` is rethrown even if a message has been displayed.  Won't that result in the message being displayed twice if this bubbles back to the top level?

itowlson: I feel like the intermingling of 'UI mode' and 'silent mode' here is causing duplication and potential bugs.  Something to consider is having the function return a `Promise<Errorable<MinikubeInfo>>` instead - then the consumer of the function could check for an error and display it or swallow it according to its posture (and similarly in the success case could display UI or consume the info programmatically).  This simplifies this function and keeps the UI in a UI-oriented caller.  If you want the UI to be reusable then you could still have a `displayMinikubeStatus` function which called this 'silent' one.

brendandburns:
Ripped out the 'notification' stuff as you suggested...

PR #352, opened 2018-08-28
Update vulnerable packages

itowlson:
Well, the Yellow Bar of Death is back.  To be honest, I kinda feel the risk of upgrading working packages is actually a lot greater than the risk that something four levels down from our code might hang VS Code by passing a rotten regex into something five levels down, but you can't argue with the Yellow Bar of Death.  Unless it keeps on nagging me with pointless vulnerabilities.  Then I might well argue with it.

Sigh.


brendandburns:
needs rebase, but otherwise LGTM.

Issue #356, opened 2018-09-03
Multi-platform dependency executable paths?


pentago:
I as a user should have the possibility of doing work across a range of platforms and defining dependency paths myself or automatically by extension. 

Reason for this is that many, including myself, users are using editor settings sync extension to keep our config consistent across machines we use.

 Currently, I use MacOS and Windows both at home and at the office and installing deps each time i pull editor settings is a bit of a pain. It would be great if there are platform-dependant settings provided like these below:

```
    "vs-kubernetes": {

        // Windows
        "vs-kubernetes.windows.helm-path": "some/path/helm.exe",
        "vs-kubernetes.windows.draft-path": "some/path/draft.exe",
        "vs-kubernetes.windows.minikube-path": "some/path/draft.exe",

        // Linux
        "vs-kubernetes.linux.helm-path": "some/path/helm",
        "vs-kubernetes.linux.draft-path": "some/path/draft",
        "vs-kubernetes.linux.minikube-path": "some/path/draft",

        // MacOS
        "vs-kubernetes.macos.helm-path": "some/path/helm",
        "vs-kubernetes.macos.draft-path": "some/path/draft",
        "vs-kubernetes.macos.minikube-path": "some/path/helm",
    }
```

Is this behavior possible or there are some limitations preventing existence of this?

itowlson:
This is a great point.  We should totally do this.  We would still maintain the existing config settings for backward compatibility and to make the common case simpler, but we could indeed allow per-platform overrides for your scenario.  Thanks!

pentago:
Just noticed similar logic on new Code release:
https://i.imgur.com/VKAOrCL.png

itowlson:
Interesting.  Unfortunately I don't think we can follow that pattern for configuration entries, so I'm planning to stick with your suggestion of extra config entries with a naming convention.

PR #357, opened 2018-09-03
Refactoring: encapsulate configuration access

itowlson:
We currently have a _lot_ of places in the code that read or write the extension configuration.  In preparation for addressing #356, it will be useful to encapsulate this access (the issue is concerned only with tool paths, but we may as well clean house while we're in there); this will allow us to inject the per-platform logic at the configuration layer rather than up at the application layer, where it will clutter the application logic and be duplicated around the various tool wrappers.


brendandburns:
LGTM.

PR #361, opened 2018-09-10
Proposed extension API framework

itowlson:
This is a stab at a framework for an extension API which meets our requirements around compatibility, maintainability and discoverability.  It is for review only at this stage.  See the wiki for some of the thinking and the example in the read-me for an example of usage.

The principal changes to previous stabs at an API are:

* No public surfacing of internal constructs.  All API access is through immutable, versioned interfaces defined _solely for the purpose of the API_.  Therefore, we can change our internal structures and all we need to do (all, he said airily) is map the API interface onto the new internals.
* Negotiated versioning.  A consumer asks for a version of an API, and receives that version or an error. This contrasts with the previous attempt where we simply reported which version of the API we were serving that day.
* Per-component APIs.  Instead of asking for 'version 1', the consumer asks for 'the cluster provider API version 1' or 'the tree API version 1.'  This allows us to rev individual APIs without having to duplicate all the other APIs and mappings that _haven't_ changed.
* Designed for packaging.  The API is now built around a TypeScript declarations file which (as I understand it) can be packaged and published on NPM, so extension authors using TypeScript can consume the API in a strong-typed way.  I am not sure I'm doing this the right way -- it feels a bit awkward -- but using the TypeScript compiler's "generate declarations" option spews out declaration files for _every damn export in the extension_.  Which is far worse.

I'm posting this for feedback and suggestions.  When we've reached a design we're happy with, I'll add documentation, do some proper testing, and take it forward to publication.


itowlson:
There's still some validity to the extension side of this, but the consumer side doesn't feel comfortable, and the interface bits feel distinctly like they're taking us in the wrong direction.  And it's very out of date.  Closing in favour of a different approach.

PR #371, opened 2018-09-21
Switch to the official kubernetes client library.

brendandburns:
This enables `explain` to work with AAD+AKS and GKE clusters...

This also opens the door for using the library for more features like streaming logs and port forwarding, instead of `kubectl`


src/explainer.ts
```diff
@@ -1,30 +1,37 @@
...
-        const kapi = k8sImpl.api(apiCredentials(kc));
-        fixUrl(kapi);
-        kapi.get('swagger.json', (err, data) => {
-            if (err) {
-                reject(err);
-            } else {
-                resolve(data);
+        request(uri, opts, (error, response, body) => {
+            if (error) {
+                reject(error);
```
itowlson: Do we need to `return` from the lambda after these `reject` calls?  I'm not very familiar with explicit promise construction, but these look like after calling `reject` they would fall through to the next statement...?

itowlson: (or of course use `else` to prevent fallthrough)

src/explainer.ts
```diff
@@ -1,30 +1,37 @@
...
+import * as kubernetes from '@kubernetes/client-node';
 import * as pluralize from 'pluralize';
-import * as kubeconfig from './kubeconfig';
 import { formatComplex, formatOne, Typed, formatType } from "./schema-formatting";
 
 export function readSwagger(): Promise<any> {
-    return kubeconfig.readKubectlConfig().then((kc) => readSwaggerCore(kc));
-}
+    const kc = new kubernetes.KubeConfig();
+    kc.loadFromDefault();
```
itowlson: Does this behave correctly if the user has overridden the default `kubeconfig` path in their settings?

itowlson: The previous version didn't, so it's not a regression (and never mattered much because we were only getting the Swagger, not performing cluster operations), but it would be good not to propagate the error.

src/explainer.ts
brendandburns: good point, done!

src/explainer.ts
brendandburns: Fixed.

PR #372, opened 2018-09-21
Add error handling to many promises.

brendandburns:
In the process of working on hover, I noticed that many promises weren't handling errors.

Next step would be to switch to async/await

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -360,7 +360,7 @@ function registerTelemetry(context: vscode.ExtensionContext): vscode.Disposable 
 }
 
 function provideHover(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken, syntax): Promise<vscode.Hover> {
-    return new Promise(async (resolve) => {
+    return new Promise(async (resolve, reject) => {
         if (!explainActive) {
             resolve(null);
             return;
@@ -372,7 +372,8 @@ function provideHover(document: vscode.TextDocument, position: vscode.Position, 
         try {
             obj = syntax.parse(body);
         } catch (err) {
-            // Bad document
+            // Bad document, return nothing
+            // TODO: at least verbose log here?
             resolve(null);
             return;
         }
@@ -398,7 +399,8 @@ function provideHover(document: vscode.TextDocument, position: vscode.Position, 
         }
 
         explain(obj, field).then(
-            (msg: string) => resolve(new vscode.Hover(msg))
+            (msg: string) => resolve(new vscode.Hover(msg)),
+            (err: any) => reject(err)
         );
     });
 
@@ -418,7 +420,6 @@ function provideHoverYaml(document: vscode.TextDocument, position: vscode.Positi
         parse: (text) => yaml.safeLoad(text),
         findParent: (document, parentLine) => findParentYaml(document, parentLine)
     };
-
     return provideHover(document, position, token, syntax);
 }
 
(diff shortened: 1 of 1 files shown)


itowlson:
+1 to async/await all the things

src/kubeconfig.ts
```diff
@@ -2,12 +2,14 @@
...
 import { fs } from './fs';
+import * as vscode from 'vscode';
 import * as yaml from 'js-yaml';
 
 export function readKubectlConfig(): Promise<KubeConfig> {
     return new Promise((resolve, reject) => {
-        const kubeConfig = shell.combinePath(shell.home(), ".kube/config");
+        const kubeConfig = shell.combinePath(shell.home(), '.kube/config');
         fs.readFile(kubeConfig, 'utf8', (err, data) => {
+            console.log(data);
```
itowlson: I don't think we need to add the `console.log` statement, or is it just being moved from elsewhere?

itowlson:
One non-critical change request, otherwise looks good - thanks!

src/kubeconfig.ts
brendandburns: fixed.

brendandburns:
Comment addressed. ptal.

src/extension.ts
```diff
@@ -374,12 +374,14 @@ function provideHover(document, position, token, syntax): Promise<vscode.Hover>
         } catch (err) {
             // Bad document
             resolve(null);
+            vscode.window.showInformationMessage(`Error: ${err}`);
```
itowlson: I am not sure we want to bug the user with these notifications.  These hovers get attached to all JSON and YAML documents: we are not guaranteed that the document the user is looking at is a k8s resource, nor that it is even valid JSON or YAML at this moment (they may be in the middle of editing it).  It's appropriate not to explain in this case (hence the `resolve(null)`, but we shouldn't put a message in the user's face.

src/extension.ts
```diff
@@ -374,12 +374,14 @@ function provideHover(document, position, token, syntax): Promise<vscode.Hover>
...
             return;
         }
 
         // Not a k8s object.
         if (!obj.kind) {
             resolve(null);
+            vscode.window.showInformationMessage('Not a Kubernetes object!');
```
itowlson: Again, this is legit - if explain is on, the hover provider will fire on _all_ JSON and YAML documents.  We don't want to bug the user with "fool, you switched to a non-k8s document while explain was on"; we should just silently resolve as 'no hover tip'.

src/extension.ts
```diff
@@ -468,9 +470,15 @@ async function explain(obj, field) {
             swaggerSpecPromise = explainer.readSwagger();
         }
 
-        swaggerSpecPromise.then((s) => {
-            resolve(explainer.readExplanation(s, ref));
-        });
+        swaggerSpecPromise.then(
+            (s) => {
+                resolve(explainer.readExplanation(s, ref));
+                vscode.window.showErrorMessage('in after swagger spec promise');
```
itowlson: pretty sure this needs to be removed

src/extension.ts
```diff
@@ -468,9 +470,15 @@ async function explain(obj, field) {
...
+            },
+            (err) => {
+                vscode.window.showErrorMessage(`Explain failed: ${err}`);
```
itowlson: This seems like it could fire every time we try to explain an object, which if the promise has failed could result in a lot of noise.  Can we confirm that this fires only once?

src/extension.ts
brendandburns: yep, deleted.

src/extension.ts
brendandburns: Yep, sorry I got over aggressive with my rejecting :)

src/extension.ts
brendandburns: haha, my "debug with prompt" style is revealed...

Removed.

itowlson: Let he who has not left an `alert(`value of foo is ${foo}`);` in a production commit cast the first stone...

src/extension.ts
brendandburns: Better yet, I nulled out the promise so that we'll try again to load it on the error case...

brendandburns:
Comments addressed, thanks for the great feedback. Please take another look.

brendandburns:
@itowlson friendly ping on this one...

Thanks!

brendandburns:
@itowlson fwiw, reporting definitely shouldn't influence code merges. Please only merge when it's ready, regardless of who it's from...

thanks!

PR #377, opened 2018-09-26
Expand the explorer. Add some new types, add some status to Pods.

brendandburns:


src/explorer.ts
```diff
@@ -190,10 +191,12 @@ class KubernetesWorkloadFolder extends KubernetesFolder {
 
     getChildren(kubectl: Kubectl, host: Host): vscode.ProviderResult<KubernetesObject[]> {
         return [
-            new KubernetesDeploymentFolder(),
+            new KubernetesPodSelectorFolder(kuberesources.allKinds.deployment, (kubectl: Kubectl) => kubectlUtils.getPodSelector('deployments', kubectl)),
```
itowlson: This seems like duplication - surely `allKinds.deployment` contains the required string to pass to `kubectlUtils.getPodSelector`, and so we could just bake the callback into `KubernetesPodSelectorFolder`?

src/explorer.ts
```diff
@@ -264,8 +267,25 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
-    getChildren(kubectl: Kubectl, host: Host): vscode.ProviderResult<KubernetesObject[]> {
-        return [];
+    async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        if (this.kind !== kuberesources.allKinds.pod) {
+            return [];
+        }
+        const result = await kubectl.asJson<Pod>(`get pods ${this.metadata.name} -o json`);
+        if (result.succeeded) {
+            const pod = result.result;
+            let ready = 0;
```
itowlson: extract into a `count(array, predicate)` method, or use `pods.status.containerStatuses.filter((cs) => cs.ready).length`

src/explorer.ts
```diff
@@ -264,8 +267,25 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+            pod.status.containerStatuses.forEach((status) => {
+                if (status.ready) {
+                    ready++;
+                }
+            });
+            return [
```
itowlson: I'm a bit doubtful about using child nodes to show properties of a resource - that's not an idiom we've used before, and it feels a bit odd.  A possible alternative would be to show this info on the tooltip, but that's problematic because I think we need to construct the tooltip synchronously.  Or a 'summary status' command perhaps.

src/explorer.ts
```diff
@@ -264,8 +267,25 @@ class KubernetesResource implements KubernetesObject, ResourceNode {
...
+                new DummyObject(`${pod.status.phase} (${ready}/${pod.status.containerStatuses.length})`),
+                new DummyObject(pod.status.podIP),
+            ];
+        }
+        return [ new DummyObject("Error") ];
```
itowlson: Pass `result.error[0]` to the `DummyObject` constructor as the optional `diagnostic` argument - this will surface the error as a tooltip.

src/explorer.ts
```diff
@@ -348,28 +371,17 @@ class KubernetesNamespaceResource extends KubernetesResource {
     }
 }
 
-class KubernetesServiceFolder extends KubernetesResourceFolder {
-    constructor() {
-        super(kuberesources.allKinds.service);
+class KubernetesPodSelectorFolder extends KubernetesResourceFolder {
+    constructor(readonly kind: kuberesources.ResourceKind, readonly fn: (kubectl: Kubectl) => Promise<kubectlUtils.PodSelector[]>) {
```
itowlson: Rename `fn` to something more meaningful - it has a specific purpose not just a generic callback.

src/explorer.ts
```diff
@@ -348,28 +371,17 @@ class KubernetesNamespaceResource extends KubernetesResource {
...
+        super(kind);
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
-        const services = await kubectlUtils.getServices(kubectl);
+        const services = await this.fn(kubectl);
```
itowlson: Is `services` still the correct name here?

src/explorer.ts
itowlson: I think I am also finding the name `PodSelector` confusing.  It implies some strategy or condition for selecting pods, whereas what I think it's used as is something like "a Kubernetes resource that _has_ a pod selector (that is, is associated with a set of pods."  Perhaps something like `PodSelectingResource` might help, though I fear this leads us towards a combinatorial explosion if we later need other traits too.

Would TypeScript intersection types help here, e.g. would modelling things as `KubernetesResource & HasPodSelector` (or `KubernetesResource & SelectsPods`) be clearer and more sustainable?

src/explorer.ts
itowlson: If not, then maybe we need to consider adding it to the 'kind' type, or if that's not appropriate then at least consider baking the `getPodSelector` call into `KubernetesPodSelectorFolder` and only pass in the string... otherwise `KubernetesPodSelectorFolder` feels like a `KubernetesAnythingAnythingerFolder` *grin*

brendandburns: Changed the name (and removed the duplication)

src/explorer.ts
brendandburns: moved to a different PR.

src/explorer.ts
brendandburns: moved to a different PR.

src/explorer.ts
brendandburns: moved to a different PR.

src/explorer.ts
brendandburns: fixed.

src/explorer.ts
brendandburns: fixed.

src/explorer.ts
brendandburns: Renamed, let me know what you think.

brendandburns:
I split this PR up so that we can consider the Pod UX in a different PR.

I think I addressed all of the other comments.

Please take another look.

src/explorer.ts
```diff
@@ -348,25 +351,14 @@ class KubernetesNamespaceResource extends KubernetesResource {
...
+class KubernetesSelectsPodsFolder extends KubernetesResourceFolder {
+    constructor(readonly kind: kuberesources.ResourceKind) {
+        super(kind);
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
-        const deployments = await kubectlUtils.getDeployments(kubectl);
-        return deployments.map((dp) => new KubernetesSelectorResource(this.kind, dp.name, dp, dp.selector));
+        const objects = await kubectlUtils.getResourceWithSelector(this.kind.abbreviation, kubectl);
+        return objects.map((svc) => new KubernetesSelectorResource(this.kind, svc.name, svc, svc.selector));
```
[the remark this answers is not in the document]

src/kuberesources.ts
```diff
@@ -12,18 +12,20 @@ export const allKinds = {
     namespace: new ResourceKind("Namespace", "Namespaces", "Namespace", "namespace"),
     node: new ResourceKind("Node", "Nodes", "Node", "node"),
     deployment: new ResourceKind("Deployment", "Deployments", "Deployment", "deployment"),
+    daemonSet: new ResourceKind("DaemonSet", "DaemonSets", "DaemonSet", "daemonset"),
```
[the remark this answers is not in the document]
brendandburns: ok, I ran through the kubernetes.io docs (e.g. https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) and I standardized on that (which itself is semi inconsistent `Persistent Volumes` but `StatefulSets`) but at least there's a rubric we can follow....

src/kuberesources.ts
```diff
@@ -12,18 +12,20 @@ export const allKinds = {
...
     replicaSet: new ResourceKind("ReplicaSet", "ReplicaSets", "ReplicaSet", "rs"),
     replicationController: new ResourceKind("Replication Controller", "Replication Controllers", "ReplicationController", "rc"),
     job: new ResourceKind("Job", "Jobs", "Job", "job"),
     cronjob: new ResourceKind("Cron Job", "Cron Jobs", "CronJob", "cronjob"),
     pod: new ResourceKind("Pod", "Pods", "Pod", "pod"),
     service: new ResourceKind("Service", "Services", "Service", "service"),
-    configMap: new ResourceKind("ConfigMap", "ConfigMaps", "ConfigMap", "configmap"),
+    configMap: new ResourceKind("ConfigMap", "Config Maps", "ConfigMap", "configmap"),
```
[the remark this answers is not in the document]
brendandburns: fixed.

src/kuberesources.ts
```diff
@@ -12,18 +12,20 @@ export const allKinds = {
...
     secret: new ResourceKind("Secret", "Secrets", "Secret", "secret"),
     ingress: new ResourceKind("Ingress", "Ingress", "Ingress", "ingress"),
     persistentVolume: new ResourceKind("Persistent Volume", "Persistent Volumes", "PersistentVolume", "pv"),
     persistentVolumeClaim: new ResourceKind("Persistent Volume Claim", "Persistent Volume Claims", "PersistentVolumeClaim", "pvc"),
     storageClass: new ResourceKind("Storage Class", "Storage Classes", "StorageClass", "sc"),
+    statefulSet: new ResourceKind("StatefulSet", "Stateful Sets", "StatefulSet", "statefulset"),
```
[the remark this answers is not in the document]
brendandburns: fixed.

brendandburns:
Comments addressed, please take another look.

PR #379, opened 2018-09-27
Unify snippet structure and add new snippets

TheBay0r:
Maybe it is just me, but I think it would be good to follow the same structure in every snippet. Since available `kind` is somewhat dependent on `apiVersion` I decided to switch it in two existing templates (I know yaml doesn't care).

I also added two new snippets for ConfigMap and ReplicationController


itowlson:
Sorry for the delay getting back to you on this - curse of the weekend plus commitments elsewhere!  This sounds good, will take a look very shortly - thanks!

PR #381, opened 2018-10-02
Update README.md

AntoineGa:
Clarify Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This extension assumes you have the following binaries on your `PATH`:
    * `draft` (optional)
    * `minikube` (optional)
 
-For `kubectl`, `helm` and `draft` the binaries need not be on the system PATH, provided you tell the extension their locations using the appropriate `vs-kubernetes -> vs-kubernetes.${tool}-path` configuration setting.  See [Extension Settings](#extension-settings) below.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,7 +17,7 @@ This extension assumes you have the following binaries on your `PATH`:
    * `draft` (optional)
    * `minikube` (optional)
 
-For `kubectl`, `helm` and `draft` the binaries need not be on the system PATH, provided you tell the extension their locations using the appropriate `vs-kubernetes -> vs-kubernetes.${tool}-path` configuration setting.  See [Extension Settings](#extension-settings) below.
+For `kubectl`, `helm` and `draft` the binaries need not to be on the system PATH. You tell to the extension their locations using the appropriate `vs-kubernetes -> vs-kubernetes.${tool}-path` configuration setting.  See [Extension Settings](#extension-settings) below.
```
itowlson: In the first sentence, please remove the added 'to' - should read 'the binaries need not be...'.  Or (perhaps better) change it to 'the binaries do not need to be...'

itowlson: In the second sentence, please remove 'to' - should be 'You tell the extension...'  Also, do you think something like "In this case, you tell the extension..." would be clearer still?  Not sure and would value your feedback.

AntoineGa: i suggest "You can configure the extension by specifying the locations"

AntoineGa: Thank you for your kindness

PR #382, opened 2018-10-03
Per-OS override of tool paths

### LINKED ISSUE #356: Multi-platform dependency executable paths?

itowlson:
Fixes #356.

### DIFF
--- a/src/components/config/config.ts
+++ b/src/components/config/config.ts
@@ -1,5 +1,6 @@
 import * as vscode from 'vscode';
 import { Host } from '../../host';
+import { Shell, Platform } from '../../shell';
 
 const EXTENSION_CONFIG_KEY = "vs-kubernetes";
 const KUBECONFIG_PATH_KEY = "vs-kubernetes.kubeconfig";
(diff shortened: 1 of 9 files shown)


itowlson:
This is the scenario where Code is running under Windows but you want to invoke the WSL version of the tools?  The OS check is for the Code process, so you would use the `.windows` suffix in that case.  For a portable configuration which needed to distinguish between the Code-on-Windows-invoking-Windows-tools and Code-on-Windows-invoking-Linux-tools, we would need to add an extra suffix (and `.wsl` would seem appropriate).  If the portable config _always_ wanted to use the WSL tools on Windows, then the `.windows` suffix might suffice -- there are a bunch of issues but I think they're issues we'd run into anyway as part of getting WSL working.

itowlson:
In terms of 'is there a way to check for WSL on Windows' the answer is no and yes -- you can't run the extension under WSL, so there's no point even trying to check, but if we add WSL support _to this extension_ then of course we can check our own setting!  I hope that makes sense... have explained it a bit confusingly I fear...

src/components/config/config.ts
```diff
@@ -84,14 +85,35 @@ export function getActiveKubeconfig(): string {
 
 // Functions for working with tool paths
 
-export function getToolPath(host: Host, tool: string): string | undefined {
-    return host.getConfiguration(EXTENSION_CONFIG_KEY)[toolPathKey(tool)];
+export function getToolPath(host: Host, shell: Shell, tool: string): string | undefined {
```
bnookala: instead of passing the shell down to here, would it be acceptable to pass the platform directly? or make it a property available on `host`?

itowlson: I'm torn.  On the one hand, yes, all the function _needs_ is the platform, so why pass in a monstrous `Shell`?  On the other hand, _everywhere_ this is called, it has a `Shell` handy but it usually _hasn't_ pulled out the `platform`, so _all_ the call sites would be polluted with a call to `.platform()`.

Not keen on putting it on `Host` as that is mostly wrapping the VS Code environment.  Though the combination of a `Host` and a `Shell` is so common we could wrap that up into a single interface, something like `ExtensionContext` though that's already taken **grin**.

I am probably overthinking this.  _sigh_

PR #386, opened 2018-10-11
Add an upgrade check for minikube, also some cleanups.

brendandburns:
@itowlson

### DIFF
--- a/package.json
+++ b/package.json
@@ -255,6 +255,10 @@
                     "when": "view == extension.vsKubernetesExplorer"
                 },
                 {
+                    "command": "extension.vsKubernetesUseKubeconfig",
+                    "when": "view == extension.vsKubernetesExplorer"
+                },
+                {
                     "command": "extension.vsKubernetesRefreshHelmRepoExplorer",
                     "when": "view == extension.vsKubernetesHelmRepoExplorer",
                     "group": "navigation"
--- a/package.json
+++ b/package.json
@@ -248,15 +248,18 @@
                 },
                 {
                     "command": "extension.vsKubernetesCreateCluster",
-                    "when": "view == extension.vsKubernetesExplorer"
+                    "when": "view == extension.vsKubernetesExplorer",
+                    "group": "0"
                 },
                 {
                     "command": "extension.vsKubernetesConfigureFromCluster",
-                    "when": "view == extension.vsKubernetesExplorer"
+                    "when": "view == extension.vsKubernetesExplorer",
+                    "group": "0"
                 },
                 {
                     "command": "extension.vsKubernetesUseKubeconfig",
-                    "when": "view == extension.vsKubernetesExplorer"
+                    "when": "view == extension.vsKubernetesExplorer",
+                    "group": "1"
                 },
                 {
                     "command": "extension.vsKubernetesRefreshHelmRepoExplorer",
--- a/src/components/clusterprovider/minikube/minikube.ts
+++ b/src/components/clusterprovider/minikube/minikube.ts
@@ -7,9 +7,10 @@ import { Shell, ShellResult } from '../../../shell';
 import { Host } from '../../../host';
 import { FS } from '../../../fs';
 import * as binutil from '../../../binutil';
-import { Errorable } from '../../../errorable';
+import { Errorable, failed } from '../../../errorable';
 import { fromShellExitCodeOnly, Diagnostic } from '../../../wizard';
 import { getToolPath } from '../../config/config';
+import { installMinikube } from '../../installer/installer';
 
 export class MinikubeInfo {
     readonly running: boolean;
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -254,6 +254,10 @@
                     "command": "extension.vsKubernetesConfigureFromCluster",
                     "when": "view == extension.vsKubernetesExplorer"
                 },
+                {
+                    "command": "extension.vsKubernetesUseKubeconfig",
+                    "when": "view == extension.vsKubernetesExplorer"
```
itowlson: Let's put a separator between this and the Add/Create Cluster commands (add `"group": 0` to the cluster commands and group 1 to this one).

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -75,6 +78,34 @@ class MinikubeImpl implements Minikube {
...
+    }
+}
+
+async function minikubeUpgradeAvailable(context: Context) {
+    if (!await checkPresent(context, CheckPresentMode.Alert)) {
+        // not installed, no upgrade.
+        return;
+    }
+
+    const sr = await context.shell.exec(`${context.binPath} update-check`);
```
itowlson: Check `sr.code`

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -75,6 +78,34 @@ class MinikubeImpl implements Minikube {
...
+    const lines = sr.stdout.split(os.EOL);
+    const currentVersion = extractVersion(lines[0]);
```
itowlson: I might be a bit paranoid but we could check here that `lines.length >= 2` so that if the output is not what we expect we can fail gracefully.  Probably not strictly necessary though.

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -75,6 +78,34 @@ class MinikubeImpl implements Minikube {
...
+    const availableVersion = extractVersion(lines[1]);
+    if (currentVersion !== availableVersion) {
+        vscode.window.showInformationMessage('Minikube upgrade available.', 'Install').then((value: string) => {
+            if (value === 'Install') {
+                vscode.commands.executeCommand('vscode.open', vscode.Uri.parse('https://github.com/kubernetes/minikube/releases/latest'));
```
itowlson: When we offer an install (via the `Install Dependencies` button) we usually download the dependency, drop it into the extension tools directory, and add a config entry so that the extension uses it.  See `src/components/installer/installer.ts` which already has an `installMinikube` function.  It would be great if we could adapt that function to fetch the latest version rather than whatever version was current when we last updated that code.  _cough cough_

Alternatively, rename this button to something like `View Latest Release`, so it's clear that it opens the release page rather than actually modifying the user's system.

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -75,6 +78,34 @@ class MinikubeImpl implements Minikube {
...
+        // not installed, no upgrade.
+        return;
+    }
+
+    const sr = await context.shell.exec(`${context.binPath} update-check`);
+    const lines = sr.stdout.split(os.EOL);
+    const currentVersion = extractVersion(lines[0]);
+    const availableVersion = extractVersion(lines[1]);
+    if (currentVersion !== availableVersion) {
+        vscode.window.showInformationMessage('Minikube upgrade available.', 'Install').then((value: string) => {
```
itowlson: It might be useful to display the new version (and possibly the version the user is using?).

src/kubectl.ts
```diff
@@ -129,13 +133,13 @@ async function checkForKubectlInternal(context: Context, errorMessageMode: Check
...
-    return await binutil.checkForBinary(context, bin, binName, inferFailedMessage, configuredFileMissingMessage, errorMessageMode !== 'silent');
+    return await binutil.checkForBinary(context, bin, binName, inferFailedMessage, configuredFileMissingMessage, errorMessageMode !== CheckPresentMessageMode.Silent);
 }
 
 function getCheckKubectlContextMessage(errorMessageMode: CheckPresentMessageMode): string {
-    if (errorMessageMode === 'activation') {
-        return ' Kubernetes commands other than configuration will not function correctly.';
-    } else if (errorMessageMode === 'command') {
+    if (errorMessageMode === CheckPresentMessageMode.Activation) {
+        return ' Kubernetes Commands other than configuration will not function correctly.';
```
itowlson: This should remain lower-case.

src/components/clusterprovider/minikube/minikube.ts
itowlson: Thinking about it, since you know `availableVersion` from the version check, it should be possible to modify the existing `installMinikube` function to take it as an argument and use it when constructing the download URL.  So making this follow the existing 'install dependencies' paradigm might not be too hard after all.

src/components/clusterprovider/minikube/minikube.ts
brendandburns: done.

package.json

src/components/clusterprovider/minikube/minikube.ts
brendandburns: done.

src/components/clusterprovider/minikube/minikube.ts
brendandburns: done.

src/components/clusterprovider/minikube/minikube.ts
brendandburns: done.

src/kubectl.ts
brendandburns: done.

brendandburns:
Comments addressed, sorry it took so long...

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -75,6 +79,45 @@ class MinikubeImpl implements Minikube {
...
+}
+
+async function minikubeUpgradeAvailable(context: Context) {
+    if (!await checkPresent(context, CheckPresentMode.Alert)) {
+        // not installed, no upgrade.
+        return;
+    }
+
+    const sr = await context.shell.exec(`${context.binPath} update-check`);
+    if (sr.code != 0) {
```
itowlson: Think you need triple-equals to satisfy the linter.

```suggestion
    if (sr.code !== 0) {
```

itowlson:
Confirmed it's just that one lint issue causing the CI to fail - if we can fix that then we're good to go (I can do that if you want).

brendandburns:
lint fixed.

brendandburns:
@itowlson friendly ping on this one.

Thanks!

itowlson:
Sorry about that!

brendandburns:
np, thanks!

PR #390, opened 2018-10-19
untested changes to check if kubectl is already installed

mtarng:
PR into my own master??? need to verify this actually works....


mtarng:
PR into the wrong repository...

Issue #389, opened 2018-10-19
Kubernetes extension breaks file links in "Go Tests" OUTPUT channel?


brackendawson:
With the Kubernetes extension disabled the line references in the stack trace are clickable:
![image](https://user-images.githubusercontent.com/1909753/47211145-add37a00-d38c-11e8-9662-fd37639ec32c.png)

But when enabled they are not:
![image](https://user-images.githubusercontent.com/1909753/47211126-a2804e80-d38c-11e8-81c6-7fa6e63fdabf.png)

itowlson:
That's so weird!  Apologies for the inconvenience this is causing you; I'll take a look.

itowlson:
Okay, this seems to happen because we colourise the output channel.  The way the colouriser works is global, and appears to override the built-in linkifier.  It looks like other extensions that colourise the output channel (such as https://marketplace.visualstudio.com/items?itemName=IBM.output-colorizer) have the same problem - I suspect whichever extension loads last overrides all the previous extensions _and_ the built-in highlighter/linkifier.

I think we probably need to remove the colouriser from the extension.  Unfortunately, the way VS Code loads syntaxes, we can't as far as I know make this conditional, so we can't even allow users to opt in via config.  But we don't make a lot of use of colourisation so I don't think it's a big loss!

itowlson:
Closed via #391.

PR #391, opened 2018-10-23
Don't apply a language config to the output window

itowlson:
We applied a language setting to the output window's MIME type in order to colourise headers for output printed to that window (#28).  Unfortunately, this hijacks the language away from the VS Code `Log` 'language', resulting in the default linkifier not running on stuff in the output window.  This means that when, say, test runners output line references to the output window, those references cease to be linkified (#389).

Losing links in test runs is worse than having to grovel for command headers in k8s output, so let's remove the colourisation.


itowlson:
Fixes #389.

Issue #335, opened 2018-08-07
Launch commands in Windows Bash WSL


jseguillon:
Hello,
Under Win10, I prefer using helm in bash subsystem than in powershell, simply because most of helm plugins are not working on powershell. I'd like to use my helm and kubectl from bash inside VS Code. 

Overidding `"terminal.integrated.shell.windows": "C:\\WINDOWS\\System32\\bash.exe"` does not apply. I also tried to change `"vs-kubernetes.helm-path": "C:\\WINDOWS\\System32\\bash.exe -c helm""` but then plugin answers : "C:\WINDOWS\System32\bash.exe -c helm does not exist!" 

Is there any way to launch helm and kubectl in bash ?

itowlson:
@radu-matei Sound like you started looking at this - is it okay to assign it to you?

radu-matei:
Surely, you can assign it to me.

@jseguillon - I also tried to add variations of `bash.exe -c` as the path for tools, but that is not going to work for various reasons:

- first, we check if it's Windows, and if so, `.exe` is added - https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/binutil.ts#L45

- then, the command itself (`get pods -o json`, for example) is added on top of the path + binary. Because of how we should call `bash.exe` binaries (mainly `bash.exe -c "kubectl get pods -o json"` - note the quotes), even if we skip adding the `.exe` extension, we still miss the correct quotes.

The fix I have in mind - and @itowlson, @bnookala and @jseguillon - tell me if it's acceptable - would be to add a top level `useWSL` config to the settings, and if true, handle all paths accordingly.

On top of that, there is also the discussion of: which Kubernetes config should we use - the one in Windows (`C:/Users/<user>/.kube/config` - this can be accessed from WSL as `/mnt/c/...`) or the one in WSL (`~/.kube/config`)

Thoughts?

itowlson:
Agree that whether to use the Windows or Linux binaries should be controlled by a setting, rather than kludged in by using a `bash` command line in a tool path.  Note that we don't yet have a consistent way of invoking tools so this might end up touching quite a few places (or you might have to try to rationalise the tool invocation code!).

[...]

PR #394, opened 2018-10-25
Add support for using the kubernetes extension on Windows via WSL kubectl/helm

brendandburns:
@itowlson

### DIFF
--- a/src/components/config/config.ts
+++ b/src/components/config/config.ts
@@ -128,7 +128,7 @@ export async function setAlwaysCleanUp(): Promise<void> {
 
 // Use WSL on Windows
 
-const USE_WSL_KEY = "useWsl";
+const USE_WSL_KEY = "use-wsl";
 
 export function getUseWsl(): boolean {
     return vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[USE_WSL_KEY];
--- a/src/components/config/config.ts
+++ b/src/components/config/config.ts
@@ -126,6 +126,14 @@ export async function setAlwaysCleanUp(): Promise<void> {
     await setConfigValue(AUTO_CLEANUP_DEBUG_KEY, true);
 }
 
+// Use WSL on Windows
+
+const USE_WSL_KEY = "useWsl";
+
+export function getUseWsl(): boolean {
+    return vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[USE_WSL_KEY];
+}
+
 // Other bits and bobs
 
 export function getOutputFormat(): string {
--- a/src/explainer.ts
+++ b/src/explainer.ts
@@ -7,27 +7,28 @@ import * as shelljs from 'shelljs';
 
 import { formatComplex, formatOne, Typed, formatType } from "./schema-formatting";
 import { getActiveKubeconfig, getUseWsl } from './components/config/config';
+import { shell } from '../test/fakes';
 
-function loadKubeconfig(): kubernetes.KubeConfig {
+async function loadKubeconfig(): Promise<kubernetes.KubeConfig> {
     const kc = new kubernetes.KubeConfig();
     const kubeconfig = getActiveKubeconfig();
     if (kubeconfig) {
         kc.loadFromFile(kubeconfig);
-    } else {
-        if (getUseWsl()) {
-            const result = shelljs.exec('wsl.exe cat $HOME/.kube/config', { silent: true }) as shelljs.ExecOutputReturnValue;
-            if (result.code === 0) {
-                kc.loadFromString(result.stdout);
-                return kc;
-            }
-        }
-        kc.loadFromDefault();
+        return kc;
     }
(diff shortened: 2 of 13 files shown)


itowlson:
cc @radu-matei since you were interested in this

itowlson:
Relevant: #335.

itowlson:
@brendandburns The CI environment has changed and broken the linter - could you rebase onto master please, as this contains a fix/workaround, then your CI should start passing.  Thanks!

brendandburns:
@itowlson rebased.

brendandburns:
@itowlson I fixed this so that port-forward, describe, attach, logs (with following) and terminal all appear to work correctly.

I think this is pretty much complete at this point...

brendandburns:
Fixed the lint problem...

brendandburns:
@itowlson I think this is ready to go. ptal.

Thanks!

src/components/config/config.ts
```diff
@@ -126,6 +126,14 @@ export async function setAlwaysCleanUp(): Promise<void> {
     await setConfigValue(AUTO_CLEANUP_DEBUG_KEY, true);
 }
 
+// Use WSL on Windows
+
+const USE_WSL_KEY = "useWsl";
```
itowlson: We usually hyphenate rather than camel-case config keys I think - suggest

```suggestion
const USE_WSL_KEY = "use-wsl";
```

src/explainer.ts
```diff
@@ -3,18 +3,31 @@
...
 
-export function readSwagger(): Promise<any> {
+function loadKubeconfig(): kubernetes.KubeConfig {
     const kc = new kubernetes.KubeConfig();
     const kubeconfig = getActiveKubeconfig();
     if (kubeconfig) {
         kc.loadFromFile(kubeconfig);
     } else {
+        if (getUseWsl()) {
+            const result = shelljs.exec('wsl.exe cat $HOME/.kube/config', { silent: true }) as shelljs.ExecOutputReturnValue;
```
itowlson: Do we need to handle the case where they have WSL, have `kubectl` under WSL, but use a KUBECONFIG under WSL to use a non-default kubeconfig file?

itowlson: Consider making this async, as I believe this is called only from methods that return promises, and could therefore return a promise itself without having too much of a ripple effect.

src/shell.ts
```diff
@@ -50,7 +50,7 @@ export interface ShellResult {
 export type ShellHandler = (code: number, stdout: string, stderr: string) => void;
 
 function isWindows(): boolean {
-    return (process.platform === WINDOWS);
+    return (process.platform === WINDOWS) && !getUseWsl();
```
itowlson: I'm anxious this is going to break places where we use `isWindows()` to understand how to work with the _host_ filesystem, e.g. constructing file URIs for use with the `Diff` command.

itowlson: Think you need to update the `platform()` function too, which also examines `process.platform`.  `platform()` is used for installing tools so need to consider whether these get correctly installed into WSL in this situation (e.g. we download and untar to the file system directly, and currently assume it's the host file system).

src/kubectl.ts
```diff
@@ -194,6 +194,10 @@ async function spawnAsChild(context: Context, command: string[]): Promise<ChildP
 
 async function invokeInTerminal(context: Context, command: string, pipeTo: string | undefined, terminal: Terminal): Promise<void> {
     if (await checkPresent(context, 'command')) {
+        // You might be tempted to think we needed to add 'wsl' here if user is using wsl
+        // but this runs in the context of a vanilla terminal, which is controlled by the
+        // existing preference, so it's not necessary.
+        // But a user does need to default VS code to use WSL in the settings.json
```
itowlson: Might be good to check for this and provide a warning notification if they are using a PowerShell terminal with a WSLised extension, as this could cause weird errors e.g. a resource appears in the extension, but commands that interact with it in the terminal fail because the Windows kubeconfig has a different active cluster from the WSL kubeconfig.

src/components/config/config.ts
brendandburns: done.

src/explainer.ts
brendandburns: done.

src/shell.ts
brendandburns: Tested that diff works correctly.

brendandburns: Install now works correctly.

src/kubectl.ts
brendandburns: Can we get at vs code preferences like this?

src/explainer.ts
brendandburns: done.

brendandburns:
@itowlson I think this is now ready to go.... (at least for a first draft)

brendandburns:
@itowlson friendly ping on this one.

Thanks!

PR #396, opened 2018-10-25
Nail down Node version so CI doesn't break

### LINKED ISSUE #395: Something changed in CI and now everything is terrible
CI PR builds have been failing for a few days with:

```
ReferenceError: internalBinding is not defined
    at fs.js:25:1
...
The command "gulp tslint" exited with 1.
```

I re-ran the master and that (which previously passed) also failed with the same error.  This suggests something has changed in the CI environment and now no builds are passing.

I hate computers.

itowlson:
Fixes #395.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-- 'node'
+- "10"
 
 os:
 - linux


itowlson:
Emergency fix so PRs in flight can complete - probably good to clarify if there is a real underlying issue on Node 11 but this gets us building again for now.

PR #404, opened 2018-11-14
Update gulp to fix some security issues...

brendandburns:
@itowlson 

the dreaded red banner strikes again :)


brendandburns:
Actually, this revs vscode too, I don't think we want that, I'll revert that part...

bnookala:
@brendandburns @itowlson the build still fails since the `package-lock.json` file still has mentions of the problematic version of `event-stream`: https://github.com/Azure/vscode-kubernetes-tools/pull/404/files#L2050. 

To resolve this, we'll need to delete and re-generate the package-lock file, and add it again.

itowlson:
@bnookala can you regenerate package-lock and push to @brendandburns' branch directly using your maintainer powers?  If not, I can do it later when I'm near a computer.  Thanks!

Issue #399, opened 2018-10-31
ResourceLimitsLinter should be optional


tehho:
Hi.

Could you implement a disabler for ResourceLimitsLinter?
Im not really interested in resource management for my pods and would like the green squiggles to go away.

Many thanks :)

brendandburns:
Fixed by #411

PR #411, opened 2018-12-05
Add configuration for enable/disable linters.

### LINKED ISSUE #408: Config Lint Rules?
It seems like there is a limited linter built into this plugin and it looks like it reports when a resource limit hasn't been set in a container spec. Is there a way to turn off all linting, or better yet, a specific rule? I end up with big blocks of distracting green squiggles when I'm mid way defining a new deployment and haven't gotten around to thinking about what resource limits I want.

I found [this code](https://github.com/Azure/vscode-kubernetes-tools/blob/c5c14b0872107e9d9e4d46c5a38a386e5c43abdd/src/components/lint/resourcelimits.ts), but I'm not seeing anything that indicates some kind of preferences check (though I'm not really familiar with the implementation of the project, so I likely wouldn't see it if it's there). Thanks.

brendandburns:
Fixes #408 #399


src/extension.ts
```diff
@@ -1941,12 +1941,30 @@ function isLintable(document: vscode.TextDocument): boolean {
...
     if (!isLintable(document)) {
         return;
     }
-    const linterPromises = linters.map((l) => l.lint(document));
+    const enabledLinters = config.getEnabledLinters();
+    if (enabledLinters.length === 0) {
+        return;
+    }
+    const linterPromises: Promise<vscode.Diagnostic[]>[] = [];
+    linters.map((l) => {
```
itowlson: Don't use `map` for side-effectful iteration.  `map` is for transforming one array to another.  I think you can write this as something like:

```
const linterPromises =
    linters.filter((l) => linterEnabled(enabledLinters, l.name()))
           .map((l) => l.lint(document));
```

(I would probably add an `isEnabled` helper function to make the `filter` more readable.)

src/components/config/config.ts
```diff
@@ -139,3 +139,19 @@ export function getConfiguredNamespace(): string | undefined {
 export function affectsUs(change: vscode.ConfigurationChangeEvent) {
     return change.affectsConfiguration(EXTENSION_CONFIG_KEY);
 }
+
+export function getEnabledLinters(): string[] {
+    const config = vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY);
+    return getEnabledLintersInternal(config['disable-lint'], config['use-linters']);
```
itowlson: These are new config settings - they should be described in `package.json::contibutions` and `readme`.

src/components/config/config.ts
```diff
@@ -139,3 +139,19 @@ export function getConfiguredNamespace(): string | undefined {
...
+}
+
+// only exported for testing...
+export function getEnabledLintersInternal(disableLint: string, linters: string): string[] {
+    if (disableLint === 'true') {
+        return [];
+    }
+    if (linters) {
+        return linters.trim().split(',').map((linter: string) => linter.trim());
```
itowlson: Consider making the `use-linters` config entry an array rather than a comma separated string.

src/components/config/config.ts
```diff
@@ -139,3 +139,19 @@ export function getConfiguredNamespace(): string | undefined {
...
+    }
+    return ["all"];
```
itowlson: I might be wrong but this looks like it would result in only linters named `all` being enabled - we don't seem to check for this magic string elsewhere.

src/extension.ts
```diff
@@ -1941,12 +1941,30 @@ function isLintable(document: vscode.TextDocument): boolean {
     return document.languageId === 'yaml' || document.languageId === 'json' || document.languageId === 'helm';
 }
 
+function linterEnabled(enabledLinters: string[], name: string): boolean {
+    for (const linter of enabledLinters) {
```
itowlson: Pretty sure you can use `some` or even `indexOf` here:

`return enabledLinters.some((l) => l === name);`

or

`return enabledLinters.indexOf(name) >= 0;`

src/extension.ts
```diff
@@ -1941,12 +1941,30 @@ function isLintable(document: vscode.TextDocument): boolean {
...
+    const enabledLinters = config.getEnabledLinters();
+    if (enabledLinters.length === 0) {
+        return;
+    }
+    const linterPromises: Promise<vscode.Diagnostic[]>[] = [];
+    linters.map((l) => {
+        if (linterEnabled(enabledLinters, l.name())) {
+            linterPromises.push(l.lint(document));
+        }
+    });
```
itowlson: Depending on your config defaults it looks like this requires users to opt into every linter.  This makes new linters essentially useless as nobody will ever read the docs to find out about them.  It may also vex Seans Knox and McKenna, who asked for the resource limits linter because people were shooting themselves in the foot deploying unrestricted pods to AKS and then raising starvation as a service issue.

A preferable user interface would be:

* Linters are **on by default**.
* You can deal with a lint warning by fixing it, by disabling the linter at workspace level, or by disabling the linter at global level.

I would prefer not to have the option to disable all (present and future) linting, but if others disagree then we can have that as a manual setting (_not_ exposed as a diagnostic fix in the UI).

src/components/config/config.ts
brendandburns: done.

src/components/config/config.ts
brendandburns: done.

src/components/config/config.ts
brendandburns: fixed.

src/extension.ts

src/extension.ts

src/extension.ts
brendandburns: Fixed this so the list is ones to disable, not ones to enable, and the default is all on.

I don think we want a "make it stop" option. 

With things like linting someone is always going to be annoyed, we need them to stop the bleeding as much as they can.

brendandburns:
@itowlson Comments addressed, revereed from 'enabled' to 'disabled' which feels cleaner.

Please take another look.

src/components/lint/resourcelimits.ts
```diff
@@ -7,6 +7,10 @@ import { warningOn, childSymbols } from './linter.utils';
 // .containers[each].resources.limits.{cpu,memory}
 
 export class ResourceLimitsLinter implements LinterImpl {
+    name(): string {
+        return "resource-limits";
```
bnookala: doesn't need to happen here, but we should document standard linters

brendandburns: added doc to the readme.

package.json
```diff
@@ -179,6 +179,14 @@
...
                             "type": "string",
                             "description": "Output format for Kubernetes specs. One of 'json' or 'yaml' (default)."
+                        },
+                        "disable-lint": {
+                            "type": "boolean",
+                            "description": "Disable all linting of Kubernetes files"
+                        },
+                        "disable-linters": {
+                            "type": "array",
+                            "description": "Lit of linters by name to disable"
```
itowlson: typo 
```suggestion
                            "description": "List of linters by name to disable"
```

src/extension.ts
```diff
@@ -1941,12 +1941,23 @@ function isLintable(document: vscode.TextDocument): boolean {
...
+    return disabledLinters.some((l) => l === name);
+}
+
 async function kubernetesLint(document: vscode.TextDocument): Promise<void> {
     // Is it a Kubernetes document?
     if (!isLintable(document)) {
         return;
     }
-    const linterPromises = linters.map((l) => l.lint(document));
+    if (config.getDisableLint()) {
```
itowlson: Move this to the top of the function: if linting is globally turned off, there's no point examining the document to see if we are interested in it.

package.json

src/extension.ts

brendandburns:
Comments addressed, please take another look (after the holidays of course :)

Thanks!

PR #412, opened 2018-12-07
add resource limit to code snippet

andxu:
Suppress the error message "One or more containers do not have resource limits - this could starve other processes"


brendandburns:
lgtm.

PR #418, opened 2018-12-13
Send telemetry when create and add cluster

Flanker32:
Send telemetry when create and add cluster

### DIFF
--- a/src/components/clusterprovider/azure/azureclusterprovider.ts
+++ b/src/components/clusterprovider/azure/azureclusterprovider.ts
@@ -217,7 +217,7 @@ async function createCluster(previousData: any, context: azure.Context): Promise
         } };
     const createResult = await azure.createCluster(context, options);
 
-    reporter.sendTelemetryEvent("createCluster", { result: createResult.result.succeeded ? "success" : "fail", clusterType: previousData.clusterType });
+    reporter.sendTelemetryEvent("clustercreation", { result: createResult.result.succeeded ? "success" : "failure", clusterType: previousData.clusterType });
 
     const title = createResult.result.succeeded ? 'Cluster creation has started' : `Error ${createResult.actionDescription}`;
     const additionalDiagnostic = diagnoseCreationError(createResult.result);
@@ -284,7 +284,7 @@ function renderConfigurationResult(configureResult: ActionResult<azure.Configure
     const configResult = result.result;
     const clusterServiceString = result.result.clusterType === "aks" ? "Azure Kubernetes Service" : "Azure Container Service";
 
-    reporter.sendTelemetryEvent("addCluster", { result: configureResult.result.succeeded ? "success" : "fail", clusterType: clusterServiceString });
+    reporter.sendTelemetryEvent("clusterregistration", { result: configureResult.result.succeeded ? "success" : "failure", clusterType: configResult.clusterType });
 
     const pathMessage = configResult.cliOnDefaultPath ? '' :
         '<p>This location is not on your system PATH. Add this directory to your path, or set the VS Code <b>vs-kubernetes.kubectl-path</b> config setting.</p>';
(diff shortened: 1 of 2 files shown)


squillace:
Hi there! Modifying telemetry is something we do very slowly. Can you exclaim the precise purpose you have here?

Flanker32:
@squillace For now we send telemetries when use click create kubernetes, but it does not mean they really compete the creation, we want to know how many users really create and use it, so i add telemetries when we finished create operation and add clusters.

itowlson:
Who is 'we' please?  You've described what your changes _do_, but we knew that from the Code - 'purpose' means who wants to use this information and for what?  Telemetry has a cost in terms of community acceptance and trust; any new telemetry really needs to be discussed and agreed with program managers who can factor in that cost vs. the value of what we learn from it.

Flanker32:
@itowlson I come from VSCode Java Team, the goal of my team is to win more Java developers using Azure, and one of the BI task is to get the number of Java developer attaching Azure in VS Code.  I add this telemetry is to get the number of user who really create and use kubernetes provided by Azure and found the java user among them.

itowlson:
We are now good to go with this, though have had a big refactor so it will need a rebase I'm afraid.

src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -283,6 +284,8 @@ async function createCluster(previousData: any, context: azure.Context): Promise
         } };
     const createResult = await azure.createCluster(context, options);
 
+    reporter.sendTelemetryEvent("createCluster", { result: createResult.result.succeeded ? "success" : "fail", clusterType: previousData.clusterType });
```
itowlson: The disclosure statement in the README needs to be updated to reflect the additional telemetry.

itowlson: "success/failure" or "succeeded/failed"

itowlson: Elsewhere we have all-lowercased event names e.g. `cloudselection`, and we've used nouns rather than verbs.  Suggest `clustercreation` perhaps, or `clustercreated` or something like that?

src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -353,6 +356,8 @@ function renderConfigurationResult(configureResult: ActionResult<azure.Configure
     const configResult = result.result;
     const clusterServiceString = result.result.clusterType === "aks" ? "Azure Kubernetes Service" : "Azure Container Service";
 
+    reporter.sendTelemetryEvent("addCluster", { result: configureResult.result.succeeded ? "success" : "fail", clusterType: clusterServiceString });
```
itowlson: Similar comments on the `result` strings and the event name (consider `clusterregistration` or `clusteradded`)

itowlson: `clusterServiceString` is the long name ('Azure Kubernetes Service'); elsewhere telemetry uses the cluster type ID (e.g. `aks`).

src/components/clusterprovider/azure/azureclusterprovider.ts
itowlson: Is there anything we can do to pre-empt GDPR classification for these new events and fields, or do we have to wait for them to roll in and the 'you have unclassified data' notifications to start?

Flanker32: Yes，the readme has been updated

Flanker32: @itowlson We could trigger this telemetry by ourself when pr merged , so that the GDPR checker could find the new trace and tag the traces soon, and it seems that's all what we can do for GDPR.

itowlson:
Needed a couple of tweaks and a rebase, otherwise good to go if CI passes.

PR #425, opened 2019-01-07
Switch from previewHtml to Webview API

### LINKED ISSUE #401: Upgrade to VS Code's webview API
Hi, I'm on the VS Code team. I noticed that your extension uses the `vscode.previewHtml` command which is deprecated and which we are actively working to remove: https://github.com/Microsoft/vscode/issues/62630

We've developed a [webview API][docs] that provides a much better developer experience and offers a number of important security and compatibility benefits over `previewHtml`. We cannot fix `previewHtml` without breaking backwards compatibility, and have instead opted to remove it. 

To ensure that your extension continues to work properly in VS Code, please try upgrading to use the new Webview API. You can find documentation on the API usage [here][docs]. Let me know if you have any questions or concerns about this migration

[docs]: https://code.visualstudio.com/docs/extensions/webview

itowlson:
We are required to do this because VS Code is deprecating the `previewHtml` command and will shortly start to issue user-facing bleats about it.  Main change is to the 'create cluster/add existing cluster' wizardry, but also a couple of Helm windows (to which I have taken a 'minimal effort' approach as they weren't mixed up with HTTP nonsense).

Fixes #401.


src/components/clusterprovider/clusterproviderutils.ts
```diff
@@ -1,50 +1,63 @@
 'use strict';
 
 import * as clusterproviderregistry from './clusterproviderregistry';
-import * as clusterproviderserver from './clusterproviderserver';
+// import * as clusterproviderserver from './clusterproviderserver';
```
brendandburns: delete dead code.

src/components/clusterprovider/clusterproviderutils.ts
```diff
@@ -1,50 +1,63 @@
...
 
 export async function renderWizardContainer(action: clusterproviderregistry.ClusterProviderAction): Promise<string> {
-    await clusterproviderserver.init();
+    return '';
+//     await clusterproviderserver.init();
```
brendandburns: here too.

src/components/clusterprovider/common/form.ts
```diff
@@ -1,11 +1,17 @@
...
 
 // HTML rendering boilerplate
 
+const DO_NOT_PROPAGATE = ['nextStep', SENDING_STEP_KEY];
+
 export function propagationFields(previousData: any): string {
     let formFields = "";
     for (const k in previousData) {
-        formFields = formFields + `<input type='hidden' name='${k}' value='${previousData[k]}' />\n`;
+        if (DO_NOT_PROPAGATE.indexOf(k) < 0) {
```
brendandburns: why not `DO_NOT_PROPOGATE.includes(k)` ?

src/components/wizard/wizard.ts
```diff
@@ -0,0 +1,70 @@
...
+        for (const e of document.forms['${formId}'].elements) {
+            s[e.name] = e.value;
+        }
+        vscode.postMessage(s);
+    }
+    </script>`;
+
+    const html = `<div id='wvcontent__' />${nextScript}`;
+
+    const w = vscode.window.createWebviewPanel('vsk8s-dialog', tabTitle, vscode.ViewColumn.Active, {
```
brendandburns: Do we want to make this a singleton?

src/utils/preview.ts
```diff
@@ -0,0 +1,16 @@
+import * as vscode from "vscode";
+
+export async function preview(uri: vscode.Uri, column: vscode.ViewColumn, tabTitle: string): Promise<void> {
+    const html = await getHTML(uri);
+    const w = vscode.window.createWebviewPanel('vsk8s-preview', tabTitle, column, {
```
brendandburns: as above wrt singleton

src/utils/preview.ts
```diff
@@ -0,0 +1,16 @@
...
+        retainContextWhenHidden: false,
+        enableScripts: false,
+    });
+    w.webview.html = html;
+    w.reveal();
+}
+
+async function getHTML(uri: vscode.Uri): Promise<string> {
```
brendandburns: it would be nice to move all of the HTML snippets (eventually) to files (or file fragments) in the extension itself...

brendandburns:
basically LGTM, small comments.

src/components/clusterprovider/common/form.ts
itowlson: TypeScript gives me an error if I use `includes`.  I'm not sure why that is.  Perhaps I need to update the target JS version in the TypeScript config, but I'm going to call 'out of scope' on that for this PR...

src/components/wizard/wizard.ts
itowlson: No, that would result in the same window being reused if a user started one cluster operation while another was in progress at the same time, and might make the state management harder though I've tried to avoid that.  New wizard should be new panel in new tab.

src/utils/preview.ts
itowlson: Definitely not in this case, because a user might very well want to have two previews up at the same time.  This is shared infrastructure between multiple commands so even if one command wanted to reuse its webview, that should not leak across to sharing with other commands.

src/utils/preview.ts
itowlson: Yes, though a lot of them use JavaScript string interpolation so I'm not sure how feasible that will be.  Definitely out of scope for this PR which is quite scary enough already!

itowlson:
Rebased and addressed comments; please take a look if you have time, otherwise will merge in a couple of days.

itowlson:
Forgot to say - thanks for the suggestions, and for catching the mountains of dead code!

brendandburns:
LGTM.

PR #430, opened 2019-01-08
Fix for running Helm Update Dependencies from the file explorer

itowlson:
In the olden days, if you right-clicked `requirements.yaml` in the file explorer and chose Update Dependencies, it would ignore the context and prompt you for which chart to update.  This was bad.  With the coming of #426, this was improved by making it misinterpret the context and silently error out of the command.  This was worse.

This PR makes it update the dependencies of the chart containing the clicked `requirements.yaml`, without prompting or crashing or any of that fun stuff.  This... _glances to side at PM, blinks rapidly to reader in Morse code_... this is better.

### DIFF
--- a/src/helm.exec.ts
+++ b/src/helm.exec.ts
@@ -81,19 +81,20 @@ export function helmTemplatePreview() {
     helm.recordPreviewHasBeenShown();
 }
 
-export function helmDepUp(arg: any) {
-    if (arg) {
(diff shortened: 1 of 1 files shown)


itowlson:
FINE FINE OKAY I'LL PUT BACK THE `IGNITE WHALE FAT` COMMAND BUT I STILL SAY YOU AND RALPH ARE THE ONLY ONES WHO USE IT.

PR #432, opened 2019-01-09
CRD implementation

### LINKED ISSUE #431: CRD support
It would be cool if the extension supported Kubernetes [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/). Ideally, the extension could allow

- showing installed CRD *types* on the cluster
- showing/editing actual CRD *instances* on the cluster

dgkanatsios:
Fixes #431 
thanks @itowlson for the awesome guidance!

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -407,8 +407,8 @@ class KubernetesCRDFolder extends KubernetesFolder {
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
-        const objects = await kubectlUtils.getGlobalResources(kubectl, "crd");
-        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.metadata.name, obj.metadata.name, obj.metadata.name, obj.metadata.name)));
+        const objects = await kubectlUtils.getCRDTypes(kubectl);
+        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.metadata.name, obj.pluralName, obj.kind, obj.abbreviation)));
     }
 }
 
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -145,6 +145,7 @@ class KubernetesContextNode implements KubernetesObject {
             new KubernetesResourceFolder(kuberesources.allKinds.ingress),
             new KubernetesStorageFolder(),
             new KubernetesConfigFolder(),
+            new KubernetesCRDFolder(kuberesources.allKinds.crd),
             new HelmReleasesFolder(),
         ];
     }
@@ -200,7 +201,6 @@ class KubernetesWorkloadFolder extends KubernetesFolder {
             new KubernetesResourceFolder(kuberesources.allKinds.job),
             new KubernetesResourceFolder(kuberesources.allKinds.cronjob),
             new KubernetesResourceFolder(kuberesources.allKinds.pod),
-            new KubernetesCRDFolder(kuberesources.allKinds.crd),
         ];
     }
 }
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -145,7 +145,7 @@ class KubernetesContextNode implements KubernetesObject {
             new KubernetesResourceFolder(kuberesources.allKinds.ingress),
             new KubernetesStorageFolder(),
             new KubernetesConfigFolder(),
-            new KubernetesCRDFolder(kuberesources.allKinds.crd),
+            new KubernetesCRDFolder(),
             new HelmReleasesFolder(),
         ];
     }
@@ -402,13 +402,13 @@ class KubernetesSelectsPodsFolder extends KubernetesResourceFolder {
 }
 
 class KubernetesCRDFolder extends KubernetesFolder {
-    constructor(readonly kind: kuberesources.ResourceKind) {
-        super(kind.abbreviation, kind.pluralDisplayName);
+    constructor() {
+        super(kuberesources.allKinds.crd.abbreviation, kuberesources.allKinds.crd.pluralDisplayName);
     }
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
         const objects = await kubectlUtils.getCRDTypes(kubectl);
-        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.metadata.name, obj.spec.names.plural, obj.kind, obj.spec.names.shortNames !== null ? (obj.spec.names.shortNames.length > 0 ? obj.spec.names.shortNames[0] : "") : "")));
+        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.spec.names.singular, obj.spec.names.plural, obj.spec.names.kind, obj.spec.names.shortNames !== null ? (obj.spec.names.shortNames.length > 0 ? obj.spec.names.shortNames[0] : obj.metadata.name) : obj.metadata.name)));
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/explorer.ts
```diff
@@ -200,6 +200,7 @@ class KubernetesWorkloadFolder extends KubernetesFolder {
             new KubernetesResourceFolder(kuberesources.allKinds.job),
             new KubernetesResourceFolder(kuberesources.allKinds.cronjob),
             new KubernetesResourceFolder(kuberesources.allKinds.pod),
+            new KubernetesCRDFolder(kuberesources.allKinds.crd),
```
bnookala: You should be able to use `KubernetesSelectsPodsFolder` here instead of creating `KubernetesCRDFolder`

bnookala: actually @itowlson informs me that your implementation is fine, so resolving this!

src/kuberesources.ts
```diff
@@ -19,6 +19,7 @@ export const allKinds = {
     job: new ResourceKind("Job", "Jobs", "Job", "job"),
     cronjob: new ResourceKind("CronJob", "CronJobs", "CronJob", "cronjob"),
     pod: new ResourceKind("Pod", "Pods", "Pod", "pod"),
+    crd: new ResourceKind("CRD", "CRDs", "CRD", "crd"),
```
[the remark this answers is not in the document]
itowlson: To clarify: the first two arguments are about how to display the resource type, and can be a matter of taste.  The third **MUST** correspond to the `kind:` value in the k8s manifest YAML.  The fourth must be something you can pass to `kubectl get ...` to list instances.  It would not be a bad plan if I documented this better.

src/explorer.ts
itowlson: I'm doubtful about putting this under Workloads.  A CRD may represent a workload, but it may also represent a piece of configuration data that does not directly translate into executable work, such as a storage resource or an external service (a la svccat scenarios).

If other k8s UIs put CRDs under Workloads then let's go along with that, otherwise I'd suggest putting this at the top level.

src/kuberesources.ts
itowlson: Don't have strong feelings about this but wondered if we should spell it out in the display strings - `Custom Resource/s` or `Custom Resource Definition/s`.  Thoughts?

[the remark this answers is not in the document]
dgkanatsios: fixing, thanks!

dgkanatsios: Custom Resource/s seems better, fixed!

src/explorer.ts
dgkanatsios: You've got a point, changing that as well.

squillace: Agree. CRDs could literally be anything, so top-level for now would be my take, but Ivan decides.

src/kuberesources.ts
squillace: yup

src/explorer.ts
```diff
@@ -400,6 +401,17 @@ class KubernetesSelectsPodsFolder extends KubernetesResourceFolder {
...
 }
 
+class KubernetesCRDFolder extends KubernetesFolder {
+    constructor(readonly kind: kuberesources.ResourceKind) {
+        super(kind.abbreviation, kind.pluralDisplayName);
+    }
+
+    async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        const objects = await kubectlUtils.getGlobalResources(kubectl, "crd");
+        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.metadata.name, obj.metadata.name, obj.metadata.name, obj.metadata.name)));
```
itowlson: I hadn't realised before, but the CRD resource type has `names.plural`, `names.singular` and `names.kind` (and an optional `names.shortNames` array).  I think we should _at minimum_ use `names.kind` as the manifest kind (argument 3), and consider using `names.singular` as the abbreviation (argument 4).  Arguments 1 and 2 aren't completely clear cut, because the singular and plural in the manifest are programmatic not display names, but they would arguably still be friendlier than the CRD name in all its dotted glory.  Thoughts?

itowlson:
I noticed if I try to edit a CRD instance I get a red wiggly on the first line with the diagnostic message `Unable to load schema from 'kubernetes://schema/test.hestia.cc/v1@reticulator': No content.`  Should we try to tackle that as part of this work, or defer for a separate work item?  If the latter we should raise an issue to track it.

src/explorer.ts
dgkanatsios: hm, the `obj` is of type `KubernetesResource` which only has `.metadata.name` and `.kind` properties. Should we create a `getCRDs` method that returns a `CRDInfo` array (or something like that)?

dgkanatsios:
Cannot repro, seems normal to me (or am I looking at the wrong thing?)

capture

src/explorer.ts
```diff
@@ -400,6 +401,17 @@ class KubernetesSelectsPodsFolder extends KubernetesResourceFolder {
...
 }
 
+class KubernetesCRDFolder extends KubernetesFolder {
+    constructor(readonly kind: kuberesources.ResourceKind) {
+        super(kind.abbreviation, kind.pluralDisplayName);
+    }
+
+    async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
+        const objects = await kubectlUtils.getCRDTypes(kubectl);
+        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.metadata.name, obj.spec.names.plural, obj.kind, obj.spec.names.shortNames !== null ? (obj.spec.names.shortNames.length > 0 ? obj.spec.names.shortNames[0] : "") : "")));
```
itowlson: Are you sure you want `obj.kind` rather than `obj.names.kind`?  Shouldn't the first argument (`displayName`) be `obj.names.singular`?  And for abbreviation (fourth argument) I'd suggest that if `shortNames` is empty then it should be `obj.metadata.name` rather than the empty string?

src/kuberesources.objectmodel.ts
```diff
@@ -52,6 +52,20 @@ export interface ContainerStatus {
...
+    readonly spec: CRDSpec;
+}
+
+export interface CRDSpec {
+    readonly names: CRDNames;
+}
+
+export interface CRDNames {
+    readonly plural: string;
+    readonly singular: string;
```
itowlson: This should have a `kind` member as well.

src/explorer.ts
dgkanatsios: Yeah, I did it this way because I recalled that singular is not required. From what I see `obj.names.singular` is not required **but** it `Defaults to lowercased <kind>` (kubectl explain crd.spec.names.singular), so I'm gonna do it your way, thanks!

src/kubectlUtils.ts
```diff
@@ -130,6 +130,22 @@ export async function getGlobalResources(kubectl: Kubectl, resource: string): Pr
...
+    const crdTypes = await kubectl.asJson<KubernetesCollection<any>>(`get crds -o json`);
+    if (failed(crdTypes)) {
+        vscode.window.showErrorMessage(crdTypes.error[0]);
+        return [];
+    }
+
+    return crdTypes.result.items.map((item) => {
+        return {
+            metadata: item.metadata,
+            kind: item.spec.names.kind,
```
itowlson: I don't think this is right.  The `kind` of a Custom Resource _Definition _is `CustomResourceDefinition`; this seems to confuse the kind of the _definition_ with the kind of an _instance_.  I don't think we use this, though, so I think maybe we should just merge the PR to unblock you, and tweak later.

Issue #435, opened 2019-01-16
CRD explorer throws error


shreyu86:
Thanks for adding support for browsing CRD's while trying out I noticed that the explorer throws an error in the following form: 
![image](https://user-images.githubusercontent.com/311217/51229731-3f15a680-1912-11e9-8b17-b1a8bd4f3467.png)

PR #436, opened 2019-01-16
Fix for empty CRD error

shreyu86:
Possible fix for #435 , I am new to Typescript but turning on the debugger allowed me to possibly identify a fix.

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -408,7 +408,7 @@ class KubernetesCRDFolder extends KubernetesFolder {
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
         const objects = await kubectlUtils.getCRDTypes(kubectl);
-        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.spec.names.singular, obj.spec.names.plural, obj.spec.names.kind, obj.spec.names.shortNames !== null ? (obj.spec.names.shortNames.length > 0 ? obj.spec.names.shortNames[0] : obj.metadata.name) : obj.metadata.name)));
+        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.spec.names.singular, obj.spec.names.plural, obj.spec.names.kind, obj.spec.names.shortNames !== undefined ? (obj.spec.names.shortNames.length > 0 ? obj.spec.names.shortNames[0] : obj.metadata.name) : obj.metadata.name)));
     }
 }
 


dgkanatsios:
Not a TS expert myself, but from [what I'm reading](https://stackoverflow.com/questions/28975896/is-there-a-dedicated-function-to-check-null-and-undefined-in-typescript) maybe we could do a `obj.spec.names.shortNames != null` to check for both null and undefined at once?

shreyu86:
@dgkanatsios that works, if that checks for both the conditions.

itowlson:
I don't think that does check for both the conditions though - `null !== undefined` is true I believe.

src/explorer.ts
```diff
@@ -408,7 +408,7 @@ class KubernetesCRDFolder extends KubernetesFolder {
 
     async getChildren(kubectl: Kubectl, host: Host): Promise<KubernetesObject[]> {
         const objects = await kubectlUtils.getCRDTypes(kubectl);
-        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.spec.names.singular, obj.spec.names.plural, obj.spec.names.kind, obj.spec.names.shortNames !== null ? (obj.spec.names.shortNames.length > 0 ? obj.spec.names.shortNames[0] : obj.metadata.name) : obj.metadata.name)));
+        return objects.map((obj) => new KubernetesResourceFolder(new kuberesources.ResourceKind(obj.spec.names.singular, obj.spec.names.plural, obj.spec.names.kind, obj.spec.names.shortNames !== undefined ? (obj.spec.names.shortNames.length > 0 ? obj.spec.names.shortNames[0] : obj.metadata.name) : obj.metadata.name)));
```
itowlson: I think we can simplify this to something like: `(obj.spec.names.shortNames && obj.spec.names.shortNames.length > 0) ? obj.spec.names.shortNames[0] : obj.metadata.name` using the fact that null and undefined are both falsy.  (I have a feeling empty arrays are falsy too in which case we could even drop the length check.  But not sure.)

itowlson:
@shreyu86 I wanted to get a release under the way so I took the liberty of pushing an update directly to your branch.  I hope this doesn't coming across as treading on your toes.  I very much appreciate you catching this and identifying the required fix - just wanted to tweak what you had done without going around a review cycle (especially given timezones).  Thanks in advance for being understanding!

shreyu86:
@itowlson Thanks a lot! No worries at all, really appreciate the note. 

Have to save the extension has become a massive productive boost and part of the daily workflow.

itowlson:
Aargh @shreyu86 I am truly sorry for this but I was in a rush today with the release and had already done the changelog and forgot to update it to credit and thank you for your contribution.  I'll add you to the credits for 0.1.15 so it's visible in the repo and in future releases, but I don't think I can do-over this release.  Many apologies!

itowlson:
Good heavens, and I forgot you too @dgkanatsios.  I'm not exactly being a community champion today, ugh.

shreyu86:
@itowlson 🙌 thanks for the very kind note, no worries at all! Thanks again!

dgkanatsios:
Yeah, no worries, thanks!

PR #440, opened 2019-01-18
Add optional create/apply/delete on right-click.

### LINKED ISSUE #406: Right click on YAML file to create/delete/apply resource
Hey

It would be nice to be able, to deploy various kubernetes resources without resolving to commands.
Current workflow is:
1. Open YAML file.
2. Ctrl+p
3. >Kubernetes: Create/Apply/Delete

It would be if we cloud also:
1. Find file in explorer.
2. Right click on it and pick from menu what we want to do with.

In some cases it is easier to look for file in tree and then just click on it instead of opening it and doing addtional 2 steps (I might not remember names of all configurations for example).

brendandburns:
Fixes #406 

@itowlson @iniside

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,9 +24,12 @@
     "icon": "images/k8s-logo.png",
     "activationEvents": [
         "onCommand:extension.vsKubernetesCreate",
+        "onCommand:extension.vsKubernetesCreateFile",
         "onCommand:extension.vsKubernetesDelete",
         "onCommand:extension.vsKubernetesDeleteNow",
+        "onCommand:extension.vsKubernetesDeleteUri",
         "onCommand:extension.vsKubernetesApply",
+        "onCommand:extension.vsKubernetesApplyFile",
         "onCommand:extension.vsKubernetesExplain",
         "onCommand:extension.vsKubernetesLoad",
         "onCommand:extension.vsKubernetesGet",
@@ -187,6 +190,10 @@
                         "disable-linters": {
                             "type": "array",
                             "description": "List of linters by name to disable"
+                        },
+                        "resource-commands-on-files": {
+                            "type": "boolean",
+                            "description": "If true, show Kubernetes resource commands on file context menu for all YAML files"
                         }
                     },
                     "default": {
@@ -236,6 +243,24 @@
                     "when": "resourceFilename == requirements.yaml",
                     "command": "extension.helmDepUp",
                     "group": "2_helm@99"
+                },
+                {
+                    "title": "Create Kubernetes resource",
+                    "when": "resourceLangId == yaml && config.vs-kubernetes.resource-commands-on-files",
+                    "command": "extension.vsKubernetesCreateFile",
+                    "group": "2_k8s_1"
+                },
+                {
+                    "title": "Apply Kubernetes resource",
+                    "when": "resourceLangId == yaml && config.vs-kubernetes.resource-commands-on-files",
+                    "command": "extension.vsKubernetesApplyFile",
+                    "group": "2_k8s_2"
+                },
+                {
+                    "title": "Delete Kubernetes resource",
+                    "when": "resourceLangId == yaml && config.vs-kubernetes.resource-commands-on-files",
+                    "command": "extension.vsKubernetesDeleteUri",
+                    "group": "2_k8s_3"
                 },
                 {
                     "when": "",
@@ -599,6 +624,18 @@
             ],
             "commandPalette": [
                 {
+                    "command": "extension.vsKubernetesApplyFile",
+                    "when": "filesExplorerFocus"
+                },
+                {
+                    "command": "extension.vsKubernetesCreateFile",
+                    "when": "filesExplorerFocus"
+                },
+                {
+                    "command": "extension.vsKubernetesDeleteUri",
+                    "when": "filesExplorerFocus"
+                },
+                {
                     "command": "extension.vsKubernetesRefreshExplorer",
                     "when": "view == extension.vsKubernetesExplorer"
                 },
@@ -659,6 +696,11 @@
                 "category": "Kubernetes"
             },
             {
+                "command": "extension.vsKubernetesCreateFile",
+                "title": "Create Kubernetes resource",
+                "category": "Kubernetes"
+            },
+            {
                 "command": "extension.vsKubernetesDelete",
                 "title": "Delete",
                 "category": "Kubernetes"
@@ -669,8 +711,18 @@
                 "category": "Kubernetes"
             },
             {
+                "command": "extension.vsKubernetesDeleteUri",
+                "title": "Delete Kubernetes resource",
+                "category": "Kubernetes"
+            },
+            {
                 "command": "extension.vsKubernetesApply",
                 "title": "Apply",
+                "category": "Kubernetes"
+            },
+            {
+                "command": "extension.vsKubernetesApplyFile",
+                "title": "Apply Kubernetes resource",
                 "category": "Kubernetes"
             },
             {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -142,6 +142,12 @@ export async function activate(context: vscode.ExtensionContext): Promise<extens
         registerCommand('extension.vsKubernetesCreate',
             maybeRunKubernetesCommandForActiveWindow.bind(this, 'create', "Kubernetes Creating...")
         ),
+        registerCommand('extension.vsKubernetesCreateFile',
+            (uri: vscode.Uri) => kubectlUtils.createResourceFromUri(uri, kubectl)),
+        registerCommand('extension.vsKubernetesDeleteUri',
+            (uri: vscode.Uri) => kubectlUtils.deleteResourceFromUri(uri, kubectl)),
+        registerCommand('extension.vsKubernetesApplyFile',
+            (uri: vscode.Uri) => kubectlUtils.applyResourceFromUri(uri, kubectl)),
         registerCommand('extension.vsKubernetesDelete', (explorerNode: explorer.ResourceNode) => { deleteKubernetes(KubernetesDeleteMode.Graceful, explorerNode); }),
         registerCommand('extension.vsKubernetesDeleteNow', (explorerNode: explorer.ResourceNode) => { deleteKubernetes(KubernetesDeleteMode.Now, explorerNode); }),
         registerCommand('extension.vsKubernetesApply', applyKubernetes),
--- a/src/kubectlUtils.ts
+++ b/src/kubectlUtils.ts
@@ -367,3 +367,33 @@ export async function getResourceAsJson<T extends KubernetesResource | Kubernete
     }
     return shellResult.result;
 }
+
+export async function createResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    await changeResourceFromUri(uri, kubectl, 'create', 'creating', 'created');
+}
+
+export async function deleteResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    const result = await vscode.window.showWarningMessage('Are you sure you want to delete this resource?', 'Delete', 'Cancel');
+    if (result === 'Delete') {
+        await changeResourceFromUri(uri, kubectl, 'delete', 'deleting', 'deleted');
+    }
+}
+
+export async function applyResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    await changeResourceFromUri(uri, kubectl, 'apply', 'applying', 'applied');
+}
+
+async function changeResourceFromUri(uri: vscode.Uri, kubectl: Kubectl, command: string, verbParticiple: string, verbPast: string) {
+    if (uri.scheme !== 'file') {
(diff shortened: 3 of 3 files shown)


src/kubectlUtils.ts
```diff
@@ -1,344 +1,369 @@
...
+        await changeResourceFromUri(uri, kubectl, 'delete', 'deleting', 'deleted');
+    }
+}
+
+export async function applyResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    await changeResourceFromUri(uri, kubectl, 'apply', 'applying', 'applied');
+}
+
+async function changeResourceFromUri(uri: vscode.Uri, kubectl: Kubectl, verb: string, errorVerb: string, successVerb: string) {
+    const path = vscode.workspace.asRelativePath(uri);
```
itowlson: I am not sure this will work correctly in multi-root workspaces.  Is there a reason you want to pass a relative path rather than an absolute one?

src/kubectlUtils.ts
```diff
@@ -1,344 +1,369 @@
...
+    if (result === 'Delete') {
+        await changeResourceFromUri(uri, kubectl, 'delete', 'deleting', 'deleted');
+    }
+}
+
+export async function applyResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    await changeResourceFromUri(uri, kubectl, 'apply', 'applying', 'applied');
+}
+
+async function changeResourceFromUri(uri: vscode.Uri, kubectl: Kubectl, verb: string, errorVerb: string, successVerb: string) {
```
itowlson: Should we check that this is a `file:` URI?  (It always should be, I know - just thinking about meaningful errors if it somehow isn't!)

src/kubectlUtils.ts
```diff
@@ -1,344 +1,369 @@
...
-export async function getResourceAsJson<T extends KubernetesResource | KubernetesCollection<KubernetesResource>>(kubectl: Kubectl, resourceId: string, resourceNamespace?: string): Promise<T | undefined> {
-    const nsarg = resourceNamespace ? `--namespace ${resourceNamespace}` : '';
-    const shellResult = await kubectl.asJson<T>(`get ${resourceId} ${nsarg} -o json`);
-    if (failed(shellResult)) {
-        vscode.window.showErrorMessage(shellResult.error[0]);
-        return undefined;
-    }
-    return shellResult.result;
-}
+import * as vscode from "vscode";
```
itowlson: I'm seeing whitespace diffs on every line of this file... any idea what's causing those?  Would be nice to have only the functional diffs in this PR, and if our whitespace is messed up then address that in a separate non-functional PR.

src/extension.ts
```diff
@@ -143,6 +143,12 @@ export async function activate(context): Promise<extensionapi.ExtensionAPI> {
         registerCommand('extension.vsKubernetesCreate',
             maybeRunKubernetesCommandForActiveWindow.bind(this, 'create', "Kubernetes Creating...")
         ),
+        registerCommand('extension.vsKubernetesCreateFile',
+            async (uri: vscode.Uri) => kubectlUtils.createResourceFromUri(uri, kubectl)),
```
itowlson: You're not using `await` in the lambda body so you don't need the lambda to be `async`.

package.json
```diff
@@ -668,11 +698,21 @@
                 "title": "Delete Now",
                 "category": "Kubernetes"
             },
+            {
+                "command": "extension.vsKubernetesDeleteUri",
+                "title": "Delete Kubernetes resource",
+                "category": "Kubernetes Menu"
```
itowlson:
```suggestion
                "category": "Kubernetes"
```

package.json
```diff
@@ -658,6 +683,11 @@
                 "title": "Create",
                 "category": "Kubernetes"
             },
+            {
```
itowlson: I don't see an entry in the `menus.commandPalette` section, so these commands will be invokable from the command palette. If that's intentional, then you need to handle the case where they are invoked without a URI (i.e. the argument needs to be optional).

package.json
```diff
@@ -668,11 +698,21 @@
                 "title": "Delete Now",
                 "category": "Kubernetes"
             },
+            {
+                "command": "extension.vsKubernetesDeleteUri",
```
itowlson: Not sure why this one is `...Uri` when Create and Apply are `...File`?

src/kubectlUtils.ts
```diff
@@ -1,344 +1,369 @@
...
+    }
+    return shellResult.result;
+}
+
+export async function createResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    await changeResourceFromUri(uri, kubectl, 'create', 'creating', 'created');
+}
+
+export async function deleteResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    const result = await vscode.window.showWarningMessage('Are you sure you?', 'Delete', 'Cancel');
```
itowlson: That confirmation message is probably not what you intended *grin*.

src/kubectlUtils.ts
itowlson: We might want to call the last two parameters something like `verbParticiple` and `verbPast`.  For example, I could see a long-running operation using a progress box of the form `Creating foo... please wait` which would use the participle in a non-error context.  But this could be a future refactoring if and when someone used those forms differently.

src/kubectlUtils.ts
```diff
@@ -1,344 +1,369 @@
...
+    await changeResourceFromUri(uri, kubectl, 'apply', 'applying', 'applied');
+}
+
+async function changeResourceFromUri(uri: vscode.Uri, kubectl: Kubectl, verb: string, errorVerb: string, successVerb: string) {
+    const path = vscode.workspace.asRelativePath(uri);
+    const result = await kubectl.invokeAsync(`${verb} -f ${path}`);
+    if (result.code !== 0) {
+        vscode.window.showErrorMessage(`Error ${errorVerb} resource: ${result.stderr}`);
+    } else {
+        vscode.window.showInformationMessage(`Resource ${successVerb}.`);
```
itowlson: Consider identifying the resource (also in the error case, though stderr would hopefully provide the info there); this could be helpful if a user kicked off several commands (running through a directory deleting things).  Which ones have completed?  Which resource was the one that failed?

src/kubectlUtils.ts
```diff
@@ -1,344 +1,369 @@
...
+
+export async function applyResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    await changeResourceFromUri(uri, kubectl, 'apply', 'applying', 'applied');
+}
+
+async function changeResourceFromUri(uri: vscode.Uri, kubectl: Kubectl, verb: string, errorVerb: string, successVerb: string) {
+    const path = vscode.workspace.asRelativePath(uri);
+    const result = await kubectl.invokeAsync(`${verb} -f ${path}`);
+    if (result.code !== 0) {
+        vscode.window.showErrorMessage(`Error ${errorVerb} resource: ${result.stderr}`);
```
itowlson: Sometimes stderr can be too long to fit in a VS Code error notification; you might want to also dump it to the output channel.

src/kubectlUtils.ts
itowlson: Suggest renaming `verb` to `command` as this is specifically the kubectl command name, not an English descriptive verb like the other two (which is what I assumed until I read the implementation).

src/kubectlUtils.ts
```diff
@@ -1,344 +1,369 @@
...
+    }
+}
+
+export async function applyResourceFromUri(uri: vscode.Uri, kubectl: Kubectl) {
+    await changeResourceFromUri(uri, kubectl, 'apply', 'applying', 'applied');
+}
+
+async function changeResourceFromUri(uri: vscode.Uri, kubectl: Kubectl, verb: string, errorVerb: string, successVerb: string) {
+    const path = vscode.workspace.asRelativePath(uri);
+    const result = await kubectl.invokeAsync(`${verb} -f ${path}`);
```
itowlson: Consider quoting `path` as it might contain spaces.

package.json
```diff
@@ -237,6 +244,24 @@
                     "command": "extension.helmDepUp",
                     "group": "2_helm@99"
                 },
+                {
+                    "title": "Create Kubernetes resource",
+                    "when": "resourceLangId == yaml && config.vs-kubernetes.resource-context-menu",
+                    "command": "extension.vsKubernetesCreateFile",
+                    "group": "2_k8s"
```
itowlson: You're not specifying any ordering within this group.  Should we do that, e.g. Create, Apply, Delete?

package.json
```diff
@@ -187,6 +190,10 @@
                         "disable-linters": {
                             "type": "array",
                             "description": "List of linters by name to disable"
+                        },
+                        "resource-context-menu": {
```
itowlson: This name feels a bit opaque.  `resource-commands-on-files` perhaps?

package.json
```diff
@@ -187,6 +190,10 @@
...
+                            "type": "boolean",
+                            "description": "If true, show create/apply/delete in the context menu when you right click on any YAML file"
```
itowlson: I'm not sure we need the 'is true' and I don't think we should list the specific commands because that will need to be maintained.  How about e.g. `Show Kubernetes resource commands on file context menu for all YAML files`?

package.json

package.json

package.json

package.json
brendandburns: What do I need to do to hide them from the command pallette? (thanks)

package.json
brendandburns: I wanted to use vsKubernetesDeleteFile but it's already defined somewhere else :)

package.json

src/extension.ts

src/kubectlUtils.ts
brendandburns: My editor replaced \r\n with \n, it's kind of hard to revert, is it required?

src/kubectlUtils.ts
brendandburns: done.

src/kubectlUtils.ts
brendandburns: done.

brendandburns: done.

brendandburns: done.

src/kubectlUtils.ts
brendandburns: It doesn't work with WSL otherwise (the 'c:/' bit gets all mangled) I'm willing to go global and do the right fix if you insist.

src/kubectlUtils.ts
brendandburns: done.

src/kubectlUtils.ts
brendandburns: done.

brendandburns:
Comments mostly addressed, please re-check.

package.json

src/kubectlUtils.ts
brendandburns: actually had to fix this to be able to rebase :)

src/kubectlUtils.ts
itowlson: No, leave it as is; I think we have other places where we need to look at the multi-root scenario too, so let's tackle them all as a separate pass.

package.json
```diff
@@ -237,6 +244,24 @@
                     "command": "extension.helmDepUp",
                     "group": "2_helm@99"
                 },
+                {
+                    "title": "Create Kubernetes resource",
+                    "when": "resourceLangId == yaml && config.vs-kubernetes.resource-context-menu",
```
itowlson: Need to update the config entry name here and in other `when` clauses.

src/kubectlUtils.ts
```diff
@@ -367,3 +367,33 @@ export async function getResourceAsJson<T extends KubernetesResource | Kubernete
...
+    await changeResourceFromUri(uri, kubectl, 'apply', 'applying', 'applied');
+}
+
+async function changeResourceFromUri(uri: vscode.Uri, kubectl: Kubectl, command: string, verbParticiple: string, verbPast: string) {
+    if (uri.scheme !== 'file') {
+        vscode.window.showErrorMessage(`${uri.toString()} is not a file path.`);
+        return;
+    }
+    const path = vscode.workspace.asRelativePath(uri);
+    const result = await kubectl.invokeAsync(`${command} -f '${path}'`);
```
itowlson: I'm not sure single quotes will work on Windows.  Single quotes seem to be interpreted by the Windows shell as filename characters, and if I run `kubectl delete -f 'some_file_that_exists.yaml'` at the DOS prompt then I get a 'path does not exist' error.  Is it working for you in non-WSL Windows?  If not I think double quotes work on all OSes.

package.json
brendandburns: done. (I actually did all the others, but missed this one...)

src/kubectlUtils.ts
brendandburns: done.

brendandburns:
Comments addressed (thanks for the pointer one the command pallette)

please take another look.

Issue #441, opened 2019-01-18
Checking minikube updates doesn't support space in path to check


apupier:
I got this notification message poping up:

`Error checking for minikube updates: 'C:\Users\Aurelien' is not recognized as an internal or external command, operable program or batch file.`

my username is Aurelien Pupier

apupier:
I think on this line https://github.com/Azure/vscode-kubernetes-tools/blob/d4834fc95bc2f42bb6f70f683595e3d73420ffb0/src/components/clusterprovider/minikube/minikube.ts#L94

double quotes needs to be provided around ${context.binPath}

itowlson:
Fixed via #442.

PR #442, opened 2019-01-18
Support space in path to minikube bin #441

apupier:
is there a way to modify test to have a bin path with space?

please note that i'm unable to launch the test suite localy. Seems that i have a compile error (not related to this PR)
`src/components/kubectl/port-forward.ts(237,13): error TS2322: Type '{ succeeded: boolean; pod: string; namespace: string; fromOpenDocument: true; }' is not assignable to type 'PortForwardFindPodsResult'.`

### DIFF
--- a/src/components/clusterprovider/minikube/minikube.ts
+++ b/src/components/clusterprovider/minikube/minikube.ts
@@ -91,7 +91,7 @@ async function minikubeUpgradeAvailable(context: Context) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


brendandburns:
LGTM.

itowlson:
Thanks @apupier - good catch and thanks for the fix.

Our unit tests are broken I'm afraid, so don't worry about the test failure... at this stage I suspect the tests are doing more damage by confusing helpful folks like you.  Sorry!

PR #451, opened 2019-01-24
add configs

yehiyam:
Hi
First let me say that I really like the extension. I use it all the time.
I've noticed that currently attaching a debugger to pod only works for java.
In this pull request, I've added support for nodejs

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,7 +16,10 @@
             "outFiles": [
                 "${workspaceFolder}/out/**/*.js"
             ],
-            "preLaunchTask": "npm: watch"
+            "preLaunchTask": "npm: watch",
+            "skipFiles": [
+                "<node_internals>/**/*.js"
+            ]
         },
         {
             "name": "Extension Tests",
--- a/debug-on-kubernetes.md
+++ b/debug-on-kubernetes.md
@@ -13,6 +13,10 @@ One of the key features of VS Code Kubernetes Extension is its one-click debuggi
 ## 3. Extension Settings for debugging
    * `vs-kubernetes` - Parent for Kubernetes-related extension settings
       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up if you wanted to view pod logs, etc.
+      * `vs-kubernetes.nodejs-autodetect-remote-root` - If true will try to automatically get the root location of the source code in the container. The detection attempts to get the CWD of the nodejs process.  
+      * `vs-kubernetes.nodejs-remote-root` - User specified root location of the source code in the container. Remote root is passed to the debug configuration to allow vscode to map local files to files in the container. If `vs-kubernetes.nodejs-autodetect-remote-root` is set to `false` and `vs-kubernetes.nodejs-remote-root` is empty, no remoteRoot is set in the debug configuration. For more info on nodejs remote debugging see [vscode documentation](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_remote-debugging)
+      * `vs-kubernetes.nodejs-debug-port` - Sets the nodejs remote debugging port. The default for node is 9229.
+
    * `vsdocker.imageUser` - Image prefix for docker images e.g. 'docker.io/brendanburns'.
 
 ## 4. How to use it for Java debugging
--- a/package.json
+++ b/package.json
@@ -180,6 +180,18 @@
                             "type": "string",
                             "description": "Output format for Kubernetes specs. One of 'json' or 'yaml' (default)."
                         },
+                        "vs-kubernetes.nodejs-autodetect-remote-root": {
+                            "type": "boolean",
+                            "description": "If true will try to automatically get the root location of the source code in the container."
+                        },
+                        "vs-kubernetes.nodejs-remote-root": {
+                            "type": "string",
+                            "description": "User specified root location of the source code in the container."
+                        },
+                        "vs-kubernetes.nodejs-debug-port": {
+                            "type": "string",
+                            "description": "Remote debugging port for nodejs. Usually 9229."
+                        },
                         "disable-lint": {
                             "type": "boolean",
                             "description": "Disable all linting of Kubernetes files"
@@ -198,7 +210,10 @@
                         "vs-kubernetes.outputFormat": "yaml",
                         "vs-kubernetes.kubeconfig": "",
                         "vs-kubernetes.knownKubeconfigs": [],
-                        "vs-kubernetes.autoCleanupOnDebugTerminate": false
+                        "vs-kubernetes.autoCleanupOnDebugTerminate": false,
+                        "vs-kubernetes.nodejs-autodetect-remote-root": true,
+                        "vs-kubernetes.nodejs-remote-root": "",
+                        "vs-kubernetes.nodejs-debug-port": "9229"
                     }
                 },
                 "vsdocker.imageUser": {
--- a/src/debug/nodejsDebugProvider.ts
+++ b/src/debug/nodejsDebugProvider.ts
@@ -1,0 +1,102 @@
+import * as path from "path";
+import * as vscode from "vscode";
+import { ChildProcess } from "child_process";
+
+import { IDebugProvider, PortInfo } from "./debugProvider";
+import * as config from '../components/config/config';
+import { Kubectl } from "../kubectl";
+import { IDockerfile } from "../docker/parser";
+import { kubeChannel } from "../kubeChannel";
+
+interface ExecResult {
+    readonly proxyProcess: ChildProcess;
+}
+const debuggerType = 'nodejs';
+
+export class NodejsDebugProvider implements IDebugProvider {
+    remoteRoot: String;
+    shell: string;
+    public getDebuggerType(): string {
+        return debuggerType;
+    }
+    public async isDebuggerInstalled(): Promise<boolean> {
+        return true;
+    }
+    public async startDebugging(workspaceFolder: string, sessionName: string, port: number): Promise<boolean> {
+
+        const debugConfiguration = {
+            type: "node",
+            request: "attach",
+            name: sessionName,
+            hostName: "localhost",
+            skipFiles: [
+                "<node_internals>/**/*.js"
+            ],
+            port
+        };
+        if (this.remoteRoot) {
+            debugConfiguration['remoteRoot'] = this.remoteRoot;
+        }
+        const currentFolder = vscode.workspace.workspaceFolders.find((folder) => folder.name === path.basename(workspaceFolder));
+        return await vscode.debug.startDebugging(currentFolder, debugConfiguration);
+    }
+    isSupportedImage(baseImage: string): boolean {
+        throw new Error("Method not implemented.");
+    }
+
+    resolvePortsFromFile(dockerfile: IDockerfile, env: {}): Promise<PortInfo> {
+        throw new Error("Method not implemented.");
+    }
+
+    public async resolvePortsFromContainer(kubectl: Kubectl, pod: string, podNamespace: string, container: string): Promise<PortInfo> {
+        this.shell = await this.suggestedShellForContainer(kubectl, pod, podNamespace, container);
+        await this.setNodeInspectMode(kubectl, pod, podNamespace, container);
+        this.remoteRoot = await this.tryGetRemoteRoot(kubectl, pod, podNamespace, container);
+        const rawDebugPortInfo = config.getNodejsDebugPort() || '9229';
+        return {
+            debugPort: Number(rawDebugPortInfo)
+        };
+    }
+
+    async setNodeInspectMode(kubectl: Kubectl, pod: string, podNamespace: string, container: string): Promise<Boolean> {
+        kubeChannel.showOutput('Switching node to debug mode in container');
+        const nsarg = podNamespace ? `--namespace ${podNamespace}` : '';
+        const execCmd = `exec ${pod} ${nsarg} ${container ? "-c ${selectedContainer}" : ""} -- ${this.shell} -c 'kill -SIGUSR1 1'`;
+        const execResult = await kubectl.invokeAsync(execCmd);
+        return execResult.code == 0; 
+    }
+
+    async tryGetRemoteRoot(kubectl: Kubectl, podName: string, podNamespace: string | undefined, containerName: string | undefined): Promise<String> {
+        let remoteRoot: String;
+        if (config.getNodejsAutoDetectRemoteRoot()) {
+            kubeChannel.showOutput("Trying to detect remote root.");
+            const nsarg = podNamespace ? `--namespace ${podNamespace}` : '';
+            const execCmd = `exec ${podName} ${nsarg} ${containerName ? "-c ${selectedContainer}" : ""} -- ${this.shell} -c 'readlink /proc/1/cwd'`;
+            const execResult = await kubectl.invokeAsync(execCmd);
+            if (execResult.code == 0) {
+                remoteRoot = execResult.stdout.replace(/(\r\n|\n|\r)/gm, '');
+                kubeChannel.showOutput(`Got remote root from container: ${remoteRoot}`);
+            }
+        }
+        else if (config.getNodejsRemoteRoot()) {
+            remoteRoot = config.getNodejsRemoteRoot();
+            kubeChannel.showOutput(`Setting remote root to ${remoteRoot}`);
+        }
+        return remoteRoot;
+    }
+
+    async isBashOnContainer(kubectl: Kubectl, podName: string, podNamespace: string | undefined, containerName: string | undefined): Promise<boolean> {
+        const nsarg = podNamespace ? `--namespace ${podNamespace}` : '';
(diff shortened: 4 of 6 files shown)


src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+
+interface ExecResult {
+    readonly proxyProcess: ChildProcess;
+}
+const debuggerType = 'nodejs';
+
+export class NodejsDebugProvider implements IDebugProvider {
+    remoteRoot: String;
+    shell: string;
+    public getDebuggerType(): string {
```
brendandburns: nit: here and throughout, newlines before functions.

brendandburns:
Sweet! One small nit comment for readability, then I think this is good to go....

squillace:
@radu-matei for thoughts on other debuggers, and draft debugging. Great addition here to the root extension.

.vscode/launch.json
```diff
@@ -16,7 +16,10 @@
             "outFiles": [
                 "${workspaceFolder}/out/**/*.js"
             ],
-            "preLaunchTask": "npm: watch"
+            "preLaunchTask": "npm: watch",
+            "skipFiles": [
```
itowlson: Is this added intentionally?  If so, what's it for?

debug-on-kubernetes.md
```diff
@@ -13,6 +13,10 @@ One of the key features of VS Code Kubernetes Extension is its one-click debuggi
 ## 3. Extension Settings for debugging
    * `vs-kubernetes` - Parent for Kubernetes-related extension settings
       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up if you wanted to view pod logs, etc.
+      * `vs-kubernetes.nodejs-autodetect-remote-root` - If true will try to automatically get the root location of the source code in the container. The detection attempts to get the CWD of the nodejs process.
```
itowlson: I love that you've detailed the settings here - great attention to detail!  Thank you!

itowlson: Suggest some wording about debugging in this description.  Also clarify what it does when false.  E.g. "Controls how the extension locates the remote source code when debugging Node applications.  If true...  If false..."

debug-on-kubernetes.md
```diff
@@ -13,6 +13,10 @@ One of the key features of VS Code Kubernetes Extension is its one-click debuggi
...
       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up if you wanted to view pod logs, etc.
+      * `vs-kubernetes.nodejs-autodetect-remote-root` - If true will try to automatically get the root location of the source code in the container. The detection attempts to get the CWD of the nodejs process.  
+      * `vs-kubernetes.nodejs-remote-root` - User specified root location of the source code in the container. Remote root is passed to the debug configuration to allow vscode to map local files to files in the container. If `vs-kubernetes.nodejs-autodetect-remote-root` is set to `false` and `vs-kubernetes.nodejs-remote-root` is empty, no remoteRoot is set in the debug configuration. For more info on nodejs remote debugging see [vscode documentation](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_remote-debugging)
```
itowlson: This feels like it would vary a lot from application to application.  Should it be part of the debug launch config rather than an extension setting?  Are we allowed to do that?

package.json
```diff
@@ -180,6 +180,18 @@
                             "type": "string",
                             "description": "Output format for Kubernetes specs. One of 'json' or 'yaml' (default)."
                         },
+                        "vs-kubernetes.nodejs-autodetect-remote-root": {
+                            "type": "boolean",
+                            "description": "If true will try to automatically get the root location of the source code in the container."
+                        },
+                        "vs-kubernetes.nodejs-remote-root": {
+                            "type": "string",
+                            "description": "User specified root location of the source code in the container."
```
itowlson: Do we need 'user specified' here?  Seems like config entries are inherently 'user specified.'  Also should mentioned Node and debugging.  Maybe e.g.:
```suggestion
                            "description": "The root at which to find Node.JS source code in the container when debugging."
```

package.json
```diff
@@ -180,6 +180,18 @@
...
+                        },
+                        "vs-kubernetes.nodejs-debug-port": {
+                            "type": "string",
+                            "description": "Remote debugging port for nodejs. Usually 9229."
```
itowlson: Prefer Node, Node.js or Node.JS to `nodejs`.

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+import { kubeChannel } from "../kubeChannel";
+
+interface ExecResult {
+    readonly proxyProcess: ChildProcess;
+}
+const debuggerType = 'nodejs';
+
+export class NodejsDebugProvider implements IDebugProvider {
+    remoteRoot: String;
+    shell: string;
```
itowlson: `String` vs `string` across these two lines - I think we usually use `string`

debug-on-kubernetes.md
itowlson: Oh, we can't do this because we bypass `launch.json`?  Vexing if true... @radu-matei @testforstephen is there a way to put stuff like this in application data to be picked up by the debug provider?

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+            port
+        };
+        if (this.remoteRoot) {
+            debugConfiguration['remoteRoot'] = this.remoteRoot;
+        }
+        const currentFolder = vscode.workspace.workspaceFolders.find((folder) => folder.name === path.basename(workspaceFolder));
+        return await vscode.debug.startDebugging(currentFolder, debugConfiguration);
+    }
+
+    isSupportedImage(baseImage: string): boolean {
```
itowlson: This feels like it will cause problems when we try to find a debug provider for an image.  If this isn't the registered provider for any images then suggest returning `false` instead of throwing an exception.  cc @testforstephen

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        throw new Error("Method not implemented.");
+    }
+
+    resolvePortsFromFile(dockerfile: IDockerfile, env: {}): Promise<PortInfo> {
+        throw new Error("Method not implemented.");
```
itowlson: I'm not sure this is safe.  If we can't resolve the port(s) from the Dockerfile then I'd say return `undefined` here which will at least cause the session launcher to exit with a meaningful error.

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public async resolvePortsFromContainer(kubectl: Kubectl, pod: string, podNamespace: string, container: string): Promise<PortInfo> {
+        this.shell = await this.suggestedShellForContainer(kubectl, pod, podNamespace, container);
+        await this.setNodeInspectMode(kubectl, pod, podNamespace, container);
+        this.remoteRoot = await this.tryGetRemoteRoot(kubectl, pod, podNamespace, container);
```
itowlson: You don't seem to use `remoteRoot` here - is this setting it up for use by some other method that gets called later?  I'd prefer to avoid setting required state in something that appears to be a simple getter, if we can.  (But if there is nowhere else we can do it then so be it.)

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        const rawDebugPortInfo = config.getNodejsDebugPort() || '9229';
```
itowlson: Consider changing `config.getNodejsDebugPort()` to return `number | undefined` instead of `string`.

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        throw new Error("Method not implemented.");
+    }
+
+    resolvePortsFromFile(dockerfile: IDockerfile, env: {}): Promise<PortInfo> {
+        throw new Error("Method not implemented.");
+    }
+
+    public async resolvePortsFromContainer(kubectl: Kubectl, pod: string, podNamespace: string, container: string): Promise<PortInfo> {
+        this.shell = await this.suggestedShellForContainer(kubectl, pod, podNamespace, container);
+        await this.setNodeInspectMode(kubectl, pod, podNamespace, container);
```
itowlson: Should the result of this call be checked?

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        const rawDebugPortInfo = config.getNodejsDebugPort() || '9229';
+        return {
+            debugPort: Number(rawDebugPortInfo)
+        };
+    }
+
+    async setNodeInspectMode(kubectl: Kubectl, pod: string, podNamespace: string, container: string): Promise<Boolean> {
+        kubeChannel.showOutput('Switching node to debug mode in container');
+        const nsarg = podNamespace ? `--namespace ${podNamespace}` : '';
+        const execCmd = `exec ${pod} ${nsarg} ${container ? "-c ${selectedContainer}" : ""} -- ${this.shell} -c 'kill -SIGUSR1 1'`;
```
itowlson: Is this incantation one that Node debugging aficionados are likely to find familiar?  If not, could you annotate it a bit to help future maintainers understand what's going on?  (It's not obvious to me, but I realise that may just be my ignorance.)

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        const execResult = await kubectl.invokeAsync(execCmd);
+        return execResult.code === 0;
+    }
+
+    async tryGetRemoteRoot(kubectl: Kubectl, podName: string, podNamespace: string | undefined, containerName: string | undefined): Promise<String> {
```
itowlson: Again mixing `string` and `String` - I'd prefer to consistently use `string`.

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        let remoteRoot: String;
+        if (config.getNodejsAutoDetectRemoteRoot()) {
+            kubeChannel.showOutput("Trying to detect remote root.");
+            const nsarg = podNamespace ? `--namespace ${podNamespace}` : '';
+            const execCmd = `exec ${podName} ${nsarg} ${containerName ? "-c ${selectedContainer}" : ""} -- ${this.shell} -c 'readlink /proc/1/cwd'`;
+            const execResult = await kubectl.invokeAsync(execCmd);
+            if (execResult.code === 0) {
+                remoteRoot = execResult.stdout.replace(/(\r\n|\n|\r)/gm, '');
+                kubeChannel.showOutput(`Got remote root from container: ${remoteRoot}`);
+            }
```
itowlson: So if the `exec` fails this will fall through and return `undefined`?  Can we express this in the function signature please?  Also, does the provider handle the `undefined` case gracefully?

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+        else if (config.getNodejsRemoteRoot()) {
+            remoteRoot = config.getNodejsRemoteRoot();
+            kubeChannel.showOutput(`Setting remote root to ${remoteRoot}`);
+        }
+        return remoteRoot;
```
itowlson: To avoid the mutable variable and fallthrough, could we refactor this function along the lines of:

```
if (config.getAutoDetect...) {
  return await autoDetectRemoteRoot(...);
} else if (config.getNodejsRemoteRoot()) {
  return getConfiguredRemoteRoot();
} else {
  return undefined;
}
```

(and you can do the `kubeChannel` logging in the helper functions).  This will make the dispatch structure clearer I feel.

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    async isBashOnContainer(kubectl: Kubectl, podName: string, podNamespace: string | undefined, containerName: string | undefined): Promise<boolean> {
```
itowlson: Are these the same as the functions in `extension.ts`?  If so let's move them into a utility module, e.g. `components/container/container-shell.ts`.

.vscode/launch.json
yehiyam: When debugging the extension, and stepping into functions (F11) it skippes internal node code. It makes debugging easier

package.json
yehiyam: I wrote User specified as opposed to automatically detected. If you still think this is misleading, I will change it.

src/debug/nodejsDebugProvider.ts
yehiyam: `startDebugging` uses it. We could pass it back to the debugSession (like the port), but I think that since the `DebugProvider` instance lives only for the lifetime of the debug session, storing it in a member is fine. What do you think?

itowlson: I agree that storing it in a member is fine.  I just feel a bit awkward about modifying object state in a function called `resolvePortsFromContainer` (whose name implies that it is about reading the container settings to get a set of ports).  But if this is the place where we have the inputs we need (pod and container names) then I guess it is okay.  Perhaps in future we should amend the interface to provide a more natural home for this, but that can be out of scope for this PR.

yehiyam:
@itowlson, @testforstephen any more changes?

debug-on-kubernetes.md
```diff
@@ -13,6 +13,10 @@ One of the key features of VS Code Kubernetes Extension is its one-click debuggi
 ## 3. Extension Settings for debugging
    * `vs-kubernetes` - Parent for Kubernetes-related extension settings
       * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created Deployment and associated Pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up if you wanted to view pod logs, etc.
+      * `vs-kubernetes.nodejs-autodetect-remote-root` - This flags controlls how the extensions locates the root of the source code in the container. If true will try to automatically get the root location of the source code in the container. The detection attempts to get the CWD of the nodejs process. If false, the `remoteRoot` setting in the debug configuration will be set to `vs-kubernetes.nodejs-remote-root`.
```
itowlson: Typos - "controlls" -> controls, "extensions" -> extension.  Also had some wording suggestions:

```suggestion
      * `vs-kubernetes.nodejs-autodetect-remote-root` - This flags controls how the extension locates the root of the source code in the container. If true will try to automatically get the root location of the source code in the container, by getting the CWD of the nodejs process. If false, the `remoteRoot` setting in the debug configuration will be set to the value of the `vs-kubernetes.nodejs-remote-root` setting.  
```

package.json
itowlson: Yes, I would still like this to be changed please - thanks!

src/debug/debugUtils.ts
```diff
@@ -79,3 +79,16 @@ export async function getCommandsOfProcesses(kubectl: Kubectl, pod: string, podN
 
     return commandLines;
 }
+
+const isBashOnContainer = async (kubectl: Kubectl, podName: string, podNamespace: string | undefined, containerName: string | undefined): Promise<boolean> => {
```
itowlson: `async function` rather than `const`?  Some older code uses the "const of function type" design and I'm not sure why, but we are using plain ol' functions in most places.

src/debug/nodejsDebugProvider.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    public async resolvePortsFromContainer(kubectl: Kubectl, pod: string, podNamespace: string, container: string): Promise<PortInfo> {
+        this.shell = await suggestedShellForContainer(kubectl, pod, podNamespace, container);
+        const inspectModeResult = await this.setNodeInspectMode(kubectl, pod, podNamespace, container);
+        if (!inspectModeResult) {
+            kubeChannel.showOutput('Unable to set Node.js in the container to inspect mode.');
+            return undefined;
+        }
+        this.remoteRoot = await this.tryGetRemoteRoot(kubectl, pod, podNamespace, container);
+        const rawDebugPortInfo = config.getNodejsDebugPort() || '9229';
```
itowlson: Looks like the function now returns a number but you are `||`-ing it with a string.  Should it be:
```suggestion
        const rawDebugPortInfo = config.getNodejsDebugPort() || 9229;
```

src/extension.ts
```diff
@@ -51,6 +51,7 @@ import { logsKubernetes, LogsDisplayMode } from './components/kubectl/logs';
 import { Errorable, failed, succeeded } from './errorable';
 import { Git } from './components/git/git';
 import { DebugSession } from './debug/debugSession';
+import { suggestedShellForContainer } from './debug/debugUtils';
```
itowlson: This shouldn't be in `debugUtils` because it is used by non-debugging functions (specifically the `Terminal` command I think).  Consider a new module in the `utils` folder e.g. `containerutils.ts` or even `container-shell.ts` if you want to be specific.

yehiyam:
@itowlson I did the changes you requested. 
BTW, I found a few bugs that were introduced in one of the latest commits. I had to fix them too before committing my changes

PR #452, opened 2019-01-26
Add a HTML5-based logview

brendandburns:
@itowlson 

This adds a webview based log viewer. It adds 4 capabilities:
* Filter matches
* Hide matches
* Show logs that follow match
* Show logs that preceed match

Here's some screenshots to show how it works...

No filtering:
start

Matches `/health`
matches

Hides `/health`
not-matches

Logs after a specific date:
after

### DIFF
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -6,6 +6,9 @@ import * as kuberesources from '../../kuberesources';
 import { ResourceNode } from '../../explorer';
 import * as yaml from 'js-yaml';
 import * as kubectlUtils from '../../kubectlUtils';
+import { host } from '../../host';
+import { LogsPanel } from '../../components/logs/logsWebview';
+import { ShellResult } from '../../shell';
 
 export enum LogsDisplayMode {
     Show,
@@ -85,7 +88,13 @@ function getLogsForContainer(
         return;
     }
 
-    kubectl.invokeInSharedTerminal(cmd);
+    // kubectl.invokeInSharedTerminal(cmd);
+    kubectl.invokeAsync(cmd).then(
+        (result: ShellResult) => {
+            LogsPanel.createOrShow('TODO', result.stdout, `${podSummary.namespace}/${podSummary.name}`);
+        }, (err: any) => {
+            vscode.window.showErrorMessage(err);
+        });
 }
 
 /**
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -92,10 +92,12 @@ async function getLogsForContainer(
     const panel = LogsPanel.createOrShow('Loading...', resource);
 
     try {
-        console.log('executing ' + cmd);
         const result = await kubectl.invokeAsync(cmd);
-        console.log('got: ' + result.stdout);
-        panel.setInfo(result.stdout, resource);
+        if (result.code !== 0) {
+            vscode.window.showErrorMessage(`Error reading logs: ${result.stderr}`);
+        } else {
+            panel.setInfo(result.stdout, resource);
+        }
     } catch (err) {
         vscode.window.showErrorMessage(`Error reading logs ${err}`);
     }
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -89,7 +89,7 @@ async function getLogsForContainer(
     }
 
     const resource = `${podSummary.namespace}/${podSummary.name}`;
-    const panel = LogsPanel.createOrShow('Loading \u25CC', resource);
+    const panel = LogsPanel.createOrShow('Loading...', resource);
 
     try {
         console.log('executing ' + cmd);
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -88,10 +88,12 @@ function getLogsForContainer(
         return;
     }
 
-    // kubectl.invokeInSharedTerminal(cmd);
+    const resource = `${podSummary.namespace}/${podSummary.name}`;
+    const panel = LogsPanel.createOrShow('TODO', 'Loading \u25CC', resource);
+
     kubectl.invokeAsync(cmd).then(
         (result: ShellResult) => {
-            LogsPanel.createOrShow('TODO', result.stdout, `${podSummary.namespace}/${podSummary.name}`);
+            panel.setInfo(result.stdout, resource);
         }, (err: any) => {
             vscode.window.showErrorMessage(err);
         });
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -6,9 +6,7 @@ import * as kuberesources from '../../kuberesources';
 import { ResourceNode } from '../../explorer';
 import * as yaml from 'js-yaml';
 import * as kubectlUtils from '../../kubectlUtils';
-import { host } from '../../host';
 import { LogsPanel } from '../../components/logs/logsWebview';
-import { ShellResult } from '../../shell';
 
 export enum LogsDisplayMode {
     Show,
@@ -93,8 +91,8 @@ async function getLogsForContainer(
 
     try {
         const result = await kubectl.invokeAsync(cmd);
-        if (result.code !== 0) {
-            vscode.window.showErrorMessage(`Error reading logs: ${result.stderr}`);
+        if (!result || result.code !== 0) {
+            vscode.window.showErrorMessage(`Error reading logs: ${result ? result.stderr : undefined}`);
         } else {
             panel.setInfo(result.stdout, resource);
         }
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -60,13 +60,13 @@ async function getLogsForPod(kubectl: Kubectl, podSummary: PodSummary, displayMo
         return;
     }
 
-    getLogsForContainer(kubectl, podSummary, container.name, displayMode);
+    await getLogsForContainer(kubectl, podSummary, container.name, displayMode);
 }
 
 /**
  * Gets the logs for a container in a provided pod, in a provided namespace, in a provided container.
  */
-function getLogsForContainer(
+async function getLogsForContainer(
     kubectl: Kubectl,
     podSummary: PodSummary,
     containerName: string | undefined,
@@ -89,14 +89,16 @@ function getLogsForContainer(
     }
 
     const resource = `${podSummary.namespace}/${podSummary.name}`;
-    const panel = LogsPanel.createOrShow('TODO', 'Loading \u25CC', resource);
+    const panel = LogsPanel.createOrShow('Loading \u25CC', resource);
 
-    kubectl.invokeAsync(cmd).then(
-        (result: ShellResult) => {
-            panel.setInfo(result.stdout, resource);
-        }, (err: any) => {
-            vscode.window.showErrorMessage(err);
-        });
+    try {
+        console.log('executing ' + cmd);
+        const result = await kubectl.invokeAsync(cmd);
+        console.log('got: ' + result.stdout);
+        panel.setInfo(result.stdout, resource);
+    } catch (err) {
+        vscode.window.showErrorMessage(`Error reading logs ${err}`);
+    }
 }
 
 /**
--- a/src/components/logs/logsWebview.ts
+++ b/src/components/logs/logsWebview.ts
@@ -1,76 +1,26 @@
-import * as path from 'path';
 import * as vscode from 'vscode';
+import { WebPanel } from '../webpanel/webpanel';
 
-export class LogsPanel {
+export class LogsPanel extends WebPanel {
     public static readonly viewType = 'vscodeKubernetesLogs';
-    public static currentPanels = [];
+    public static currentPanels = new Map<string, LogsPanel>();
 
-    private disposables: vscode.Disposable[] = [];
-    private content: string;
-    private resource: string;
-
-    public static createOrShow(extensionPath: string, content: string, resource: string): LogsPanel {
-        const column = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.viewColumn : undefined;
-
-        // If we already have a panel, show it.
-        const currentPanel = LogsPanel.currentPanels[resource];
-        if (currentPanel) {
-            currentPanel.setInfo(content, resource);
-            currentPanel.update();
-            currentPanel.panel.reveal(column);
-            return currentPanel;
-        }
-        const panel = vscode.window.createWebviewPanel(LogsPanel.viewType, "Kubernetes Logs", column || vscode.ViewColumn.One, {
-            enableScripts: true,
-
-            // And restrict the webview to only loading content from our extension's `media` directory.
-            localResourceRoots: [
-            ],
-        });
-
-        const result = new LogsPanel(panel, content, resource);
-        LogsPanel.currentPanels[resource] = result;
-        return result;
+    public static createOrShow(content: string, resource: string): LogsPanel {
+        const fn = (panel: vscode.WebviewPanel, content: string, resource: string): LogsPanel => {
+            return new LogsPanel(panel, content, resource);
+        };
+        return WebPanel.createOrShowInternal<LogsPanel>(content, resource, LogsPanel.viewType, "Kubernetes Logs", LogsPanel.currentPanels, fn);
     }
 
     private constructor(
-        private readonly panel: vscode.WebviewPanel,
+        panel: vscode.WebviewPanel,
         content: string,
         resource: string
     ) {
-        this.content = content;
-        this.resource = resource;
-
-        this.update();
-        this.panel.onDidDispose(() => this.dispose(), null, this.disposables);
-
-        this.panel.onDidChangeViewState((e: any) => {
-            if (this.panel.visible) {
-                this.update();
-            }
-        }, null, this.disposables);
+        super(panel, content, resource, LogsPanel.currentPanels);
     }
 
-    public setInfo(content: string, resource: string) {
-        this.content = content;
-        this.resource = resource;
-        this.update();
-    }
-
-    public dispose() {
-        delete LogsPanel.currentPanels[this.resource];
-
-        this.panel.dispose();
-
-        while (this.disposables.length) {
-            const x = this.disposables.pop();
-            if (x) {
-                x.dispose();
-            }
-        }
-    }
-
-    private update() {
+    protected update() {
         this.panel.title = `Kubernetes logs ${this.resource}`;
         this.panel.webview.html = `
         <!doctype html>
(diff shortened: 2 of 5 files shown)


src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+        while (this.disposables.length) {
+            const x = this.disposables.pop();
+            if (x) {
+                x.dispose();
+            }
+        }
+    }
+
+    private update() {
+        this.panel.title = `Kubernetes logs ${this.resource}`;
```
itowlson: Logs with a capital L?  Perhaps also a separator?  To make it easy to scan I'd even suggest `Logs - ${this.resource}` so that less of it gets elided away when there are lots of tabs.

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+            <title>Kubernetes logs ${this.resource}</title>
+            <script
+                src="https://code.jquery.com/jquery-3.3.1.min.js"
+                integrity="sha256-FgpCb/KJQlLNfOu91ta32o/NMZxltwRo8QtmkMRdAu8="
+                crossorigin="anonymous"></script>
+        </head>
+        <body>
+            <div style='position: absolute; top: 15px; left: 75%'>
+                <select id='mode' style='margin-bottom: 5px' onchange='eval()'>
+                    <option value='include'>Show Matches</option>
```
itowlson: Capitalisation of options should be consistent - suggest sentence case (`Show matches` etc.).

itowlson: I feel like the interaction with the regex box is not 100% clear.  I'd suggest 5 options:

* `Show all log entries`
* `Show log entries matching pattern`
* `Show log entries not matching pattern`
* `Show log entries up to first match`
* `Show log entries after first match`

(or you could use a label `Show log entries` with options `All`, `Matching pattern`, `Not matching pattern`, etc.)

itowlson: Looking at your screenshots I worry this could overlap the first row of logs.  Another possible design would be toolbar style, with a bar at the top of the window reserved for these drop-downs, and the logs occupying the rest of the window a la:

```
+-------------------------------------------------+
| Show log entries: [...]  Match pattern: [...]   |
+-------------------------------------------------+
| 28.01.19 14:08:17 Printed l33t ASCII art        |
| ...                                             |
```

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+                    <option value='exclude'>Hide Matches</option>
+                    <option value='after'>Show logs after match</option>
+                    <option value='before'>Show logs before match</option>
+                </select>
+                <input type='text' id='regexp' onkeyup='eval()' placeholder='Regular Expression Filter' size='25'/>
```
itowlson: I'd suggest just Filter or Match Pattern or something like that, and have the mention of regex in a tooltip or help button.  Though I guess if you move to the toolbar design you'll have more space to play with; but still I personally tend to prefer terser labels.

src/components/kubectl/logs.ts
```diff
@@ -86,7 +88,15 @@ function getLogsForContainer(
         return;
     }
 
-    kubectl.invokeInSharedTerminal(cmd);
+    const resource = `${podSummary.namespace}/${podSummary.name}`;
+    const panel = LogsPanel.createOrShow('TODO', 'Loading \u25CC', resource);
+
+    kubectl.invokeAsync(cmd).then(
+        (result: ShellResult) => {
+            panel.setInfo(result.stdout, resource);
```
itowlson: Does this work with follow?  Surely `invokeAsync(...).then(...)` will show nothing until the shell invoke completes, and only _then_ send all logs (the whole of stdout) to the panel?  And `logs -f` will never complete without user intervention (which they can no longer do since there is no longer a terminal in play)?

itowlson: Shouldn't this check for nonzero exit codes?

src/components/logs/logsWebview.ts
itowlson: I'm a bit doubtful about `eval`-ing on key up - for one thing, it's quite possible for the regex to be in an invalid state at that point, and even if it is valid it could well be transient (the user in the middle of typing a pattern), so we could be filtering a lot of logs for nothing.  I'd suggest an `Apply Filter` button, or, if you want it to be automatic, a delay where it applies the filter once the user has not typed anything for half a second or a second or something like that.

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+            </div>
+            <div style='margin-top: 35px'>
+                <div id='content' style="overflow-y: scroll; width: 100%; height: 100%"></div>
+            </div>
+            <script>
+              var orig = \`${this.content}\`.split('\\n');
```
itowlson: Consider doing this in the LogsPanel rather than in the embedded page.  By having all this script in the HTML string, we lose any semblance of syntax checking.  In the cluster wizard I had buttons send messages to the host, which did all the fancy logic and then sent new content back to the webview, which kept the on-page script fairly simple.

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+              var eval = () => {
+                var regexp = $('#regexp').val().trim();
+                // TODO: This could really be improved to avoid the double loop and list construction.
+                if (regexp.length > 0) {
+                    var mode = $('#mode').val();
+                    var regex = new RegExp(regexp);
```
itowlson: Do we need error handling around this in case the user input is invalid regex text?

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+                    switch (mode) {
+                        case 'include':
+                            content = orig.filter((line) => regex.test(line));
+                            break;
+                        case 'exclude':
+                            content = orig.filter((line) => !regex.test(line));
+                            break;
+                        case 'before':
+                            content = [];
+                            for (var i = 0; i < orig.length; i++) {
```
itowlson: `for (const line of orig)`?

itowlson: Or use `findIndex` to find the index of the first matching line and then slice, though this will require no-match handling.

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+                                var line = orig[i];
+                                if (regex.test(line)) {
+                                    break;
+                                }
+                                content.push(line);
+                            }
+                            break;
+                        case 'after':
+                            var i = 0;
+                            for (; i < orig.length; i++) {
```
itowlson: Would it not be easier to use `findIndex`?

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+                                    break;
+                                }
+                            }
+                            content = orig.slice(i+1);
+                            break;
+                    }
+                } else {
+                    content = orig;
+                }
+                $('#content').html('<code><pre>' + content.join('\\n') + '</pre></code>');
```
itowlson: Do you need both `code` and `pre`?

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+            <script>
+              var orig = \`${this.content}\`.split('\\n');
+              var eval = () => {
+                var regexp = $('#regexp').val().trim();
+                // TODO: This could really be improved to avoid the double loop and list construction.
+                if (regexp.length > 0) {
+                    var mode = $('#mode').val();
+                    var regex = new RegExp(regexp);
+                    switch (mode) {
+                        case 'include':
```
itowlson: Does this pay any regard to multiline log entries e.g. if I have:

```
2019/01/28 14:37:00 Pod failed to start because:
* There was a fish in the percolator
* The owls were not what they seemed
* Fire walked with me
2019/01/28 14:37:02 Container restarted
```

and match `failed to start`, will I get the essential owl-based diagnostics or will I need to e.g. note the timestamp, remove the filter and scroll?

itowlson: Huh, 'go to timestamp' would be a really neat command for this.

src/components/kubectl/logs.ts
```diff
@@ -86,7 +88,15 @@ function getLogsForContainer(
         return;
     }
 
-    kubectl.invokeInSharedTerminal(cmd);
+    const resource = `${podSummary.namespace}/${podSummary.name}`;
+    const panel = LogsPanel.createOrShow('TODO', 'Loading \u25CC', resource);
```
itowlson: TODO?

itowlson: Consider `...` rather than the 'dotted circle' character?

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
+import * as path from 'path';
```
itowlson: Is `path` used?

src/components/logs/logsWebview.ts
itowlson: As would 'go to first/next match' so you could easily see the matched log in its context.

Perhaps we should have a command to open the logs in a normal editor window, so people could use all the fancy find features built into VS Code and into extensions (I saw there was at least one line filtering extension though I have no idea how well it works!).  Though that would still struggle with the Follow scenario.

squillace:
@itowlson I'm kinda with Brendan here, and that despite my desire to be against him. Why on earth is it a Kantian imperative to do all that extra stuff? I'm thrilled that bits and pieces of very useful stuff will appear that helps the overall devux. But doing the full operator is still a fairly hard boundary, right?

itowlson:
After a bit more nosing around, I don't think the extensions I thought might make working with logs as documents actually do what I hoped.  So maybe handling the display ourselves is indeed the way to go.

One thing I would suggest as a bridge is an "Open in Editor" command (which could go in the toolbar region if you adopt that suggestion), which simply dumps the log text into an editor window from where the user can save it, search with built-in facilities, delete stuff that is not of interest, etc.  That gives users a quick way to get over into editorland if that's what they want, while still putting the custom experience first.

Finally I still feel we need to support the 'Follow Logs' scenario, unless I'm misreading this PR and that's still going to the terminal.  Which would be fine - I'm happy to take this transition in steps.

brendandburns:
@itowlson thanks for the feedback. I'll add the "open in document" option and yes, follow logs is absolutely next on the list, I just wanted to get agreement on the added functionality before I did the work.

I'll address your detailed comments, and see how hard it is to add follow logs to this PR. It might be easier to do it in a follow on.

Thanks again for the thoughts and the feedback!

src/components/logs/logsWebview.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export class LogsPanel {
+    public static readonly viewType = 'vscodeKubernetesLogs';
+    public static currentPanels = [];
+
+    private disposables: vscode.Disposable[] = [];
+    private content: string;
+    private resource: string;
+
+    public static createOrShow(extensionPath: string, content: string, resource: string): LogsPanel {
```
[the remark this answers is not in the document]
brendandburns: done.

src/components/logs/logsWebview.ts
brendandburns: done.

src/components/logs/logsWebview.ts
brendandburns: done.

brendandburns: done.

brendandburns: done.

src/components/kubectl/logs.ts
brendandburns: Nope, not currently. I need to write that code.

src/components/logs/logsWebview.ts
brendandburns: In practice it seems to work pretty well. I do want it to be automatic.

src/components/logs/logsWebview.ts
brendandburns: Future features (but I agree...)

src/components/kubectl/logs.ts
brendandburns: removed.

brendandburns: done.

src/components/logs/logsWebview.ts
brendandburns: removed.

src/components/logs/logsWebview.ts
brendandburns: done.

brendandburns:
Comments addressed, new screen shot is attached.
logview

src/components/logs/logsWebview.ts
itowlson: Yes, okay

src/components/logs/logsWebview.ts
itowlson: _shakes head at HTML_

itowlson:
You've got a few CI errors due to a bunch of strictness options being turned on in `master`.  The CI logs don't seem to actually _list_ the warnings for some reason, but when I run `tsc -p .` locally I get:

```
src/components/describe/describeWebview.ts(2,1): error TS6133: 'Dictionary' is declared but its value is never read.
src/components/kubectl/logs.ts(9,1): error TS6133: 'host' is declared but its value is never read.
src/components/kubectl/logs.ts(11,1): error TS6133: 'ShellResult' is declared but its value is never read.
src/components/kubectl/logs.ts(96,13): error TS2532: Object is possibly 'undefined'.
src/components/kubectl/logs.ts(97,67): error TS2532: Object is possibly 'undefined'.
src/components/kubectl/logs.ts(99,27): error TS2532: Object is possibly 'undefined'.
src/components/webpanel/webpanel.ts(12,30): error TS7017: Element implicitly has an 'any' type because type 'Map<string, T>' has no index signature.
src/components/webpanel/webpanel.ts(17,13): error TS2322: Type 'undefined' is not assignable to type 'T'.
src/components/webpanel/webpanel.ts(27,9): error TS7017: Element implicitly has an 'any' type because type 'Map<string, T>' has no index signature.
src/components/webpanel/webpanel.ts(43,42): error TS6133: 'e' is declared but its value is never read.
src/components/webpanel/webpanel.ts(57,16): error TS7017: Element implicitly has an 'any' type because type 'Map<string, T>' has no index signature.
src/components/webpanel/webpanel.ts(69,24): error TS7010: 'update', which lacks return-type annotation, implicitly has an 'any' return type.
```

None of these will affect behaviour; I can fix them up before merge if you want.

itowlson:
I'm playing with the filter stuff and I've noticed some performance hiccups with large volumes of logs.  I created a `logspouter` pod for testing and it has perhaps been spouting too often for too long now... but the result is that when I switch from `Show log entries: all` to `that match`, it locks up for several seconds (the drop-down stays open and I can't click on the filter box).  Then when I type the first character of the filter, it freezes again.  Only as I type enough characters to filter the logs down to a more reasonable amount does it become responsive.

`logspouter` may well be unrepresentative of real k8s pod logs, as it was specifically created to emit logs often enough that I could experiment with tailing ideas - but then again a real pod might have lived _far_ longer than logspouter's 5 days.  You have more experience than me of the sort of volumes that are realistic, and if you feel that it's responsive enough with realistic data sets then I trust you!  But it's definitely slower than I would like in this particular test case.

(I can give you access to `logspouter` if you want to try it - just let me know.)

itowlson:
Okay sorry to keep leaving naggy comments but there is also a glitch if you switch away from the logs window.  When you switch back it appears _completely blank_ (no controls or `Loading...` message).  If, as with the `logspouter` test, it takes a long time to render, then there is no feedback during this time.

There is an option `retainContextWhenHidden` which appears to fix this - it does incur memory overhead, which is not exactly something VS Code needs more of, but worth it I feel....?  But again maybe it isn't going to be an issue for more realistic data sets.

itowlson:
If it helps to provide context, at the time I was playing with this stuff, `logspouter` had emitted about 186000 lines of logs with a total size of about 17MB.

itowlson:
Re the toolbar scrolling, I was playing with this for something else and this worked for me: https://stackoverflow.com/questions/9677894/header-div-stays-at-top-vertical-scrolling-div-below-with-scrollbar-only-attach.  The hardwired height is a bit of a cheat but we're not exactly writing a HTML style tutorial here... *grin*

brendandburns:
Thanks, yeah, I had the toolbar pinned before, and I agree, I'll figure out to pin it again.

And thanks for the performance reports, I'll try to repro and fix.

brendandburns:
Ok, so I put it under the electron performance microscope :) and it turns out that nearly all of the time is spent _rendering_, filtering is pretty fast, it's the actual pixel spitting that is slow.

I cleaned this up somewhat. Here's what I did:

* Moved the filtering to an async method, so the UX won't freeze
* Split the rendering up into batches via `setTimeout` so that the UX won't freeze and it will feel responsive (this actually works pretty slick, the only downside is that the scrollbar wiggles a little, but it's barely noticeable)
* Added [`retainContentWhenHidden`](https://code.visualstudio.com/api/extension-guides/webview#retaincontextwhenhidden) to the Webview so that the switching tabs thing doesn't happen.

At this point, I think it feels pretty good in terms of responsiveness.

Please take another look. I'll start a new PR to add support for following logs.

itowlson:
Performance is great now!  Thanks!

Unfortunately... (and apologies if this is something you're still working on)... the toolbar is now _pinned_ but the content scrolls underneath it.  The logs need to be in a separate scrolling `div`.

image

brendandburns:
Fixed. (though the length of time it took me to figure out how to do it is a clear sign I shouldn't be slinging CSS)

Please take another look...

itowlson:
Urgh, I am sorry to be a pain but... now the docking is working but the scroll bar is not visible...

image

...unless I horizontally scroll to it, which cuts off the beginning of the log lines...

image

The lines in the log file are not so long that they should require horizontal scrolling, and even if they were, it would be good if the vertical scroll bar were always visible...

(I have artifically ensmallened the VS Code window for screenshot purposes... if I embiggen it then I get a _different_ misbehaviour:

image

Here the scroll bar does not dock to the RHS of the window.  So it does feel like a div width issue.)

itowlson:
I did have a go at fixing the scrollbar positioning but have had no luck I'm afraid - I think it's tricky with the progressive loading...

brendandburns:
Yeah, I noticed that issue and hoped no one else would *sigh*

I'll work on fixing things some other way...

brendandburns:
@itowlson a miracle occurred and I was enlighted about CSS. Well, not actually, but I did figure out what was needed here (I think) please re-check!

Thanks

itowlson:
I understand miracles are a core element of the CSS design workflow.  Thanks, will take a look!

brendandburns:
Thanks (insert Neo/Matrix "I know CSS" meme here :)

I'll work on the follow logs implementation next.

PR #453, opened 2019-01-27
If we can't check compatibility, we shouldn't check if that's a compatibility issue

### LINKED ISSUE #450: Kubernetes extension consumes excessive CPU anytime a YAML file is open.
macOS mojave 10.14.2
vscode 1.30.2
Kubernetes extension 0.1.15

For the following tests, all unnecessary apps were fully closed.

### 1. vscode w/o kubernetes extension, *no* `yml` file open
CPU utilization is almost nothing.

screenshot 2019-01-24 at 09 29 29

### 2. vscode w/o kubernetes extension, `yml` file *open*
Again, CPU utilization is almost nothing.

screenshot 2019-01-24 at 09 32 15

### 3. vscode *with* kubernetes extension, *no* `yml` file open
Minimal additional CPU utilization.

screenshot 2019-01-24 at 09 33 47

### 4. vscode *with* kubernetes extension, `yml` file *open*
CPU utilization jumps to about 20%, and remains there even after `yml` file is closed. To return to normal CPU utilization vscode must be restarted.

screenshot 2019-01-24 at 09 35 33

I work with a lot of `yml` files for various applications, so I almost always have at least one open. This additional CPU usage has a significant impact on my battery life, so that I am forced to uninstall this extension.

itowlson:
Because that way madness lies.

We had a situation where if a `kubectl` command failed, we would call `kubectl version` to see if the failure might be down to a client-server mismatch.  But if `kubectl` itself errored in response to `kubectl version -o json` (for example, if `kubectl-path` pointed to a binary other than the true kubectl), then this counted as a failure that would trigger the client-server compatibility check.

As we run `kubectl config` on startup to find out what context we're in, this resulted in the extension going into an infinite loop on activation if confronted with an uncooperative kubectl.

We fix this by recording that the compatibility check has been initiated _before_ calling `kubectl version`, and not repeating it if the check has already been initiated.

Fixes #450.

### DIFF
--- a/src/kubectl.ts
+++ b/src/kubectl.ts
@@ -175,8 +175,15 @@ async function invokeAsync(context: Context, command: string, stdin?: string): P
     }
 }
 
+// TODO: invalidate this when the context changes or if we know kubectl has changed (e.g. config)
+let checkedCompatibility = false;  // We don't want to spam the user (or CPU!) repeatedly running the version check
+
 async function checkPossibleIncompatibility(context: Context): Promise<void> {
+    if (checkedCompatibility) {
+        return;
+    }
     const compat = await compatibility.check((cmd) => asJson<compatibility.Version>(context, cmd));
+    checkedCompatibility = true;
     if (!compatibility.isGuaranteedCompatible(compat) && compat.didCheck) {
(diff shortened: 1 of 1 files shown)


brendandburns:
LGTM.

PR #454, opened 2019-01-27
Enforce TypeScript 'no implicit any' option

itowlson:
As a small first step to embracing ALL THE STRICTS THE WHOLE JOLLY LOT.

I commented out the tests rather than de-any-ing them, because they're really not doing a lot for us at the moment.  I'd like to come back and figure out where we can make useful tests for specific components, and whether we can get any testing around the higher-level application logic which is a bit gnarly in places... but now is not that time.


brendandburns:
LGTM.

PR #468, opened 2019-02-12
Update to latest VS Code engine and Node 10 types

itowlson:
The latest version of VS Code runs on Node 10.  This updates our Node type declarations to v10 to better check compatibility (though it looks like we were okay).

It looks like the CI was already using Node 10, so I don't think we need to update anything there.

Creating this PR now to allow CI to noodle on it.  @bnookala would you mind taking a look to see if you think I've overlooked anything please?  But let's not merge until the latest VS Code has been out there a little longer.  Thanks!

PR #471, opened 2019-02-13
Kubectl per cluster version

### LINKED ISSUE #385: Per-cluster kubectl version
Possibly using this spiffy tool from @jakepearson: https://github.com/jakepearson/k

A user might have Kubernetes clusters of multiple versions, but we always use the `kubectl` on the path or configured location.  If these are too far out of sync, the user may see errors.  We try to detect and warn if this is a possible cause of an error, but this is putting the ambulance at the bottom of the cliff.  So it would be great if we could detect the server version and automatically use the right version of `kubectl`.

itowlson:
Some customers - specifically the critical customer demographic known as _me_ - have clusters from multiple Kubernetes versions.  It is not always possible to find a version of `kubectl` that works with all clusters, because `kubectl` is only guaranteed to work up to 2 API versions away.  For example, if I use my 1.9 kubectl then I can't see pods in the 1.7 cluster, but if I use my 1.7 kubectl then I can't deploy CRDs into the 1.9 cluster.

This PR adds an _opt-in_ option by which, instead of users specifying a `kubectl` binary, we inspect the cluster, download the matching `kubectl` binary and use that.  This is opt-in because I'm worried about performance impact; later we could change it to opt-out so as to make it more transparent.  Cluster versions and kubectl binaries are cached and persisted to minimise delays when switching clusters.  However, we incur a `kubectl config` call on _every_ command, as we need to check the current context name, and this may be non-trivial.  In future perhaps we could cache that, invalidating if the user switches context through the VS Code UI, and re-evaluate if we get an error to handle the 'changing cluster behind our back' scenario.

At the moment, the user still needs to provide a `kubectl` binary to bootstrap the process (because you need a kubectl to call `kubectl version`).  It would be good to remove this requirement and quietly download the latest `kubectl` to act as bootstrapper if no other kubectl can be found.

Fixes #385.

### DIFF
--- a/src/components/download/download.ts
+++ b/src/components/download/download.ts
@@ -3,12 +3,22 @@ import * as stream from 'stream';
 import * as tmp from 'tmp';
 
 import { succeeded, Errorable } from '../../errorable';
+import { Dictionary } from '../../utils/dictionary';
+import { sleep } from '../../sleep';
 
 type DownloadFunc =
     (url: string, destination?: string, options?: any)
          => Promise<Buffer> & stream.Duplex; // Stream has additional events - see https://www.npmjs.com/package/download
 
-let download: DownloadFunc;
+let download: DownloadFunc | undefined;
+
+const DOWNLOAD_ONCE_STATUS = Dictionary.of<DownloadOperationStatus>();
+
+enum DownloadOperationStatus {
+    Queued = 1,
+    Completed = 2,
+    Failed = 3,
+}
 
 function ensureDownloadFunc() {
     if (!download) {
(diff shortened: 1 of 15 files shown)


src/components/download/download.ts
```diff
@@ -35,9 +45,29 @@ export async function toTempFile(sourceUrl: string): Promise<Errorable<string>>
...
+
+export async function once(sourceUrl: string, destinationFile: string): Promise<Errorable<null>> {
+    const downloadStatus = DOWNLOAD_ONCE_STATUS[destinationFile];
+    if (!downloadStatus || downloadStatus === DownloadOperationStatus.Failed) {
+        DOWNLOAD_ONCE_STATUS[destinationFile] = DownloadOperationStatus.Queued;
+        const result = await to(sourceUrl, destinationFile);
+        DOWNLOAD_ONCE_STATUS[destinationFile] = succeeded(result) ? DownloadOperationStatus.Completed : DownloadOperationStatus.Failed;
+        return result;
+    } else {
+        while (true) {
```
brendandburns: I feel like this could have been done with an event handler somwhere instead of this while loop?

brendandburns:
This feels like a number of different refactors smashed into one PR, if that's really the only way to do it, then I guess it's ok, but it is somewhat hard to review as is.

itowlson:
Hmm, there are half a dozen or so lines that address the updated Node typings.  I'm pretty sure everything else is driven directly by the kubectl-per-cluster work; there's not a whole lot of refactoring going on, other than moving some stuff that is now needed in multiple places.  I will pull out the typings stuff but I don't think it will make much difference to the size or complexity of the PR; were there other things that you felt could have been pulled out into refactoring PRs?

Really, I think the trouble is that this feature requires us to touch multiple areas (config, download and kubectl invocation), some of which are already quite messy!  Perhaps I _should_ have done a refactor first...!

itowlson:
All the typings changes and consequent changes are now in #468.  This PR still needs to incorporate those commits but once that lands I can rebase this and it should be feature-only.  I think you can see only the diffs from the base PR via https://github.com/Azure/vscode-kubernetes-tools/pull/471/files/fa90a2f5496e2055c83d50916da72e5b3347d7c1..0b93740f5f3c88e42d319ec417033d66a4bc1a79

src/components/download/download.ts
itowlson: I'm not sure.  It needs to be something `await`able unless I rework the whole thing into callback style.  I guess the map could contain a promise which the `else` clause could `await`... but it feels like it could get fiddlier than the loop is... I agree the loop does not spark joy though...

itowlson:
Rebased, should now contain only feature stuff; please take another look.  Thanks, and do let me know if you still feel there are good seams along which to split it!

PR #472, opened 2019-02-14
Run CronJob immediately

### LINKED ISSUE #469: CronJob - add "Run Now" action
I typically have a Code task to do this,

```
{
            "label": "cluster: run datapump now",
            "type": "shell",
            "command": "kubectl create job --from=cronjob/datapump-nightly datapump-manual --namespace ${input:environment}",
            "problemMatcher": []
        },
```

which I had to spend some cycles to go look up how to do in the k8s documentation.  I think this would:

1. be of great utility in the context menu
2. further help new people better learn the `kubectl` commands by showing them real world examples as this outputs to the terminal window. (I know this is how I personally have gained a lot of know-how)

I would expect the task to be re-runnable (delete & create or however it can be accomplished)

itowlson:
This adds the functionality to run a CronJob immediately (that is, to create a 'normal' job from a CronJob).

Fixes #469.

~~(ETA: in order to avoid having to duplicate menu contributions for a new `vsKubernetes.resource.cronjob` tree node context, I switched over the menu contributions in `package.json` to use regular expressions on the context when clause.  This is orthogonal to the functional changes.  I can unpick it into a separate PR if it makes this one hard to read.)~~  This PR depends on #473 which separates out the refactoring of the explorer command contributions.  Once #473 is merged I can rebase this and it will reduce the noise.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -196,6 +196,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<extens
         registerCommand('extension.vsKubernetesAddFile', (explorerNode: explorer.KubernetesDataHolderResource) => { addKubernetesConfigFile(kubectl, explorerNode, treeProvider); }),
         registerCommand('extension.vsKubernetesShowEvents', (explorerNode: explorer.ResourceNode) => { getEvents(kubectl, EventDisplayMode.Show, explorerNode); }),
         registerCommand('extension.vsKubernetesFollowEvents', (explorerNode: explorer.ResourceNode) => { getEvents(kubectl, EventDisplayMode.Follow, explorerNode); }),
+        registerCommand('extension.vsKubernetesCronJobRunNow', cronJobRunNow),
         // Commands - Helm
         registerCommand('extension.helmVersion', helmexec.helmVersion),
         registerCommand('extension.helmTemplate', helmexec.helmTemplate),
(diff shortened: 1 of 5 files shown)


itowlson:
I also added the logs commands to the Job right-click menu, which is again technically orthogonal but fits with how the customer described the desired workflow around this feature.

src/utils/containercontainer.ts
```diff
@@ -0,0 +1,40 @@
...
+import { PodSummary } from "../extension";
+
+export interface ContainerContainer {
+    readonly kindName: string;
+    readonly namespace: string | undefined;
+    readonly containers?: Container[];
+    readonly containersQueryPath: string;
+}
+
+export module ContainerContainer {
```
brendandburns: This is a horrible name :P It at least deserves a descriptive comment or a better name (or both :)

brendandburns:
As with the other PR, this feels like a refactor and a feature smashed into one. Any chance we can split them out?

itowlson:
Guilty as charged - this one definitely contains some refactoring that could be reviewed independently of the feature work.  I'll split it out.  Thanks!

src/utils/containercontainer.ts
itowlson: You would not _believe_ how many people I begged for naming suggestions... and yet somehow they were _all even worse than this_.  I originally had a comment mentioning it but removed it as too self-indulgent!  I'll have another go at coming up with something better, or failing that at least put in a comment to explain what it's for.

itowlson:
@brendandburns I've broken out the other functionality into separate PRs, but this one depends on the command refactor, so I still have to include that commit in this PR.  Once the command refactor lands in master, I can rebase this one (hey, it might even rebase itself) so it's cleaner.

itowlson:
Rebased; please take another look.  Thanks!

src/extension.ts
```diff
@@ -2007,3 +2009,57 @@ async function kubernetesLint(document: vscode.TextDocument): Promise<void> {
     const diagnostics = ([] as vscode.Diagnostic[]).concat(...linterResults);
     kubernetesDiagnostics.set(document.uri, diagnostics);
 }
+
+async function cronJobRunNow(target?: any): Promise<void> {
```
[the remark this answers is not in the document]
itowlson: @bnookala Not sure what you're getting at here... we could type it as `string | ResourceNode | undefined`, but I feel like for these commands that can take parameters we've generally typed as `any`, because we can't statically verify where the command might be called from, and (admittedly inconsistently) checked expectations within the handler body.

Or were you suggesting we should switch to typing command handler args as `unknown` instead of `any`?

PR #473, opened 2019-02-17
Consolidate explorer right-click commands using context regexes

itowlson:
VS Code now supports menu contribution `when` clauses using regexes.  This means we can give each k8s resource node a distinctive context (e.g. `vsKubernetes.resource.deployment`), but apply the common commands to them via a pattern.  This reduces duplication (and errors, and inconsistencies!) compared to having to apply commands to every resource type individually.


brendandburns:
LGTM.

PR #480, opened 2019-02-24
API proposal

itowlson:
A reworked proposal for an API design.

### DIFF
--- a/docs/extending/README.md
+++ b/docs/extending/README.md
@@ -13,7 +13,8 @@ a particular version of the component API. Versioned interfaces are immutable, a
 versions should continue to work even after a new version is available, though old versions
 may eventually be retired.
 
-The easiest way to get an interface is to use the [vscode-kubernetes-tools-api NPM package](https://link/tba).
+The easiest way to get an interface is to use the [vscode-kubernetes-tools-api NPM
+package](https://www.npmjs.com/package/vscode-kubernetes-tools-api).
 Using this, you can request interfaces via convenient accessors:
 
 ```javascript
--- a/docs/extending/README.md
+++ b/docs/extending/README.md
@@ -1,0 +1,61 @@
+# Extensibility Overview
+
+## Using APIs
+
+Visual Studio Code Kubernetes Tools offers APIs for integrating with the extension.
+Currently the only API available is for writing cluster providers.  If you would like an
+API for a feature which isn't currently surfaced, please post an issue on GitHub!
+
+### Getting an API object
+
+To access an API, you request an interface for the component you want to access, specifying
+a particular version of the component API. Versioned interfaces are immutable, and existing
+versions should continue to work even after a new version is available, though old versions
+may eventually be retired.
+
+The easiest way to get an interface is to use the [vscode-kubernetes-tools-api NPM package](https://link/tba).
+Using this, you can request interfaces via convenient accessors:
+
+```javascript
+import * as k8s from 'vscode-kubernetes-tools-api';
+
+export async function activate(context: vscode.ExtensionContext) {
+    const cp = await k8s.extension.clusterProvider.v1;  // requests v1 of the cluster provider API
+}
+```
+
+The accessor returns a promise, which resolves to one of the following:
+
+* If the requested interface is available, an object with `available: true` and an `api`
+  property containing the interface itself.
+* If the requested interface is not available, an object with `available: false` and a `reason`
+  property indicating why not. You can use the `reason` value to display appropriate user
+  interface or to degrade gracefully.
+
+(In TypeScript, these are represented by the `APIAvailable<T>` and `APIUnavailable` types.)
+
+You can also request an interface using the VS Code extensions API:
+
+```javascript
+const extension = vscode.extensions.getExtension<APIBroker>(MS_KUBERNETES_EXTENSION_ID);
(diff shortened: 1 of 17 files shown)


itowlson:
I really want to push ahead with this.  @squillace is it too soon?  Do you want to get more partner input?  I worry that if we procrastinate too long it will become too late...

docs/extending/README.md
```diff
@@ -0,0 +1,61 @@
...
+API for a feature which isn't currently surfaced, please post an issue on GitHub!
+
+### Getting an API object
+
+To access an API, you request an interface for the component you want to access, specifying
+a particular version of the component API. Versioned interfaces are immutable, and existing
+versions should continue to work even after a new version is available, though old versions
+may eventually be retired.
+
+The easiest way to get an interface is to use the [vscode-kubernetes-tools-api NPM package](https://link/tba).
```
bnookala: im sure you're aware, but this package isn't published yet

docs/extending/clusterprovider.md
```diff
@@ -0,0 +1,296 @@
+# Adding Cloud or Cluster Types to the Cluster Wizard
+
+The Kubernetes extension provides the Add Existing Cluster command to add a
```
bnookala: perhaps place 'Add Existing Cluster' in backticks, since it describes a command usage

docs/extending/clusterprovider.md
```diff
@@ -0,0 +1,296 @@
...
+the `next` method.  (Pages that end the wizard - that is, that do not need to be followed by
+a call to `next` - do not need to follow this convention.)
+
+* Your page must contain a HTML form.
+* The `id` of the form must be the `WIZARD_FORM_NAME` constant defined in the NPM package.
+* The form must contain a hidden element whose name is the `CLUSTER_TYPE_KEY` constant
+  and whose value is the cluster provider's `id` property.
+* The form must contain _all_ information that you want to keep.  You cannot store state
+  in the provider between pages - you _must_ pass it from page to page.  (For example,
+  if you first page asks for a region ID and your second page for a name, then
```
bnookala: 'if your first page asks'

docs/extending/clusterprovider.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Activating the cluster provider extension
+
+Your extension needs to activate in response to the Add Existing Cluster and/or Create
+Cluster commands, so that it can register its cluster provider(s) before the wizard is
+displayed.  To do this, your `package.json` must include the following activation events:
+
+```json
+    "activationEvents": [
+        "onCommand:extension.vsKubernetesConfigureFromCluster",
```
bnookala: interesting - i had no idea an extension could shadow activation events!

docs/extending/samples/clusterprovider/contosocloud.ts
```diff
@@ -0,0 +1,139 @@
+/*
```
bnookala: is this entire file meant to be commented?

src/api/contract/clusterprovider/v1.ts
```diff
@@ -0,0 +1,34 @@
+// This module is contractual and should not be changed after release.
```
bnookala: any reason why these contract files are baked into the extension rather than being distributed by the api node module?

docs/extending/README.md
itowlson: Good point.  Added `do-not-merge` label back in!

docs/extending/clusterprovider.md
itowlson: Yeah, the activation event doesn't have to be from your own extension - I think VS Code just has global collections of commands, views, etc. and anything that knows the right name can take an interest in them.

docs/extending/samples/clusterprovider/contosocloud.ts
itowlson: Yes... if I don't comment it, then it causes build warnings on the extension which is undesirable (and no, excluding the directory doesn't seem to help).  I think I need to move samples out to a separate samples repo which can depend on the NPM package - thoughts?

src/api/contract/clusterprovider/v1.ts
itowlson: The extension needs copies of the contracts so it can check its implementation conforms to the contracts.  It could do this by importing the NPM module, but then we'd have a weird sort-of-circular-dependency situation which, although it _could_ be made to work, would be wearisome.  (Think of the workflow where you'd have to write and pack the NPM module _but not publish it_ then amend the reference in the extension to the module TGZ, then implement the contracts, then amend the reference back to an _unpublished_ version of the NPM module, push and merge the extension in a nonbuilding state (because it couldn't find the NPM module with the updated contracts), then finally publish the NPM module.)

squillace:
@bnookala I'd love to see you map your extension to this api and see what isn't there. IIRC, there were a few things @itowlson mentioned he hadn't exposed yet.

itowlson:
@bnookala I think this might be ready for merging - do you want to take another look?

itowlson:
I'm going to merge this to unblock other API work in progress.  If it turns out to be a disaster then we can always ~~change the committer so it looks like Ralph did it~~ amend or remove it before going to release.

PR #483, opened 2019-02-27
Add a refresh button for 'describe'

### LINKED ISSUE #464: Describe command could use an auto refresh option
Great work on the new updates, it's been a huge productivity win!

I really like the new describe how it outputs to the editor window vs the terminal.

When I'm troubleshooting container start ups, I find myself re-clicking describe over and over.  It would be great if there was a capability to auto refresh describe on a specified interval (it would stop once the editor window was closed by the user)

Could be accomplished either by:
- User config (`describeAutoRefresh & `describeAutoRefreshInterval`)
- Separate context menu (Describe - auto refresh)

brendandburns:
Fixes #464 

@itowlson

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
         "onCommand:extension.vsKubernetesDelete",
         "onCommand:extension.vsKubernetesDeleteNow",
         "onCommand:extension.vsKubernetesDeleteUri",
+        "onCommand:extension.vsKubernetesDescribe.Refresh",
         "onCommand:extension.vsKubernetesApply",
         "onCommand:extension.vsKubernetesApplyFile",
         "onCommand:extension.vsKubernetesExplain",
@@ -267,6 +268,12 @@
             ]
         },
         "menus": {
+            "editor/title": [
+                {
+                    "command": "extension.vsKubernetesDescribe.Refresh",
+                    "when": "vscodeKubernetesDescribeContext"
+                }
+            ],
             "explorer/context": [
                 {
                     "title": "Update Dependencies",
@@ -565,6 +572,11 @@
             ]
         },
         "commands": [
+            {
+                "command": "extension.vsKubernetesDescribe.Refresh",
+                "title": "Refresh",
+                "category": "Kubernetes"
+            },
             {
                 "command": "extension.vsKubernetesCreate",
                 "title": "Create",
@@ -907,6 +919,14 @@
                 "title": "Show Dependencies",
                 "description": "List the dependencies of a Helm chart",
                 "category": "Helm"
+            }
+        ],
+        "keybindings": [
+			{
+				"command": "extension.vsKubernetesDescribe.Refresh",
+				"key": "shift+ctrl+r",
+				"mac": "shift+cmd+r",
+				"when": "vscodeKubernetesDescribeContext"
             }
         ],
         "languages": [
--- a/src/components/describe/describeWebview.ts
+++ b/src/components/describe/describeWebview.ts
@@ -1,13 +1,22 @@
 import * as vscode from 'vscode';
+import { ShellResult } from '../../shell';
 import { WebPanel } from '../webpanel/webpanel';
 
 export class DescribePanel extends WebPanel {
     public static readonly viewType = 'vscodeKubernetesDescribe';
+    private static readonly describeContext = 'vscodeKubernetesDescribeContext';
+    private static activePanel: DescribePanel | undefined = undefined;
     public static currentPanels = new Map<string, DescribePanel>();
 
-    public static createOrShow(content: string, resource: string) {
+    public static refreshCommand() {
+        if (DescribePanel.activePanel) {
+            DescribePanel.activePanel.doRefresh();
+        }
+    }
+
+    public static createOrShow(content: string, resource: string, refresh: () => Promise<ShellResult | undefined>) {
         const fn = (panel: vscode.WebviewPanel, content: string, resource: string): DescribePanel => {
-            return new DescribePanel(panel, content, resource);
+            return new DescribePanel(panel, content, resource, refresh);
         };
         WebPanel.createOrShowInternal<DescribePanel>(content, resource, DescribePanel.viewType, "Kubernetes Describe", DescribePanel.currentPanels, fn);
     }
@@ -15,9 +24,46 @@ export class DescribePanel extends WebPanel {
     private constructor(
         panel: vscode.WebviewPanel,
         content: string,
-        resource: string
+        resource: string,
+        readonly refresh: () => Promise<ShellResult | undefined>
     ) {
         super(panel, content, resource, DescribePanel.currentPanels);
+
+        const setActiveContext = (active: boolean) => {
+            vscode.commands.executeCommand('setContext', DescribePanel.describeContext, active);
+            DescribePanel.activePanel = active ? this : undefined;
+        };
+
+        this.panel.onDidChangeViewState((evt: vscode.WebviewPanelOnDidChangeViewStateEvent) => {
+            setActiveContext(evt.webviewPanel.active);
+        });
+        setActiveContext(true);
+        this.panel.webview.onDidReceiveMessage(
+            async (message) => {
+                switch (message.command) {
+                    case 'refresh':
+                        await this.doRefresh();
+                        break;
+                }
+            },
+            undefined
+        );
+    }
+
+    protected async doRefresh() {
+        const result = await this.refresh();
+        if (!result) {
+            vscode.window.showErrorMessage('Error refreshing!');
+            return;
+        }
+        if (result.code !== 0) {
+            vscode.window.showErrorMessage(`Error refreshing: ${result.stderr}`);
+            return;
+        }
+        this.panel.webview.postMessage({
+            command: 'content',
+            content: result.stdout,
+        });
     }
 
     protected update() {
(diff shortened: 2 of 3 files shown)


brendandburns:
I'll rebase this....

brendandburns:
rebased and ready for review.

src/components/describe/describeWebview.ts
```diff
@@ -1,23 +1,51 @@
...
+        cmd: string,
+        kubectl: Kubectl
     ) {
         super(panel, content, resource, DescribePanel.currentPanels);
+
+        this.panel.webview.onDidReceiveMessage(
+            (message) => {
+                switch (message.command) {
+                    case 'refresh':
+                        kubectl.invokeAsync(cmd).then((result: ShellResult | void) => {
```
itowlson: nitpick:
```suggestion
                        kubectl.invokeAsync(cmd).then((result: ShellResult | undefined) => {
```
(you can probably even leave out the type annotation - I think TypeScript should infer it)

src/components/describe/describeWebview.ts
```diff
@@ -28,11 +56,51 @@ export class DescribePanel extends WebPanel {
...
+                    setTimeout(() => {
+                        requestRefresh();
+                        automaticRefresh();
+                    }, val * 1000);
+                }
+            };
+        </script>
     </head>
     <body>
+        <div style="position: absolute; left: 90%">
```
itowlson: This scrolls off the top if the `describe` output is long enough to merit scrolling.  It should probably be fixed... maybe a toolbar-style design like the one for logs?  (Though that does make the option very in-your-face for what might be quite a niche usage.)

src/components/describe/describeWebview.ts
```diff
@@ -28,11 +56,51 @@ export class DescribePanel extends WebPanel {
...
+          <span><a href="javascript:void;" onClick="requestRefresh();">Refresh</a></span>
```
itowlson: This ends up looking a bit odd to me:

image

The vertical positioning seems a bit tight, and the drop-down lacks context.  Again, a toolbar might provide a bit more breathing room.

src/components/describe/describeWebview.ts
```diff
@@ -1,23 +1,51 @@
...
+import { Kubectl } from '../../kubectl';
+import { ShellResult } from '../../shell';
 import { WebPanel } from '../webpanel/webpanel';
 
 export class DescribePanel extends WebPanel {
     public static readonly viewType = 'vscodeKubernetesDescribe';
     public static currentPanels = new Map<string, DescribePanel>();
 
-    public static createOrShow(content: string, resource: string) {
+    public static createOrShow(content: string, resource: string, cmd: string, kubectl: Kubectl) {
```
itowlson: Style suggestion: instead of passing in these objects and asking the view to compose the kubectl invocation, consider passing in a `refreshCallback: () => Promise<Errorable<string>>` (or `Promise<ShellResult | undefined>` if you prefer).  Having the mechanics of description retrieval split across this module and its caller (for the original content) feels a bit awkward to me.

Alternatively, you could make this module _entirely_ responsible for populating the description - so it would use `cmd` and `kubectl` to compose the initial content as well as the refreshed content.  You might even be able to use the same mechanism to inject both the initial content and the refreshed content into the view.

src/components/describe/describeWebview.ts
brendandburns: removed.

src/components/describe/describeWebview.ts
brendandburns: removed.

src/components/describe/describeWebview.ts
brendandburns: done.

brendandburns:
@itowlson I moved this into the editor/title menu, unfortunately didn't see a good way to make it automatically refresh after an interval.

I'll explore more, but this is a step forward anyway.  ptal.

Thanks!

brendandburns:
@itowlson rebased, please take another look.

src/components/describe/describeWebview.ts
```diff
@@ -1,23 +1,69 @@
...
         };
         WebPanel.createOrShowInternal<DescribePanel>(content, resource, DescribePanel.viewType, "Kubernetes Describe", DescribePanel.currentPanels, fn);
     }
 
     private constructor(
         panel: vscode.WebviewPanel,
         content: string,
-        resource: string
+        resource: string,
+        readonly refresh: () => Promise<ShellResult | undefined>
```
itowlson: `private`?
```suggestion
        private readonly refresh: () => Promise<ShellResult | undefined>
```

src/components/describe/describeWebview.ts
```diff
@@ -1,23 +1,69 @@
...
+                    case 'refresh':
+                        await this.doRefresh();
+                        break;
+                }
+            },
+            undefined
+        );
+    }
+
+    protected async doRefresh() {
```
itowlson: Why is this `protected`?  Looks like it could be `private`, or does that stop it being called by the static method somehow?

src/components/describe/describeWebview.ts
```diff
@@ -28,11 +74,43 @@ export class DescribePanel extends WebPanel {
...
+                const message = event.data;
+                switch (message.command) {
+                    case 'content':
+                        console.log('refresh');
+                        const elt = document.getElementById('content');
+                        elt.innerText = message.content;
+                }
+            });
+
+            const automaticRefresh = () => {
```
itowlson: Is this still used?

src/components/describe/describeWebview.ts
```diff
@@ -28,11 +74,43 @@ export class DescribePanel extends WebPanel {
     <head>
         <meta charset="UTF-8">
         <title>Kubernetes describe ${this.resource}</title>
+        <script>
+            const vscode = acquireVsCodeApi();
+            const requestRefresh = () => {
```
itowlson: Is this still used?

src/components/describe/describeWebview.ts
```diff
@@ -28,11 +74,43 @@ export class DescribePanel extends WebPanel {
...
+                vscode.postMessage({
+                    command: 'refresh'
+                });
+            };
+
+            window.addEventListener('message', event => {
+                const message = event.data;
+                switch (message.command) {
+                    case 'content':
+                        console.log('refresh');
```
itowlson: I'd suggest removing this now.

src/components/describe/describeWebview.ts
```diff
@@ -28,11 +74,43 @@ export class DescribePanel extends WebPanel {
...
+                }
+            };
+        </script>
     </head>
     <body>
         <code>
-            <pre>${this.content}</pre>
+            <pre id='content'>${this.content}</pre>
         </code>
+        <script>
```
itowlson: Can this script block now be removed?

src/extension.ts
```diff
@@ -1180,17 +1181,25 @@ function getPorts() {
 async function describeKubernetes(explorerNode?: explorer.ResourceNode) {
     if (explorerNode) {
         const nsarg = explorerNode.namespace ? `--namespace ${explorerNode.namespace}` : '';
-        const result = await kubectl.invokeAsync(`describe ${explorerNode.resourceId} ${nsarg}`);
+        const cmd = `describe ${explorerNode.resourceId} ${nsarg}`;
+        const result = await kubectl.invokeAsync(cmd);
+        const refresh = (): Promise<ShellResult | undefined> => {
```
itowlson: This syntax looks a bit odd to me.  Should it not be `const refresh = () => { … }` or `const refresh: () => Promise<ShellResult | undefined> = () => { … }`?  The way it's laid out looks like the Promise is a type annotation on `()`.  Or perhaps this is another TypeScript idiosyncrasy I don't yet know!

src/extension.ts
```diff
@@ -1180,17 +1181,25 @@ function getPorts() {
...
+            return kubectl.invokeAsync(cmd);
```
itowlson: You can probably simplify `=> { return kubectl.invokeAsync(…); }` to `=> kubectl.invokeAsync(…)` (i.e. lose the braces and the `return`).

package.json
```diff
@@ -565,6 +572,11 @@
             ]
         },
         "commands": [
+            {
+                "command": "extension.vsKubernetesDescribe.Refresh",
+                "title": "Refresh",
```
itowlson: This might be confusing because it specifically refreshes the `Describe` window.  It is probably okay if we set up a `commandPalette` entry so that it's only shown when a describe window is active.

src/components/describe/describeWebview.ts
brendandburns: made private.

src/components/describe/describeWebview.ts
brendandburns: removed.

src/components/describe/describeWebview.ts
brendandburns: removed.

src/components/describe/describeWebview.ts
brendandburns: removed.

src/components/describe/describeWebview.ts
brendandburns: removed.

src/extension.ts
brendandburns: What I was trying to say was:

"refresh is a function with no parameters which returns a Promise<...>" and I _think_ that's what Typescript sees... Happy to adjust if you want.

src/extension.ts

package.json
brendandburns: I always am unclear about how to do this, suggestions welcome...

brendandburns:
Comments mostly addressed, please take another look.

Thanks!

package.json
itowlson: Yeah, I haven't found a great solution.  You can either:

* Make the title very explicit: `Refresh Active Description`
* Add a `contributes/menus/commandPalette` entry with a `when` clause of `context == describePanel`  (or something like that), so that the `Refresh` entry doesn't appear in the palette unless the user is looking at a description
* Use a category: `"category": "Kubernetes Describe", "title": "Refresh"` - when seen in the command palette it will be qualified by the category, disambiguating it from any other Refresh commands

PR #484, opened 2019-02-27
When deleting context, press on even if cluster or user didn't delete

### LINKED ISSUE #482: Delete context fails because cluster already deleted
With two contexts pointing at the same cluster, if you delete one context it will delete the cluster. Then attempting to delete the second context fails because the cluster doesn't exist. Result: Must manually hack the config.

image

Build: January 2019.

itowlson:
If two contexts point to the same cluster, deleting both would fail because the first would remove the underlying cluster and user from the kubeconfig file, and the second would then complain because trying to delete the cluster failed.  Because one of these prior deletions can fail on 'not in kubeconfig', we should press on to try to delete the context even if cluster or user deletion fails.

A smart fix would check if the reason for failure was that the cluster (or user) didn't exist.  This is not a smart fix.

Fixes #482.

### DIFF
--- a/src/kubectlUtils.ts
+++ b/src/kubectlUtils.ts
@@ -105,29 +105,27 @@ export async function getContexts(kubectl: Kubectl): Promise<KubectlContext[]> {
     });
 }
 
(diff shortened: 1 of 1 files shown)


brendandburns:
LGTM.

PR #486, opened 2019-03-02
Add logic to load and suggest ports from the container.

### LINKED ISSUE #261: Detect ContainerPorts on the Pod for Port Forwarding
If a Pod explicitly exposes ports in its definition, we should offer a way for our users to Port Forward to them without having to dig for those ports. 

Example pod yaml that exposes ports:

```yaml
apiVersion: apps/v1 # for versions before 1.9.0 use apps/v1beta2
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2 # tells deployment to run 2 pods matching the template
  template: # create pods using pod definition in this template
    metadata:
      # unlike pod-nginx.yaml, the name is not included in the meta data as a unique name is
      # generated from the deployment name
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.7.9
        ports:
        - containerPort: 80
```

brendandburns:
Fixes #261

### DIFF
--- a/src/components/kubectl/port-forward.ts
+++ b/src/components/kubectl/port-forward.ts
@@ -105,34 +105,46 @@ export async function portForwardKubernetes (kubectl: Kubectl, explorerNode?: an
 }
 
 /**
+ * Given a JSON representation of a Pod, extract the ports to suggest to the user for
+ * for port forwarding.
+ */
+function extractPodPorts(podJson: string): string | undefined {
+    const pod = JSON.parse(podJson) as kubernetes.V1Pod;
+    const containers = pod.spec.containers;
+    const ports = Array.of<number>();
+    containers.forEach((container) => {
+        if (container.ports) {
+            const containerPorts = container.ports.map((port) => port.containerPort);
+            ports.push(...containerPorts);
+        }
+    });
+    if (ports.length > 0) {
+        const portPairs = ports.map((port) => `${port}:${port}`);
+        return portPairs.join(' ');
+    }
+    return;
+}
+
+/**
  * Prompts the user on what port to port-forward to, and validates numeric input.
  * @returns An array of PortMapping objects.
  */
 async function promptForPort (kubectl?: Kubectl, podName?: string, namespace?: string): Promise<PortMapping[]> {
     let portString: string | undefined;
-    let defaultValue = undefined;
+    let defaultValue: string | undefined = undefined;
     if (podName && kubectl) {
         const ns = namespace || 'default';
         try {
             const result = await kubectl.invokeAsync(`get pods ${podName} --namespace ${ns} -o json`);
-            if (result && result.code === 0) {
-                const pod = JSON.parse(result.stdout) as kubernetes.V1Pod;
-                const containers = pod.spec.containers;
-                const ports = [] as number[];
-                containers.forEach((container: kubernetes.V1Container) => {
-                    if (container.ports) {
-                        container.ports.forEach((port: kubernetes.V1ContainerPort) => {
-                            ports.push(port.containerPort);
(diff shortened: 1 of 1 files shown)


src/components/kubectl/port-forward.ts
```diff
@@ -90,23 +96,50 @@ export async function portForwardKubernetes (kubectl: Kubectl, explorerNode?: an
...
+    let defaultValue = undefined;
+    if (podName && kubectl) {
+        const ns = namespace || 'default';
+        try {
+            const result = await kubectl.invokeAsync(`get pods ${podName} --namespace ${ns} -o json`);
+            if (result && result.code === 0) {
+                const pod = JSON.parse(result.stdout) as kubernetes.V1Pod;
+                const containers = pod.spec.containers;
+                const ports = [] as number[];
+                containers.forEach((container: kubernetes.V1Container) => {
```
itowlson: Do you need the type declaration here?  TypeScript is usually pretty good at inferring this (assuming it's inferred `containers` `Container[]` rather than `any`!).

src/components/kubectl/port-forward.ts
```diff
@@ -90,23 +96,50 @@ export async function portForwardKubernetes (kubectl: Kubectl, explorerNode?: an
...
+                    if (container.ports) {
+                        container.ports.forEach((port: kubernetes.V1ContainerPort) => {
```
itowlson: Again is the type annotation needed?

itowlson: Consider instead:

```
ports.push(...container.ports.map((p) => p.containerPort));
```

or to be more explicit

```
const containerPorts = container.ports.map((p) => p.containerPort);
ports.push(...containerPorts);
```

src/components/kubectl/port-forward.ts
```diff
@@ -90,23 +96,50 @@ export async function portForwardKubernetes (kubectl: Kubectl, explorerNode?: an
...
  */
-async function promptForPort (): Promise<PortMapping[]> {
+async function promptForPort (kubectl?: Kubectl, podName?: string, namespace?: string): Promise<PortMapping[]> {
     let portString: string | undefined;
+    let defaultValue = undefined;
+    if (podName && kubectl) {
+        const ns = namespace || 'default';
+        try {
+            const result = await kubectl.invokeAsync(`get pods ${podName} --namespace ${ns} -o json`);
+            if (result && result.code === 0) {
```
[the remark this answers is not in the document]
brendandburns: done (refactored a little) console.log(...) for the errors instead.

src/components/kubectl/port-forward.ts
```diff
@@ -90,23 +96,50 @@ export async function portForwardKubernetes (kubectl: Kubectl, explorerNode?: an
...
+                const pod = JSON.parse(result.stdout) as kubernetes.V1Pod;
+                const containers = pod.spec.containers;
+                const ports = [] as number[];
```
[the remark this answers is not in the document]
brendandburns: done (couldn't accept suggestion due to refactor)

src/components/kubectl/port-forward.ts
brendandburns: removed.

src/components/kubectl/port-forward.ts
brendandburns: removed.

brendandburns: done. I prefer the second.

It's amazing how much like Lisp Javascript has become :)

brendandburns:

Thanks!

PR #498, opened 2019-03-13
Update to use native vscode apis.

brendandburns:
@itowlson

### DIFF
--- a/src/components/kubectl/dashboard.ts
+++ b/src/components/kubectl/dashboard.ts
@@ -1,7 +1,7 @@
 'use strict';
 
 import * as vscode from 'vscode';
-import opn = require('opn');
+import * as browser from '../platform/browser';
 
 import { createReadStream } from 'fs';
 import { resolve } from 'path';
@@ -85,7 +85,7 @@ async function openDashboardForAKSCluster (kubectl: Kubectl): Promise<void> {
     const boundPort = await portForwardToPod(kubectl, dashboardPod, portMapping, 'kube-system');
 
     setTimeout(() => {
-        opn(`http://localhost:${boundPort[0]}`);
+        browser.open(`http://localhost:${boundPort[0]}`);
     }, 2500);
     return;
 }
@@ -109,7 +109,7 @@ export async function dashboardKubernetes (kubectl: Kubectl): Promise<void> {
 
     // If we've already got a terminal instance, just open the dashboard.
     if (terminal) {
-        opn(KUBE_DASHBOARD_URL);
+        browser.open(KUBE_DASHBOARD_URL);
         return;
     }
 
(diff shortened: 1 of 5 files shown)


src/components/kubectl/dashboard.ts
```diff
@@ -85,7 +85,8 @@ async function openDashboardForAKSCluster (kubectl: Kubectl): Promise<void> {
     const boundPort = await portForwardToPod(kubectl, dashboardPod, portMapping, 'kube-system');
 
     setTimeout(() => {
-        opn(`http://localhost:${boundPort[0]}`);
+        const openExternal: any = (<any>vscode.env).open ? (<any>vscode.env).open : opn;
```
itowlson: Suggest we put this in a utility module somewhere, so that the API munging logic doesn't pollute the call sites.

src/debug/debugSession.ts
```diff
@@ -454,7 +454,8 @@ export class DebugSession implements IDebugSession {
     private async openInBrowser(errorMessage: string, link: string): Promise<void> {
         const answer = await vscode.window.showErrorMessage(errorMessage, "Open in Browser");
         if (answer === "Open in Browser") {
-            opn(link);
+            const openExternal: any = (<any>vscode.env).open ? (<any>vscode.env).open : opn;
```
itowlson: Oh, it's used more than once - definitely move it to a module then!  *grin*

src/extension.ts
```diff
@@ -1812,8 +1812,13 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
     }
 }
 
-function copyKubernetes(explorerNode: explorer.KubernetesObject) {
-    clipboard.writeSync(explorerNode.id);
+async function copyKubernetes(explorerNode: explorer.KubernetesObject) {
+    const clipboard = (<any>vscode.env).clipboard;
+    if (clipboard) {
```
itowlson: Again I would suggest a separate module exporting a 'copy text to clipboard' function that hides these details away.

src/extension.ts
```diff
@@ -1812,8 +1812,13 @@ async function deleteContextKubernetes(explorerNode: explorer.KubernetesObject)
...
+        await clipboard.writeText(explorerNode.id);
+    } else {
+        await clipboard.write(explorerNode.id);
```
itowlson: Local `clipboard` will shadow the imported `clipboard` won't it?  In VS Code builds without `env.clipboard` I think this might fail with an undefined error; at best, the different meanings of the name are confusing.  Change the name of the local or the import.

src/components/kubectl/dashboard.ts
Chuxel: I noticed our example was out of date - the API name got updated to "openExternal", so this should be `(<any>vscode.env).openExternal`.  Sorry about that!

brendandburns: @itowlson extracted to a utility module.

@Chuxel updated.

src/debug/debugSession.ts
brendandburns: done.

src/extension.ts

src/extension.ts
brendandburns: fixed.

brendandburns:
@itowlson fixed, please take another look.

itowlson:
CI is showing a failure which is probably related to the files I can't find, but the logs are completely unhelpful...

brendandburns:
@itowlson was missing some files. fixed.

PR #530, opened 2019-04-26
Match appended view item contexts for commands

itowlson:
The UI customisation API opens up the possibility for evildoers (not _you_, dear reader; lollards and schismatics, but never you) to modify the `treeItem.contextValue` of explorer nodes.  While we can't (short of wrapping the entire TreeItem API) stop them doing this, we can encourage them to append to the context value instead of replacing it entirely.

Our current menu contribution entries don't quite allow for this though, as they either do an exact match or do a regex match with start and finish metacharacters.  This PR modifies it so commands always pattern-match, and allow the pattern to appear anywhere in the string.  So commands should continue to appear as long as we have _civilised_ evildoers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -353,168 +353,168 @@
             "view/item/context": [
                 {
                     "command": "extension.vsKubernetesUseContext",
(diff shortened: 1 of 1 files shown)


brendandburns:
LGTM.

PR #549, opened 2019-05-23
Reduce keywords to 5

### LINKED ISSUE #534: Reduce keyword list to 5 keywords
@brendandburns turns out that when we package the extension it limits the `keywords` section to 5 keywords, which we now go over.  What are your preferences in terms of pruning it back to 5?

cc @squillace

brendandburns:
Fixes #534 

@itowlson

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,14 +17,10 @@
     ],
     "keywords": [
         "kubernetes",
-        "k8s",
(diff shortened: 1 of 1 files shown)


brendandburns:
@itowlson @bnookala friendly ping....

PR #553, opened 2019-05-27
Languages

remcohaszing:
The `ignore` language id is also used for many other ignore files.

Requirements is written in YAML format.


bnookala:
This is great - thank you @remcohaszing!

PR #557, opened 2019-06-01
Add web view support for follow logs.

brendandburns:
cc @itowlson

### DIFF
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -8,6 +8,7 @@ import * as yaml from 'js-yaml';
 import * as kubectlUtils from '../../kubectlUtils';
 import { LogsPanel } from '../../components/logs/logsWebview';
 import { ContainerContainer } from '../../utils/containercontainer';
+import { ChildProcess } from 'child_process';
 
 export enum LogsDisplayMode {
     Show,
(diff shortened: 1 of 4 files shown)


brendandburns:
@itowlson 

How do you see this working? Would we modify the `invokeAsync` call to use `spawn-rx`

I'm a little vague on the concrete proposal, I agree it looks a little cleaner (though this PR's not _that_ bad)

itowlson:
@brendandburns What I was thinking of was something like introducing the following method for `shell`/`kubectl` (roughly analogous to your `execStreaming`):

```javascript
export function invokeTracking(command: string, args: string[]): rx.Observable<string> {
    const linesSubject = new rx.Subject<string>();
    let pending = '';
    const stdout = spawnrx.spawn(command, args);
    stdout.subscribe((chunk) => {
        const todo = pending + chunk;
        const lines = todo.split('\n').map((l) => l.trim());
        const lastIsWholeLine = todo.endsWith('\n');
        pending = lastIsWholeLine ? '' : lines.pop()!;
        for (const line of lines) {
            linesSubject.next(line);
        }
    });
    return linesSubject;
}
```

You can then use Rx to subscribe to the stream of lines being emitted, and append them to the web view content using the messaging protocol or by updating the HTML property.  This would look pretty similar to what you have except that you wouldn't need to manage line breaking etc. in the panel - it would be handled by `invokeTracking/execStreaming/whatever`.

What I like about this is that it encapsulates the management of line breaks, child processes etc. and gives us just a general-purpose interface for getting line-oriented output from a running process - I like the abstraction this gives me over dealing directly with the ChildProcess API in the application code.  It also means we call a synchronous method and get back an object that raises events, instead of calling an async method and having to skip past the result.  The Rx library also means we can do batching and buffering so that if, say, a whole bunch of log lines (or stdout chunks) appear in 0.1 second, we can package them up and do one update to the HTML instead of several.

Though... reviewing what I was finding in the Stern POC... what I _don't_ like about it is that it doesn't give us a convenient way to kill the process (this seems to be a limitation of `spawn-rx`).  And that's probably a deal-breaker for this use case in the real world.  So maybe we should build on your approach of going all the way down to the ChildProcess level.  But I still wouldn't overload `invokeAsync` - I'd still approach it via a new method that returns a Disposable and hands lines to a callback (or perhaps even better, returns an object that raises events for stdout and also has a method for killing the child process).

And in the end, it _doesn't_ make a huge difference - and I know you are super busy.  So as I mentioned, if you don't feel this justifies your time, then I'm happy to take the PR as it stands.  I'd quite understand that!

itowlson:
Taking this as it is because the technical suggestion isn't worth letting it languish.  Thanks @brendandburns!

PR #559, opened 2019-06-05
Add Buildah as an alternative container image build tool

azatsarynnyy:
In this PR I propose to add [Buildah](https://github.com/containers/buildah) support as an alternative container image build tool that is used for basic image operations by `Kubernetes: Run` and `Kubernetes: Debug` commands.

By default, Docker is used. And if the user prefers using Buildah it can be enabled with `preferBuildah` configuration.


azatsarynnyy:
cc @gorkem

src/imageBuild/buildToolsRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+export function getPushCommand(image: string): string {
+    const buildTool = supportedTools[currentBuildTool];
+    return buildTool.getPushCommand(image);
+}
+
+/** Currently used image build tool. */
+export let currentBuildTool: string;
+
+/** Starts listening changing the preferred build tool. */
+export function start(): vscode.Disposable {
```
itowlson: What would be the cost of just looking up the config entry each time, instead of having a mutable variable and an event listener?  We do this throughout the extension and it feels simpler and doesn't seem to cause performance problems (he said cautiously, acutely aware that we _do_ have performance problems).

src/imageBuild/buildToolsRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+    detectCurrentBuildTool();
```
itowlson: This makes me anxious... in practice it should never be an issue, but in theory `detectCurrentBuildTool` could take a long time to run on extension startup, and so `currentBuildTool` could be accessed before it had been set.

src/imageBuild/buildToolsRegistry.ts
itowlson: Oh, it's because it's not just looking in config, it's also checking if `buildah` is present.  And we definitely _don't_ want to do that every time!

package.json
```diff
@@ -231,6 +231,10 @@
                         "resource-commands-on-files": {
                             "type": "boolean",
                             "description": "If true, show Kubernetes resource commands on file context menu for all YAML files"
+                        },
+                        "preferBuildah": {
```
itowlson: I am not sure this is the right design here - what if someone comes along with a PR to support another build tool, say `img`?  How would we manage the interaction between `preferBuildah` and a hypothetical `preferImg`?

Suggest instead a `preferredImageBuildTool` with an enum that restricts it to be `Docker` or `Buildah` (cf. the `kubectlVersioning` config setting).

src/imageBuild/buildToolsRegistry.ts
```diff
@@ -0,0 +1,94 @@
+import * as vscode from 'vscode';
+
+import { getPreferBuildah } from '../components/config/config';
+import { shell } from '../shell';
+
+interface ContainerImageBuildTool {
```
itowlson: I like the way you've done this - should make it easy if people want to add other build tools in future.

src/imageBuild/buildToolsRegistry.ts
itowlson: For most tools we defer detection until the user tries to use the tool.  I think we check for `kubectl` on startup but that is just to warn - we don't rely on it for a setting.  We also have existing mechanisms for checking if a binary is present (see e.g. `binutil.ts`) which might be faster than calling the executable.

azatsarynnyy:
Hello @itowlson, thank you for your review!
I've addressed all your notes and updated the PR:
- reworked the config design to the list of the possible tools;
- replaced a tool detection on the extension start with the existing mechanisms provided by `binutil.ts`.

itowlson:
Could you check the status of your CLA and sign if you haven't already done so please?  Unfortunately we can't take this PR without a CLA.

https://github.com/Azure/vscode-kubernetes-tools/pull/559#issuecomment-499184283

itowlson:
Thank you @azatsarynnyy for the CLA sign-off - we have lift-off!

PR #560, opened 2019-06-07
Cluster explorer refactor

itowlson:
The cluster explorer code has become extremely complex and fragile.  This is hampering rapid development and specifically making it more complicated and confusing to support extending it through the API.  As the cluster explorer is widely used, and is seen as a key aspect of the API, it's worth taking a bit of time to make it simpler.

~At this stage, this is just breaking stuff out and trying to improve naming and safety (e.g. not overloading a top-level `id` property).  Before this goes to release, I want to pare back some of the special-case handling and clarify the way built-in features are surfaced through the API, probably by aligning the internal structures more closely to the API.  Hopefully this will make it easier to implement new types of tree node both internally and via the API, and I won't disappear into a hall of mirrors every time I crack open the code.~  Externalising the special case handling is now done.  I could still do with betterifying the way builtins are surfaced through the API.


src/components/clusterexplorer/explorer.ts
```diff
@@ -0,0 +1,123 @@
...
+    'configitem' |
+    'helm.release' |
+    'extension';
+
+export function create(kubectl: Kubectl, host: Host): KubernetesExplorer {
+    return new KubernetesExplorer(kubectl, host);
+}
+
+export function isKubernetesExplorerResourceNode(obj: any): obj is ClusterExplorerResourceNode {
+    return obj && obj.nodeCategory === KUBERNETES_EXPLORER_NODE_CATEGORY && obj.nodeType === 'resource';
```
brendandburns: String constant (`resource`) duplicated at least twice, extract as a `const` for sanity.

(here and everywhere else this happens)

src/components/clusterexplorer/explorer.ts
```diff
@@ -0,0 +1,123 @@
...
+            if (affectsUs(change)) {
+                this.refresh();
+            }
+        });
+    }
+
+    getTreeItem(element: ClusterExplorerNode): vscode.TreeItem | Thenable<vscode.TreeItem> {
+        const treeItem = element.getTreeItem();
+
+        const treeItem2 = providerResult.transform(treeItem, (ti) => {
```
brendandburns: more descriptive names?

src/components/clusterexplorer/explorer.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private getChildrenBase(parent?: ClusterExplorerNode): vscode.ProviderResult<ClusterExplorerNode[]> {
+        if (parent) {
+            return parent.getChildren(this.kubectl, this.host);
+        }
+        return this.getClusters();
+    }
+
+    refresh(): void {
```
brendandburns: you don't need an explicit `void` in TypeScript do you?

src/components/clusterexplorer/explorer.ts
```diff
@@ -0,0 +1,123 @@
...
+        // This is a pretty niche case, so I'm not too worried if this isn't perfect.
+        //
+        // TODO: VS Code now doesn't require a reload on extension install.  Do we need
+        // to listen for the extension install event and refresh, in case an extension
+        // attempts to contribute while the tree view is already open?
+        //
+        // TODO: we need to check collapsibleStates in case someone adds child nodes to a
+        // parent which currently has CollapsibleState.None.
+        if (extender.contributesChildren(undefined)) {
+            sleep(50).then(() => refreshExplorer());
```
brendandburns: This is clearly a hack (and duplicated below) refactor so that we have one impl of the hack and an explanation of why we need it with that impl.

src/components/clusterexplorer/extension.nodesources.ts
```diff
@@ -0,0 +1,63 @@
+import * as kuberesources from '../../kuberesources';
+import { ExplorerExtender } from './explorer.extension';
+import { ClusterExplorerNode } from './node';
+import { ContextNode } from './node.context';
+import { FolderNode } from './node.folder';
+import { ContributedGroupingFolderNode } from './node.folder.grouping.custom';
+import { ResourceFolderNode } from './node.folder.resource';
+
+export abstract class NodeSourceImpl {
```
brendandburns: This needs better documentation.

src/components/clusterexplorer/node.configurationvalue.ts
```diff
@@ -0,0 +1,26 @@
+import * as vscode from 'vscode';
+
+import { Kubectl } from '../../kubectl';
+import { Host } from '../../host';
+import { ClusterExplorerNode, ClusterExplorerNodeImpl, ClusterExplorerConfigurationValueNode } from './node';
+import { ResourceKind } from '../../kuberesources';
+
+export class ConfigurationValueNode extends ClusterExplorerNodeImpl implements ClusterExplorerConfigurationValueNode {
+    constructor(readonly configData: any, readonly key: string, readonly parentKind: ResourceKind, readonly parentName: string) {
+        super("configitem");
```
brendandburns: nit: single quotes.

not-nit `configitem` string is repeated several times, extract a constant.

brendandburns:
I added a few comments, but this is hard to review without more documentation about how it all fits together.

I would also suggest that as part of refactors like this we start adding in unit test coverage. I realize adding tests to the entire codebase may be impractical, but given the motivations for the refactor it sounds like it might be sensible to add it here....

src/components/clusterexplorer/explorer.ts
itowlson: If I don't put the `void` then TypeScript infers the return type.  Specifying `void` catches the case where a maintainer inadvertently changes it to return something - I want them to be forced to document to the return type if that happens.

src/components/clusterexplorer/explorer.ts
itowlson: This used to be impossible because constants had type `string` and the discriminator has type `'resource'`.  But I think TypeScript 3.4 now allows string literal constants to have string literal types.  Good call - I will look into this.

itowlson:
@brendandburns Thanks for the feedback.  I've added some documentation which hopefully makes things clearer, and improved the rest of the code per your comments.  If you have the time and inclination to take another look then that would be great, but I know it's a bit of a sprawling beast.

What I _haven't_ done, although I agree with the sentiment, is to implement unit tests for this stuff.  We used to have unit tests, but they weren't very effective, because most of what we do is so integration-y - working with `kubectl`, talking to git, proffering UI commands, etc.  Hopefully with refactors like this we can start to separate out a core of 'logical' (pure) components from the 'integration-y' periphery and unit test those.  I'd like to put that out of scope for this PR, though, so that we can merge it before it becomes a rebasing nightmare.  But yes, this is definitely a good thing to aim for in future refactors (and as a phase 2 of this one).

PR #566, opened 2019-06-14
Upgrade telemetry package

itowlson:
Because it pulls in a very old version of Application Insights, and AI has been implicated in an issue with HTTP requests from other extensions; so if we are going to need support with that then we want to be on the latest version.


bnookala:
I'm very excited for [dependabot](https://dependabot.com/) to be added to all github repositories - should we set it up ourselves?

PR #567, opened 2019-06-14
Fix node pods not displayed with kubectl 1.12+

### LINKED ISSUE #565: Can no longer see all pods in a node
Previously, I could click on a node and get a listing of all pods on that node. However, it seems that after upgrading to somewhere around gke's k8s version 1.12+ I no longer have this capability. I am still able to list pods using my local k8s which is running v 1.10.

Obviously, this is due to some change with K8s. Is this a known issue with newer versions of Kubernetes?

itowlson:
Fixes #565.

The underlying issue is that someone forgot that `kubectl` table output is meant to be parseable by tools such as `awk`, and that multi-word column headers should therefore be hyphenated instead of spaced.  This person added a pod column called `NOMINATED NODE`.  Because we parse based on space-separation, we interpreted this as two columns, `NOMINATED` and `NODE`.  And the spurious `NODE` column duplicated and overwrite the real `NODE` column which we needed to relate pods to nodes.

Fortunately it seems that we can - for now at least - rely on at least three spaces between columns, per the TabWriter implementation and the settings in the k8s `tabwriter.go` file.  So there is a simple fix to set the column separator regex to at least two spaces.  It would be good to get this fixed upstream for the benefit of other tools though.  But we should also implement robustness improvements in our own parser, e.g. to deal with spaces in data fields.

NOTE: Unfortunately we can't use structured-text solutions such as `-o json`, or solutions where we control the output such as `-o custom-columns`, to work around this, because we need the `STATUS` field.  This does not map to a single server field, but is instead synthesised by the `kubectl` table formatter.  To my mind it is a bug in `kubectl` that we can't access this value without surrendering to an unpredictable output format.  But there it is.

### DIFF
--- a/src/kubectl.ts
+++ b/src/kubectl.ts
@@ -10,7 +10,7 @@ import * as compatibility from './components/kubectl/compatibility';
 import { getToolPath, affectsUs, getUseWsl, KubectlVersioning } from './components/config/config';
 import { ensureSuitableKubectl } from './components/kubectl/autoversion';
 
-const KUBECTL_OUTPUT_COLUMN_SEPARATOR = /\s+/g;
+const KUBECTL_OUTPUT_COLUMN_SEPARATOR = /\s\s+/g;
 
 export interface Kubectl {
     checkPresent(errorMessageMode: CheckPresentMessageMode): Promise<boolean>;


bnookala:
Lol… this is a fun one. You mention that we need the `STATUS` field from the table output, because it's synthesized from `kubectl`. I noticed that outputting with `-o json` outputs a `phase` field ie.              `"phase": "Running"`. This field also seems to map to the `phase` field in the PodStatus object on the Pod: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase.

Wondering if we could use this in the future.

bnookala:
Also how could they ;_;

PR #575, opened 2019-07-03
Format minikube status dependent on its version

ederst:
This fixes an error when querying the status of minikube by choosing a `--format` string appropriate for its version.

Fixes issue: https://github.com/Azure/vscode-kubernetes-tools/issues/467

### DIFF
--- a/src/components/clusterprovider/minikube/minikube.ts
+++ b/src/components/clusterprovider/minikube/minikube.ts
@@ -121,7 +121,14 @@ async function minikubeUpgradeAvailable(context: Context): Promise<void> {
         return;
     }
 
-    const versionInfo = await getVersionInfo(context);
+    let versionInfo: MinikubeVersionInfo;
+
+    try {
+        versionInfo = await getVersionInfo(context);
+    } catch (err) {
+        vscode.window.showErrorMessage(`Failed to determine minikube version: ${err}`);
+        return;
+    }
 
     if (versionInfo.currentVersion !== versionInfo.availableVersion) {
         const value = await vscode.window.showInformationMessage(`Minikube upgrade available to ${versionInfo.availableVersion}, currently on ${versionInfo.currentVersion}`, 'Install');
--- a/src/components/clusterprovider/minikube/minikube.ts
+++ b/src/components/clusterprovider/minikube/minikube.ts
@@ -9,6 +9,7 @@ import { Errorable, failed } from '../../../errorable';
 import { fromShellExitCodeOnly, Diagnostic } from '../../../wizard';
 import { getToolPath, getCheckForMinikubeUpgrade } from '../../config/config';
 import { installMinikube } from '../../installer/installer';
+import * as semver from 'semver';
 
 export class MinikubeInfo {
     readonly running: boolean;
@@ -100,9 +101,16 @@ async function getVersionInfo(context: Context): Promise<MinikubeVersionInfo> {
         throw new Error(`Unexpected output for minikube version check: ${lines}`);
     }
 
+    const currentVersion = semver.clean(extractVersion(lines[0]));
+    const availableVersion = semver.clean(extractVersion(lines[0]));
+
+    if (currentVersion === null || availableVersion === null) {
+        throw new Error(`Unable to get version from minikube version check: ${lines}`);
+    }
+
     return {
-        currentVersion: extractVersion(lines[0]),
-        availableVersion: extractVersion(lines[1])
+        currentVersion: currentVersion,
+        availableVersion: availableVersion
     } as MinikubeVersionInfo;
 }
 
@@ -210,14 +218,12 @@ async function stopMinikube(context: Context): Promise<void> {
 }
 
 function getStatusFormatString(currentVersion: string): string {
-    if (currentVersion.startsWith("v0.")) {
-        if (parseInt(currentVersion.split(".")[1]) < 31) {
-            // minikube < 0.31.0 uses this format
-            return `"{{.MinikubeStatus}}","{{.ClusterStatus}}","{{.KubeconfigStatus}}"`;
-        } else {
-            // minikube < 1.0.0 uses this format
-            return `"{{.Host}}",{ "kubelet": "{{.Kubelet}}", "apiserver": "{{.ApiServer}}" },"{{.Kubeconfig}}"`;
-        }
+    if (semver.lt(currentVersion, "0.31.0")) {
+        // minikube < 0.31.0 uses this format
+        return `"{{.MinikubeStatus}}","{{.ClusterStatus}}","{{.KubeconfigStatus}}"`;
+    } else if (semver.lt(currentVersion, "1.0.0")) {
+        // minikube < 1.0.0 uses this format
+        return `"{{.Host}}",{ "kubelet": "{{.Kubelet}}", "apiserver": "{{.ApiServer}}" },"{{.Kubeconfig}}"`;
     }
 
     // minikube >= 1.0.0 uses this format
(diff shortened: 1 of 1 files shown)


src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -198,13 +198,42 @@ async function stopMinikube(context: Context): Promise<void> {
     });
 }
 
+function getStatusFormatString(currentVersion: string): string {
+    if (currentVersion.startsWith("v0.")) {
+        if (parseInt(currentVersion.split(".")[1]) < 31) {
```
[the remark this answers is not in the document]
ederst: i knew there was a library for that; will try to implement.

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -198,13 +198,42 @@ async function stopMinikube(context: Context): Promise<void> {
...
+            // minikube < 0.31.0 uses this format
+            return `"{{.MinikubeStatus}}","{{.ClusterStatus}}","{{.KubeconfigStatus}}"`;
+        } else {
+            // minikube < 1.0.0 uses this format
+            return `"{{.Host}}",{ "kubelet": "{{.Kubelet}}", "apiserver": "{{.ApiServer}}" },"{{.Kubeconfig}}"`;
+        }
+    }
+
+    // minikube >= 1.0.0 uses this format
+    return `"{{.Host}}",{ "kubelet": "{{.Kubelet}}", "apiserver": "{{.APIServer}}" },"{{.Kubeconfig}}"`;
```
itowlson: I think this will result in `JSON.parse` returning an object for the second element of the array?  But the second element gets parsed into `MinikubeInfo.cluster` which is currently declared as `string`.  Can we tweak the `MinikubeInfo` interface to reflect this?  (Though it doesn't look like `cluster` is used anywhere so maybe we could just get rid of it.)

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -198,13 +198,42 @@ async function stopMinikube(context: Context): Promise<void> {
...
+    if (!sr || sr.code !== 0) {
+        throw new Error(`Error checking for minikube version: ${sr ? sr.stderr : 'cannot run minikube'}`);
+    }
+    const lines = sr.stdout.split('\n')
+                           .map((l) => l.trim())
+                           .filter((l) => l.length > 0);
+    if (lines.length !== 1) {
+        throw new Error(`Unexpected output for minikube version check: ${lines}`);
+    }
+    const currentVersion = extractVersion(lines[0]);
```
itowlson: I think we have similar code to this in `minikubeUpgradeAvailable` - is it worth pulling out the common code?

src/components/clusterprovider/minikube/minikube.ts
itowlson: Later on we determine `MinikubeInfo.isRunning` by comparing the first element of the JSON array to `Stopped`.  What does `{{.Host}}` return?  Is the comparison to `Stopped` meaningful in this case?

ederst: the only place where i could find the usage of this code is here: https://github.com/Azure/vscode-kubernetes-tools/blob/7efd603d46e80da9a5c4181143150dc133aedbed/src/extension.ts#L201

it references the first element (`running`), i don't know why other fields are returned though, because they get never used. i just chose to adapt the return value for the newer output accordingly.

i probably could get rid of all the additional elements, just return `.status` (aka a boolean), and probably let status be something like `running: 'Stopped' !== obj[0] && obj[1]`; but i don't know if this is really necessary, and somehow blows up the scope of this pull request.

src/components/clusterprovider/minikube/minikube.ts
ederst: yes, i copied the code (for me it was the fastest way; sometimes i'm lazy like this), but i will look into pulling out the common code

src/components/clusterprovider/minikube/minikube.ts
ederst: see https://github.com/Azure/vscode-kubernetes-tools/pull/575#discussion_r300956555; btw i did not change the logic here, so i'm not sure if it makes sense.

anyhow, as far as i understand minikube:
host: Stopped/Running -> status of the vm
kubelet: Stopped/Running -> status of the kubelet in the minikube vm
apiserver: Stopped/Running -> status of the apiserver of k8s in the minikube vm

and for the old return values i'd guess that "clusterstatus" is "kubelet" and "apiserver" combined. i'd have to verify this in the minikube docu/code but sounds right to me.

ederst:
I tried something here:
All commits, except the last one address the change requests for using "semver" and "pulling out the common code for determining the version".

However, the last one (currently "Refactor getting status of Minikube", https://github.com/Azure/vscode-kubernetes-tools/pull/575/commits/628955f9a370e7f06a8b5eecdc082b7e46a69767; see it as optional for this pull request) revamps the whole status thing to be not dependent on any version anymore:
* Calls `minikube status` without a format string
* "Parses" first line for `.running` to support it for `minikubeStart/Stop()`
* Displays status message in "Kubernetes channel" in tab "Output"

So this would address the whole ordeal with "isRunning", is it necessary to parse states apart from ".Host", how to parse/reflect the new vs the old status, and renders the whole parse output dependent on version unnecessary.

@itowlson, @bnookala: Maybe this is the way to go?

This would of course more or less imply that I have to change the title of the pull request, since the changes there do not reflect its title.

src/components/clusterprovider/minikube/minikube.ts
itowlson: Sounds good - I wasn't sure what `Host` contained but you have confirmed it's a status field so that suffices for me!  As you say if we want to get fancier with the logic that's outside the scope of this PR.

itowlson:
@brendandburns You did the original Minikube implementation and the Minikube Status stuff.  @ederst's comment at https://github.com/Azure/vscode-kubernetes-tools/pull/575#issuecomment-509599891 sounds good to me - are you okay if he changes it to work this way?  Thanks!

itowlson:
@ederst I'm comfortable with putting the last commit on hold and making it a separate PR, so as to keep this one simple and least-possible-change (and to give Brendan time to weigh in!).  It's your call though - we very much appreciate the fix / improvement either way!

ederst:
@itowlson sorry it took me so long, i removed the last commit, so that we can handle the whole version dependent formatting in this pull request (like it is supposed to be). i will create a separate pull request for removing the version dependency of the status call. there we can discuss with the original creator if we want to remodel the feature this or another way.

hope i got everything from your review comments for now.

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -83,31 +89,44 @@ class MinikubeImpl implements Minikube {
...
+    const performUpgradeCheck = await checkPresent(context, CheckPresentMode.Silent) && getCheckForMinikubeUpgrade();
+    if (!performUpgradeCheck) {
         return;
     }
-    const currentVersion = extractVersion(lines[0]);
-    const availableVersion = extractVersion(lines[1]);
-    if (currentVersion !== availableVersion) {
-        const value = await vscode.window.showInformationMessage(`Minikube upgrade available to ${availableVersion}, currently on ${currentVersion}`, 'Install');
+
+    const versionInfo = await getVersionInfo(context);
```
itowlson: `getVersionInfo` has some explicit `throw`s in it - should this be in a try block?

ederst: i think https://github.com/Azure/vscode-kubernetes-tools/pull/575/commits/504284da3bbc78122cfb00ce429fd09f9679a701 will do

Issue #542, opened 2019-05-12
Fix slow initial load


itowlson:
We have a number of autogenerated reports of high CPU load during the first run of the extension; anecdotally it is feeling slow to load/unpack again.  This is a tracking issue for that problem.  See also #500.

radu-matei:
FWIW, I can also confirm the initial load time of the extension when VS Code starts has significantly increased.

itowlson:
The good news: Webpack does make this _hugely_ faster.

The bad news: Webpack is currently blocked by the `jsonpath` dependency (via `@kubernetes/client-node`): https://github.com/webpack/webpack/issues/9367

PR #576, opened 2019-07-04
Get rid of 30% of the files in the VSIX

itowlson:
We had a dependency on restify, which we no longer use.  The `restify` package includes a benchmarks directory, which has its own `node_modules` directory containing OVER TWO THOUSAND files.

This PR, landing correctly at Thanet, does to `restify` what the Wave of Danes did to England.  🔥  ⚔️ 

image


itowlson:
This is a partial remediation for #542 but I still think we are going to need to Webpack the thing and that is proving a pain...

bnookala:
mother of god

PR #581, opened 2019-07-17
Make "minikube status" version independent

ederst:
Like suggested in https://github.com/Azure/vscode-kubernetes-tools/pull/575 this will remove the version dependency of the `minikube status --format` output and do the following:
* Use the first line of the output (.Host/.MinikubeStatus) for determining the status of Minikube (running/stopped)
* Instead of the notification bar (`showInfoMessage`) the whole output of `minikube status` is shown in the `Output -> Kubernetes` channel; So now users can see the status of the apiserver and kubelet, and the configuration of the kubectl.

@brendandburns: Like @itowlson already mentioned in https://github.com/Azure/vscode-kubernetes-tools/pull/575#issuecomment-509927412 I am changing the the original behavior of your work, and want to include you in the discussion whether this is ok or not.

### DIFF
--- a/src/components/clusterprovider/minikube/minikube.ts
+++ b/src/components/clusterprovider/minikube/minikube.ts
@@ -9,12 +9,10 @@ import { Errorable, failed } from '../../../errorable';
 import { fromShellExitCodeOnly, Diagnostic } from '../../../wizard';
 import { getToolPath, getCheckForMinikubeUpgrade } from '../../config/config';
 import { installMinikube } from '../../installer/installer';
-import * as semver from 'semver';
 
 export class MinikubeInfo {
     readonly running: boolean;
-    readonly cluster: string;
-    readonly kubectl: string;
+    readonly message: string;
 }
 
 export class MinikubeVersionInfo {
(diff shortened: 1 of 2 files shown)


itowlson:
If you can rebase this on top of latest master (we usually squash-merge PRs, so I'm afraid it doesn't just slot in smoothly on top of #575) then I think it's ready to go straight in...

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -224,37 +222,22 @@ async function stopMinikube(context: Context): Promise<void> {
...
     }
 
-    const statusFormatString = getStatusFormatString((await getVersionInfo(context)).currentVersion);
+    const result = await context.shell.exec(`"${context.binPath}" status`);
 
-    const result = await context.shell.exec(
-        `"${context.binPath}" status --format '[${statusFormatString}]'`);
     if (result && result.stderr.length === 0) {
-        const obj = JSON.parse(result.stdout);
+        const hostStatus = result.stdout.split('\n')[0].split(': ')[1].toLowerCase();
```
ederst: Personally, I do not like the way i did the parsing here. Maybe there is a more safe and elegant way...

ederst:
rebased

PR #586, opened 2019-07-23
Bundle the extension

itowlson:
Bundling velocitates extensioning.  By quite a bit.

**NOTE:** To get this to webpack, I had to rev a couple of packages that were bringing in unpackable packages.  This caused some minor knock-on effects in the application code, e.g. in port-forward.

**THINGS I STILL NEED TO DOUBLE CHECK:**

* ~~Telemetry (`aiKey` being loaded correctly)~~ Confirmed
* ~~Completion driven off the `snippets` directory~~ Confirmed


itowlson:
+ @lstocchi for moar eyes on this beast

src/yaml-support/yaml-constant.ts
```diff
@@ -8,10 +8,10 @@ export const VSCODE_YAML_EXTENSION_ID = 'redhat.vscode-yaml';
 
 export const KUBERNETES_SCHEMA_VERSION = '1.12.2';
 
-export const KUBERNETES_SCHEMA_FILE = path.join(__dirname, `../../../schema/swagger-v${KUBERNETES_SCHEMA_VERSION}.json`);
-export const FALLBACK_SCHEMA_FILE = path.join(__dirname, `../../../schema/swagger-v${KUBERNETES_SCHEMA_VERSION}.json`);
+export function KUBERNETES_SCHEMA_FILE(context: ExtensionContext) { return context.asAbsolutePath(`schema/swagger-v${KUBERNETES_SCHEMA_VERSION}.json`); }
+export function FALLBACK_SCHEMA_FILE(context: ExtensionContext) { return context.asAbsolutePath(`schema/swagger-v${KUBERNETES_SCHEMA_VERSION}.json`); }
 
-export const KUBERNETES_SCHEMA_ENUM_FILE = path.join(__dirname, `../../../schema/schema_enums-v${KUBERNETES_SCHEMA_VERSION}.json`);
+export function KUBERNETES_SCHEMA_ENUM_FILE(context: ExtensionContext) { return context.asAbsolutePath(`schema/schema_enums-v${KUBERNETES_SCHEMA_VERSION}.json`); }
```
lstocchi, line 14: In all three functions you could use the `assetPath` function and so avoid passing the context everytime, no?

itowlson: I think you're right.  This is most likely a historical accident where these changes date from the Before Giving Up On Avoiding Global State era of the refactor (where `assetPath` was my admission of defeat).  But it's just possible there might be some subtle ordering dependency where these functions could run before `setAssetContext` got called (due to code running during imports) - I don't remember such a thing but will double check, and I'll make the change you recommend unless I can find a good reason for the difference.  Thanks!

itowlson: Changed - thanks!

gorkem:
Is this fixing #542 ?

itowlson:
@gorkem Yes - bundling dramatically improves startup speed, and reduces CPU load and download size.

gorkem:
@bnookala this looks like it is waiting on your review. 

@itowlson can we move this along, it is an important improvement

bnookala (approved):
long overdue +1, but also we should probably get that merge conflict fixed

gorkem:
🎉 🎉 🎉

PR #595, opened 2019-07-31
Fix config value display truncation

### LINKED ISSUE #594: URI query parameters are not shown in secret preview
For some obscure reason query parameters are stripped from URIs when viewing an individual value of a secret through the Clusters panel.

## How to reproduce

1. Create a secret like this and apply it:

```yml
apiVersion: v1
kind: Secret
metadata:
  name: test
stringData:
  URI: http://example.com/test?blah=sth
```

2. Reload the Clusters panel.

3. Click on the `test` secret in the tree view. You'll see that the `URI` value has been correctly Base64-encoded:

```yml
data:
  # -> http://example.com/test?blah=sth
  URI: bW9uZ29kYjovL2V4YW1wbGUuY29tL3Rlc3Q/YmxhaD1zdGg=
```

4. Now click on `URI` (the individual value nested inside `test`) in the tree view:

```
mongodb://example.com/test
```

As you can see, the URI does not have any query parameters anymore.

itowlson:
To display values out of secrets and config maps, we stuff the base64-encoded value and its name into a URL, then the document provider extracts the value by taking the first chunk of the URL path.  Unfortunately, base64 encoding allows `/` as a character in the encoded string.  When we naively split the URL path to retrieve the value, if it contained a `/`, we get the truncated portion up to the `/`.

Frustratingly, VS Code's `Uri.parse` function is clever enough to spot URI-encoded characters and 'helpfully' decode them, so the obvious fix of encoding the base64 string does not work.  My proposed workaround is to instead encode as `hex`, which is inefficient, but is built in and is extremely safe.  A more efficient solution could be achieved by bringing in a base32 encoding package, or by adding an additional transformation of the base64 encoded string (e.g. changing `/` to `!` before putting it into the Uri, and vice versa after taking it out).

Fixes #594.

### DIFF
--- a/src/configMap.ts
+++ b/src/configMap.ts
@@ -17,16 +17,16 @@ export class ConfigMapTextProvider implements vscode.TextDocumentContentProvider
 
     provideTextDocumentContent(uri: vscode.Uri, _token: vscode.CancellationToken): vscode.ProviderResult<string> {
         const parts = uri.path.split('/');
-        const b: Buffer = new Buffer(parts[1], 'base64');
+        const encodedData = parts[1];
+        const b: Buffer = Buffer.from(encodedData, 'hex');
         return b.toString();
     }
 }
 
 export function loadConfigMapData(obj: ClusterExplorerConfigurationValueNode) {
-    let encodedData = obj.configData[obj.key];
-    if (obj.parentKind.abbreviation === allKinds.configMap.abbreviation) {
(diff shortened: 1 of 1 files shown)


itowlson:
@bnookala I don't think we stuff anything else base64-encoded into URLs.  But yeah, it would be good to check.  We certainly use URLs for Things That URLs Should Not Be Used For.

Issue #620, opened 2019-09-19
Extension adds one additional curly brace in helm-related files


rtim75:
**vs-code-kubernets version: 1.0.4**
**vs-code version: 1.38.1**

When the extension is enabled hitting '{' key once creates '{}' as it should do. But, if I hit  '{' two times I get '{{}}}' which brakes every helm template and forces to manually remove that curly brace.

xarek1986:
I found that with vscode v1.37.1 it works as it has to, but with vscode v1.38.1 it adds additional brace

itowlson:
Sorry for the delay in replying to this - I've been away.  @xarek1986 thanks for the version info - this may be a regression (or the fix of a bug we inadvertently relied on!) in VS Code so I may need to flag it upstream to them.  Thanks all for reporting it!

itowlson:
@rtim75 @xarek1986 Would you mind giving this a go with #623 and let me know if that fixes this issue as well as the one it cites, please?

itowlson:
I believe #623 fixes this.

xarek1986:
@itowlson Yes, it works, thank you, I guess this issue can be closed

PR #623, opened 2019-09-25
fix auto closing pairs issue

### LINKED ISSUE #571: [Helm] `{{` Autocomplete adds extra `-}}` when `-` is typed
If I type `{{` code automatically adds `}}` and everything is fine. but if I write `-`, then code automatically adds `-}}` and I'll end up with `{{--}}}}` which is not fine.

ahmadalli:
fixes #571

### DIFF
--- a/language-configuration.json
+++ b/language-configuration.json
@@ -16,7 +16,6 @@
     // symbols that are auto closed when typing
     "autoClosingPairs": [
         ["{", "}"],
-        ["{{", "}"],
         ["{{-", "-"],
         ["[", "]"],
         ["(", ")"],
--- a/language-configuration.json
+++ b/language-configuration.json
@@ -15,8 +15,9 @@
     ],
     // symbols that are auto closed when typing
     "autoClosingPairs": [
-        ["{{", "}}"],
         ["{", "}"],
+        ["{{", "}"],
+        ["{{-", "-"],
         ["[", "]"],
         ["(", ")"],
         ["\"", "\""],


language-configuration.json
```diff
@@ -15,8 +15,9 @@
     ],
     // symbols that are auto closed when typing
     "autoClosingPairs": [
-        ["{{", "}}"],
```
zackijack: @ahmadalli I think you just need to remove the double bracket pairs because the single bracket will handle that.

ahmadalli: yeah that's doable too

language-configuration.json
```diff
@@ -15,8 +15,8 @@
...
         ["{", "}"],
+        ["{{-", "-"],
```
zackijack: Again, this is just my assumption, why not just add `["-", "-"]` instead of `["{{-", "-"]`?

ahmadalli: we don't want to have `--` for every `-` user types. do we?

zackijack: Ah I see

itowlson:
@ahmadalli Testing this, when I type `{{-` I end up with `{{--}}`.  This seems to be one more `-` than is desired - don't you want it to be `{{-}}`?  If you're happy that it's doing the right thing then I'll merge but this doesn't look quite right to me...?

ahmadalli:
@itowlson both `{{- code }}` and `{{- code -}}` are valid with different use cases. the decision of the contributors was having `-}}` when user typed `{{-` but the functionality was faulty (what I described on https://github.com/Azure/vscode-kubernetes-tools/issues/571#issue-457501899) so it was removed to fix #571. this PR fixes both the original issue of #571 and the new issue reported on #571 and #620

buddyledungarees:
For anyone else that runs into this autocomplete issue, this MR didn't change the behavior to how I prefer it so my workaround is to overwrite the packaged language-configuration.json under `~/.vscode/ms-kubernetes-tools.vscode-kubernetes-tools-{version}/` with the following setting:

```
{
    "comments": {
        // comments applicable to yaml files.
        "lineComment": "#",
        // comments applicable to tpl files.
        "blockComment": [ "{{- /*\n", "\n*/ -}}" ]
    },
    // symbols used as brackets
    "brackets": [
        ["{{", "}}"],
        ["{{-", "-}}"],
        ["{", "}"],
        ["[", "]"],
        ["(", ")"]
    ],
    // symbols that are auto closed when typing
    "autoClosingPairs": [
        ["{", "}"],
        ["{{-", "}}"],
        ["[", "]"],
        ["(", ")"],
        ["\"", "\""],
        ["'", "'"],
        ["`", "`"]
    ],
    // symbols that that can be used to surround a selection
    "surroundingPairs": [
        ["{{", "}}"],
        ["{{-", "}}"],
        ["{{-", "-}}"],
        ["{", "}"],
        ["[", "]"],
        ["(", ")"],
        ["\"", "\""],
        ["'", "'"],
        ["`", "`"]
    ]
}
```

My personal experience with Helm templating is that it's easier to manage if you do not default to chomping the trailing newline and only chomp leading whitespace ie `{{- blah }}` since the usual intention is to remove the template line if it ends up not being rendered (if, with, range etc).

PR #625, opened 2019-09-29
Avoid polluting Array prototype

### LINKED ISSUE #613: Array.prototype pollution causing interactions in other extensions
I've had a [couple of reports](https://github.com/Mermade/openapi-lint-vscode/issues/9) of `Array.prototype` pollution which we have narrowed down to this extension, a quick search points to this code probably being the culprit:

https://github.com/Azure/vscode-kubernetes-tools/blob/070b7d4278b13f2d24fa17c5fb2ccc208becaf30/src/utils/array.ts#L16

Is it possible this function could be made non-enumerable, or some other workaround employed?

itowlson:
Fixes #613 by making `choose` method non-enumerable.

cc @MikeRalphson in case you want to sanity check this!

### DIFF
--- a/src/utils/array.ts
+++ b/src/utils/array.ts
@@ -12,8 +12,13 @@ declare global {
     }
 }
 
+function choose<T, U>(this: T[], fn: (t: T) => U | undefined): U[] {
+    return this.map(fn).filter((u) => u !== undefined).map((u) => u!);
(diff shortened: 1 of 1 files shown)


MikeRalphson:
LGTM, thanks.

PR #635, opened 2019-10-19
Prevent triple closing brackets

rogst:
Typing a `{` in a helm template would automatically add a closing bracket resulting in `{}`, if you try to type another `{` bracket, that would result in `{{}}}` instead of the expected `{{}}`.

### DIFF
--- a/language-configuration.json
+++ b/language-configuration.json
@@ -15,7 +15,6 @@
     ],
     // symbols that are auto closed when typing
     "autoClosingPairs": [
-        ["{{", "}}"],
         ["{", "}"],
         ["[", "]"],
         ["(", ")"],


itowlson:
@rogst I _think_ this is already covered by #623.  I'm sorry for the confusion - I dropped the ball on #623 due to time off, and so it was left open and not included in the latest release.  Thank you for re-raising the problem and prompting me to go back and check on that one.  However, #623 is now merged (although not released) and so I think we can close this PR.  Have a look and let me know if you think it's still needed!  And apologies again!

rogst:
Hi, thank you, I was not aware of #623, which covers the issue. Thanks for a great vscode plugin, closing this PR

PR #638, opened 2019-10-21
Fix error if kubectl path contains spaces

### LINKED ISSUE #637: Tool throws an error saying that the command isn't recognized when User Name has spaces in path
if you have a space in your user name like "c:\users\some body" the extension doesn't work correctly.

itowlson:
If the path to `kubectl` contained spaces, we were throwing an error because we did not quote the path before composing the command string and passing it to the shell.

Fixes #637.

### DIFF
--- a/src/kubectl.ts
+++ b/src/kubectl.ts
@@ -285,13 +285,21 @@ function kubectlDone(context: Context): ShellHandler {
     };
 }
 
-async function baseKubectlPath(context: Context): Promise<string> {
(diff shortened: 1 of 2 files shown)


fedorzh:
I am getting exactly the same error with the path which contains an opening and closing bracket..
@itowlson

Moazzem-Hossain-pixel:
Dear concern, 

I will be grateful if you can take initiative for fixing my issues. 

Kind Regards

Moazzem-Hossain-pixel:
We appreciate your efforts

PR #649, opened 2019-11-13
Install latest version of Helm

itowlson:
Previously the version of Helm to download in 'install dependencies' was hardcoded.  This PR uses @bacongobbler's trick of querying the GitHub latest release page to find out which release it redirects to, and then downloads that.  If the query fails it still falls back to a hardcoded version rather than failing: this fallback has been updated to 3.0.0 (yay Helm 3).

### DIFF
--- a/src/components/installer/installer.ts
+++ b/src/components/installer/installer.ts
@@ -1,5 +1,6 @@
 'use strict';
 
+import * as https from 'https';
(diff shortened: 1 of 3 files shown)


itowlson:
Thanks for the quick review folks - just breaking out the dependency fixes into a separate PR/commit before merging.

PR #661, opened 2019-11-28
Add syntax highlighting for Chart.lock

remcohaszing:
Chart.lock replaces requiremens.lock in Helm 3.


itowlson:
Looks good to me - just looping in the Helm folks and then we'll merge it.  Thanks!

PR #674, opened 2019-12-18
Add dotnet attach debugging support

tedchamb:
## 6. dotnet debugging
   1. Prerequisites
      * [C# for Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=ms-vscode.csharp) extension must be installed on local machine.
      * `ps` must be available on the docker container. `ps` is used to try and autoselect the process id to debug. It is also used by the `C#` extension to look up processes for the process picker if there is more than one dotnet process running. You can install it by running the following command on the container:

      `apt-get update && apt-get -y install procps`

      * [`vsdbg`](https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes) must be installed on the container. You can download it by running the following command on the container.
https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes
      `curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ~/vsdbg`

      \- or -

      `wget https://aka.ms/getvsdbgsh -O - 2>/dev/null | /bin/sh /dev/stdin -v latest -l ~/vsdbg`

   2. Notes:
      * `dotnet` debugging is only supported for the `attach` scenario. The `launch` scenario is not yet supported.
      * If `dotnet` does not show up in debugger picker list, than the extension was able to view the processes running on the container and there were no `dotnet` processes running at that time.
      * If `ps` is not installed on the machine, than `dotnet` will show up in the debugger pick list. If `dotnet` is chosen it will give you an error message with a link to this page.
      * If `dotnet` is chosen as the debugger and only one `dotnet` process is running on the container, that process will automatically be chosen.  If there is more than one dotnet process running at that time, you will be presented a process pick list, and you can choose the process you would like to debug.
      * The [C# for Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=ms-vscode.csharp) extension is powered by OmniSharp. Found out more information regarding remote debugging with OmniSharp [here](https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes).

### DIFF
--- a/src/debug/debugSession.ts
+++ b/src/debug/debugSession.ts
@@ -5,7 +5,8 @@ import * as path from "path";
 import * as portfinder from "portfinder";
 import { ChildProcess } from "child_process";
 
-import { IDebugProvider } from "./debugProvider";
+import { IDebugProvider, PortInfo } from "./debugProvider";
+import { ProcessInfo, getProcesses } from "./debugUtils";
 import * as providerRegistry from "./providerRegistry";
 
 import { kubeChannel } from "../kubeChannel";
@@ -115,7 +116,7 @@ export class DebugSession implements IDebugSession {
 
                 // Start debug session.
                 p.report({ message: `Starting ${this.debugProvider!.getDebuggerType()} debug session...`});  // safe because checked outside the lambda
-                await this.startDebugSession(appName, cwd, proxyResult);
+                await this.startDebugSession(appName, cwd, proxyResult, podName, undefined);
             } catch (error) {
                 vscode.window.showErrorMessage(error);
                 kubeChannel.showOutput(`Debug on Kubernetes failed. The errors were: ${error}.`);
@@ -145,12 +146,6 @@ export class DebugSession implements IDebugSession {
 
         const { targetPod, targetPodNS, containers } = targetPodInfo;
 
-        // Select the image type to attach.
-        this.debugProvider = await this.pickupAndInstallDebugProvider();
-        if (!this.debugProvider) {
-            return;
-        }
-
         // Select the target container to attach.
         // TODO: consolidate with container selection in extension.ts.
         const targetContainer = await this.pickContainer(containers);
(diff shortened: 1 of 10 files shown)


src/debug/dotNetDebugProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    public async resolvePortsFromFile(dockerfile: IDockerfile, env: Dictionary<string>): Promise<PortInfo | undefined> {
+        return undefined;
+    }
+
+    public async resolvePortsFromContainer(kubectl: Kubectl, pod: string, podNamespace: string | undefined, container: string): Promise<PortInfo | undefined> {
+        return undefined;
+    }
+
+    public filterSupportedProcesses(processes: ProcessInfo[]) : ProcessInfo[] | undefined {
```
tedchamb: If other debug providers implement this, then in the 99% case, you can auto choose the correct debug provider.

src/debug/dotNetDebugProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+           kubeChannel.showOutput(defaultDotnetDebuggerConfigType + " debug attach failed for pod " + pod + ".\nSee https://github.com/Azure/vscode-kubernetes-tools/blob/master/debug-on-kubernetes.md for troubleshooting.", "Failed to attach");
+        }
+        return result;
+    }
+
+    public isSupportedImage(baseImage: string): boolean {
+        // todo: add support for debug from file
+        return false;
+    }
+
```
tedchamb: OmniSharp does not use ports for remote debugging.

src/debug/providerRegistry.ts
```diff
@@ -27,14 +30,38 @@ async function showProviderPick(): Promise<IDebugProvider | undefined> {
...
     let debugProvider = null;
     if (baseImage) {
         debugProvider = supportedProviders.find((provider) => provider.isSupportedImage(baseImage));
     }
-    if (!debugProvider) {
-        debugProvider = await showProviderPick();
+
+    if (!debugProvider)
+    {
+        let candidateProviders: IDebugProvider[];
```
tedchamb: we are filtering the `debugproviders` shown in picker, based on the processes running on the container if we know them.  If `filterSupportedProcesses` returns `undefined` then it will always show that `debugprovider`

src/debug/providerRegistry.ts
```diff
@@ -27,14 +30,38 @@ async function showProviderPick(): Promise<IDebugProvider | undefined> {
...
+                const filteredProcesses = provider.filterSupportedProcesses(runningProcesses);
+                if (!filteredProcesses || filteredProcesses.length > 0) {
+                    candidateProviders.push(provider);
+                }
+            }
+        } else {
+            candidateProviders = supportedProviders
+        }
+
+        if (candidateProviders.length === 0) {
```
patrickcarnahan: should this be 

```
if (candidateProviders.length === 1)
```

tedchamb: yes it should, i"ll fix that

itowlson:
@tedchamb Apologies for the delay on getting back to this - I've been travelling and then dealing with post-travel stuff (and, I fear, shiny toys).  I will try to look at it tomorrow or early next week, and do feel free to remind me if I'm being slack about it.

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-kubernetes-tools",
     "displayName": "Kubernetes",
     "description": "Develop, deploy and debug Kubernetes applications",
-    "version": "1.0.9",
+    "version": "1.0.11",
```
itowlson: We will rev the version as part of a separate publish commit.  You may need to rev the version for testing but we should exclude it from the PR.

itowlson: This should also save having to update `package-lock.json` as part of the PR.

src/debug/debugSession.ts
```diff
@@ -145,39 +146,53 @@ export class DebugSession implements IDebugSession {
...
+        const processes = await getProcesses(this.kubectl, targetPod, targetPodNS, targetContainer);
+
+        // Select the image type to attach.
+        this.debugProvider = await this.pickupAndInstallDebugProvider(undefined, processes);
+        if (!this.debugProvider) {
             return;
         }
 
+        const supportedProcesses = processes ? this.debugProvider.filterSupportedProcesses(processes) : undefined;
+        const pidToDebug = supportedProcesses && supportedProcesses.length === 1 ? +supportedProcesses[0].pid : undefined;
```
itowlson: Suggest pulling out a function here with an intention-revealing name e.g. `extractTargetPID` or `findDebuggeePID` (of type `(ProcessInfo[] | undefined) => (number | undefined)`)

src/debug/debugUtils.ts
```diff
@@ -57,31 +57,40 @@ export async function promptForAppPort(ports: string[], defaultPort: string, env
  * @param container the container name.
  * @return the commands associated with the processes.
  */
-export async function getCommandsOfProcesses(kubectl: Kubectl, pod: string, podNamespace: string | undefined, container: string): Promise<string[]> {
-    const commandLines: string[] = [];
-    const nsarg = podNamespace ? `--namespace ${podNamespace}` : '';
+export async function getProcesses(kubectl: Kubectl, pod: string, podNamespace: string | undefined, container: string | undefined): Promise<ProcessInfo[] | undefined> {
+    const processes: ProcessInfo[] = [];
+    const nsarg = podNamespace && podNamespace !== "default" ? `--namespace ${podNamespace}` : '';
```
itowlson: Is this additional check required?  Feels like it could go wrong if the `kubectl` context namespace was something other than `default`, and the user then tried to debug something in the `default` namespace.  Whereas in the 'normal' case the worst that would happen is that we have a redundant `--namespace default` on the generated `kubectl` command line.  I might be missing something though...?

src/debug/debugUtils.ts
```diff
@@ -57,31 +57,40 @@ export async function promptForAppPort(ports: string[], defaultPort: string, env
...
  */
-export async function getCommandsOfProcesses(kubectl: Kubectl, pod: string, podNamespace: string | undefined, container: string): Promise<string[]> {
-    const commandLines: string[] = [];
-    const nsarg = podNamespace ? `--namespace ${podNamespace}` : '';
+export async function getProcesses(kubectl: Kubectl, pod: string, podNamespace: string | undefined, container: string | undefined): Promise<ProcessInfo[] | undefined> {
+    const processes: ProcessInfo[] = [];
+    const nsarg = podNamespace && podNamespace !== "default" ? `--namespace ${podNamespace}` : '';
     const containerCommand = container ? `-c ${container}` : '';
-    const execCmd = `exec ${pod} ${nsarg} ${containerCommand} -- ps -ef`;
+    const execCmd = `exec ${pod} ${nsarg} ${containerCommand} -- ps -o pid,command -e -w -w`;
```
itowlson: There are two occurrences of `-w` here - is this intentional?

tedchamb: The second `-w` ensures that all columns are displayed despite terminal width.  I can add a comment here.  Probably doesn't matter for this case, because we aren't outputting to a terminal, but helps someone debugging and who wants to run the same command locally and get the same output that gets parsed.

```
     -w      Use 132 columns to display information, instead of the default which is your window size.  If the -w option is specified more than once, ps will use as many columns as
             necessary without regard for your window size.  When output is not to a terminal, an unlimited number of columns are always used.
```

tedchamb: added comment

src/debug/debugSession.ts
tedchamb: fixed

src/debug/debugUtils.ts
tedchamb: I may have added that thinking that a specific namespace name was required rather than the term `default`.  I have removed the check.

tedchamb:
@itowlson , thanks for the review.  I think I have addressed all your comments.  Let me know if not.

Thanks,
Ted

PR #685, opened 2020-01-21
Standardize and expose method for getting the kubeconfig in use

amsoedal:


src/components/config/config.ts
```diff
@@ -90,8 +91,15 @@ export async function setActiveKubeconfig(kubeconfig: string): Promise<void> {
     await addPathToConfig(KUBECONFIG_PATH_KEY, kubeconfig);
 }
 
-export function getActiveKubeconfig(): string {
-    return vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY];
+export function getKubeconfig(): string {
+	let kubeConfig = vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY];
```
daniv-msft: >	 [](start = 0, length = 1)

Nit: it looks like the convention in this repo is to use spaces rather than tabs. Could you please replace these by spaces for consistency?

src/components/config/config.ts
```diff
@@ -90,8 +91,15 @@ export async function setActiveKubeconfig(kubeconfig: string): Promise<void> {
     await addPathToConfig(KUBECONFIG_PATH_KEY, kubeconfig);
 }
 
-export function getActiveKubeconfig(): string {
-    return vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY];
+export function getKubeconfig(): string {
```
daniv-msft: >getKubeconfig [](start = 16, length = 13)

Please consider renaming to "getKubeconfigPath", so that it is clear that this is the path of the kubeconfig and not its content.

daniv-msft:
Just to be sure I understand the process: when will you send the PR on vscode-kubernetes-tools-api to actually surface this new function in the API? As soon as the current PR is checked-in?

src/shell.ts
```diff
@@ -187,7 +187,7 @@ export function shellEnvironment(baseEnvironment: any): any {
         }
     }
 
-    const kubeconfig = getActiveKubeconfig();
+    const kubeconfig = getKubeconfig();
     if (kubeconfig) {
```
daniv-msft: >if (kubeconfig) { [](start = 4, length = 17)

The behavior changed a bit here compared to the previous version of the code, as now this condition will always be true (before we were only looking at the value of vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY]).
@itowlson > Please let us know if you expect this to create any side-effects.

src/explainer.ts
```diff
@@ -6,11 +6,11 @@ import * as pluralize from 'pluralize';
...
 
 import { formatComplex, formatOne, Typed, formatType } from "./schema-formatting";
-import { getActiveKubeconfig, getUseWsl } from './components/config/config';
+import { getKubeconfig, getUseWsl } from './components/config/config';
 
 async function loadKubeconfig(): Promise<kubernetes.KubeConfig> {
     const kc = new kubernetes.KubeConfig();
-    const kubeconfig = getActiveKubeconfig();
+    const kubeconfig = getKubeconfig();
     if (kubeconfig) {
```
daniv-msft: >if (kubeconfig) { [](start = 4, length = 17)

Similar to my other comment: following the change, this is always going to be true. Given that we have specific cases to handle the absence of a value in vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY], I assume this is a behavior we intend to keep.
@itowlson > Could you please provide guidance/context on this? Should we keep using vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY] here?

src/components/config/config.ts
amsoedal: Yes!

src/components/config/config.ts
amsoedal: Good point.

amsoedal:
> Just to be sure I understand the process: when will you send the PR on vscode-kubernetes-tools-api to actually surface this new function in the API? As soon as the current PR is checked-in?

Yes, I was planning to open that PR after this one has been merged.

src/explainer.ts
itowlson: The check here appears to be because of needing to get the kubeconfig from a Windows user's Linux home directory if they have specified to use the WSL version of the tools.  The logic is:

* If the user has overridden kubeconfig in extension settings, then that's what they should get - if they have also asked for WSL then they should jolly well be specifying a WSL file path.
* If the user _hasn't_ overriden kubeconfig in extension settings, and is on WSL, we _don't_ want to use the Windows default path, because that path won't work in WSL (and won't be pointing at the same kubeconfig that other kubectl commands, which _are_ WSLised, use).  We want to use the same default that kubectl uses on Linux (Linux $KUBECONFIG or home dir).
* Otherwise, we use the default that kubectl uses on Windows (Windows %KUBECONFIG% or home dir)

So yes, we need to keep this behaviour.  It's kind of ugly having to drop down to the config-level API from this high level but it may be the most pragmatic design - maybe add a TODO for it and move on.

However, this does make me think there is a potentially nasty can of worms at the API level: consumer extensions which use the `KubectlV1.*` functions get consistent behaviour guaranteed, but consumers who use the kubeconfig path _need to know whether that path relates to the host OS or to WSL_.  It may be that this isn't a consideration for your extension - I don't know what your use case for the path is.  But I think it needs to be visible so that extension authors at least need to handle the possibility, if only to say "sorry, user, we don't offer this feature if you're using the WSL versions of the tools."

So maybe the API needs to return something like an `{ isHostPath: true; path: string } | { isHostPath: false; guestPathType: 'wsl'; guestPath: string }` so that consumers can't naively use the path without at least considering whether and how to handle the WSL scenario...?

src/shell.ts
itowlson: I think this is pretty safe to drop the check.  It may result in a superfluous overwrite of the KUBECONFIG env var but it should be harmless.

itowlson: Of course one of our users will surprise me with some edge case where it does cause a change because that is how kubectl rolls.  _mutter mutter_

src/api/contract/kubectl/v1.ts
```diff
@@ -7,6 +7,7 @@ import * as vscode from 'vscode';
 export interface KubectlV1 {
     invokeCommand(command: string): Promise<KubectlV1.ShellResult | undefined>;
     portForward(podName: string, podNamespace: string | undefined, localPort: number, remotePort: number, options?: KubectlV1.PortForwardOptions): Promise<vscode.Disposable | undefined>;
+    getKubeconfigPath(): string;
```
itowlson: As mentioned in email, I'm anxious even about 'add a member' changes to existing APIs.  A user might pin to an old version of the k8s extension, say because I introduced a bug that affects them in a more recent one (oh, say it ain't so); in which case consumers of KubectlV1 would expect the new method to be available but get an error at runtime.  We should either roll forward to a `KubectlV1_1` interface or put it on a new 'configuration' or 'environment' interface.

src/components/config/config.ts
```diff
@@ -90,8 +91,15 @@ export async function setActiveKubeconfig(kubeconfig: string): Promise<void> {
...
 }
 
-export function getActiveKubeconfig(): string {
-    return vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY];
+export function getKubeconfigPath(): string {
+    let kubeConfig = vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)[KUBECONFIG_PATH_KEY];
+    if (!kubeConfig) {
+        kubeConfig = process.env['KUBECONFIG'];
+    }
+    if (!kubeConfig) {
```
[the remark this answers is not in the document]
amsoedal: Makes sense! Moved to the kubectl/kubeconfig module.

src/shell.ts
amsoedal: I got rid of the check, but please let me know if you think of any such edge case!

src/api/contract/kubectl/v1.ts
amsoedal: Yes! Moved to its own api.

src/components/kubectl/kubeconfig.ts
```diff
@@ -2,21 +2,59 @@ import * as vscode from 'vscode';
...
+import * as shelljs from 'shelljs';
 import { refreshExplorer } from '../clusterprovider/common/explorer';
-import { getActiveKubeconfig } from '../config/config';
+import { getActiveKubeconfig, getUseWsl } from '../config/config';
 
 interface Named {
     readonly name: string;
 }
 
+export function getKubeconfigPath(): { pathType: 'host'; hostPath: string } | { pathType: 'wsl'; wslPath: string }  {
```
daniv-msft: >{ pathType: 'host'; hostPath: string } | { pathType: 'wsl'; wslPath: string }  [](start = 37, length = 78)

Is there any reason not to use `HostKubeconfigPath | WslKubeconfigPath` here as well? If we can avoid it, there is no reason to redeclare these types here. :)

src/components/kubectl/kubeconfig.ts
```diff
@@ -2,21 +2,59 @@ import * as vscode from 'vscode';
...
+        pathType: 'host',
+        hostPath: kubeconfigPath
+    };
+}
+
 export async function mergeToKubeconfig(newConfigText: string): Promise<void> {
-    const kcfile = kubeconfigPath();
-    if (!(await fs.existsAsync(kcfile))) {
-        vscode.window.showErrorMessage("Couldn't find kubeconfig file to merge into");
+    const config = getKubeconfigPath();
```
daniv-msft: >config [](start = 10, length = 6)

Nit: we have this pattern multiple times, where we first call getKubeconfigPath and then reads its result the next line.
Because "config" is an ambiguous name, I'd suggest to go for "kubeconfigPath" instead. I agree that it's similar to the other "kubeconfigFilePath" variable declared, but I think that's OK, and looks better than using "config" in my opinion.

src/components/kubectl/autoversion.ts
```diff
@@ -101,11 +101,12 @@ async function ensureCacheIsForCurrentKubeconfig(): Promise<void> {
 }
 
 async function getKubeconfigPathHash(): Promise<string | undefined> {
-    const kubeconfigPath = getActiveKubeconfig() || process.env['KUBECONFIG'] || path.join(shell.home(), '.kube/config');
-    if (!await fs.existsAsync(kubeconfigPath)) {
+    const kubeconfigPath = getKubeconfigPath();
+    const kubeconfigFilePath = kubeconfigPath.pathType === "host" ? kubeconfigPath.hostPath : kubeconfigPath.wslPath;
+    if (!await fs.existsAsync(kubeconfigFilePath)) {
```
itowlson: This has surfaced an issue in this procedure - in WSL mode, the path will be a Unix path, but `fs.existsAsync` will still look in the host filesystem.  I don't think you need to fix this, as the bug was already there but latent, and the impact is to defeat caching rather than to do anything calamitous, but I wanted to note it for future work.

src/components/kubectl/kubeconfig.ts
```diff
@@ -2,21 +2,60 @@ import * as vscode from 'vscode';
...
+import { ConfigurationV1 } from '../../api/contract/configuration/v1';
 import { refreshExplorer } from '../clusterprovider/common/explorer';
-import { getActiveKubeconfig } from '../config/config';
+import { getActiveKubeconfig, getUseWsl } from '../config/config';
 
 interface Named {
     readonly name: string;
 }
 
+export function getKubeconfigPath(): ConfigurationV1.KubeconfigPath  {
```
itowlson: I feel like internal functions shouldn't be bound to API signatures.  I'd prefer `kubeconfig.ts` to define a canonical internal representation, and the API surfaces that private representation that (either, as in this case, directly via a compatible interface definition, or indirectly via an adapter).  Until and unless we can make the API so foundational that _even the core extension_ can be built out of the API, I feel that this is a mixing of layers: currently the API sits over the top of the internal code, rather than the internal code interoperating via the API, if that makes sense.

src/components/kubectl/kubeconfig.ts
```diff
@@ -2,21 +2,60 @@ import * as vscode from 'vscode';
...
+    };
+}
+
 export async function mergeToKubeconfig(newConfigText: string): Promise<void> {
-    const kcfile = kubeconfigPath();
-    if (!(await fs.existsAsync(kcfile))) {
-        vscode.window.showErrorMessage("Couldn't find kubeconfig file to merge into");
+    const kubeconfigPath = getKubeconfigPath();
+    const kubeconfigFilePath = kubeconfigPath.pathType === "host" ? kubeconfigPath.hostPath : kubeconfigPath.wslPath;
+    if (!(await fs.existsAsync(kubeconfigFilePath))) {
```
itowlson: Again this will go awry if they are on WSL, and again I don't think this is your problem to fix.  I am loving that this work gives us a really clear way to see where the WSL case has been overlooked!  Thank you!

src/components/kubectl/kubeconfig.ts
```diff
@@ -2,21 +2,60 @@ import * as vscode from 'vscode';
...
+        vscode.window.showErrorMessage(`Couldn't find kubeconfig file to merge into: '${kubeconfigFilePath}'`);
         return;
     }
 
-    const kubeconfigText = await fs.readTextFile(kcfile);
+    const kubeconfigText = await fs.readTextFile(kubeconfigFilePath);
```
itowlson: I know that the `kcfile` naming was not great but I would suggest reinstating it so we don't see so many diffs here.  We can change it in a separate renaming pass if we feel strongly.

src/api/contract/configuration/v1.ts
```diff
@@ -0,0 +1,23 @@
+// This module is contractual and should not be changed after release.
+// It should be in sync with vscode-kubernetes-tools-api/ts/configuration/v1.ts
+// at all times.
+
+export interface ConfigurationV1 {
```
itowlson: You will need to update `api/apibroker.ts` to expose the new interface category on the extension API object.  `apibroker` is the secret handshake between a consumer of the extension API and the various interface implementations - I'm afraid it's not very visible and I should have flagged it when we first talked about making a new interface category.  Sorry!

src/components/kubectl/kubeconfig.ts
itowlson: I don't feel super strongly about this, and there is absolutely no harmful impact of leaving the code as you have it, so feel free to WONTFIX this comment!

src/components/kubectl/autoversion.ts
amsoedal: Added a TODO!

src/components/kubectl/kubeconfig.ts
amsoedal: Thanks Ivan! :)  Added another comment for future work

src/api/contract/configuration/v1.ts
amsoedal: Ahh that makes sense! Thanks for pointing that out

PR #700, opened 2020-02-11
Delete namespace

### LINKED ISSUE #696: add "delete namespace"
I mentioned this to @squillace, who advised I create an issue. The one thing I'd like to see added to the extension (which is **amazingly awesome**, btw), is right-click-delete support for namespaces. Ideally, it would bark when the user tried it on a non-empty namespace, advising them to clear out said namespace prior to deletion. But for dev/test it'd be rad to have the delete namespace capability so customers who are rapidly testing out environmental configuration and deployment could blow away their unused namespaces once they're empty. 

Love the extension - literally makes K8S development with .NET a breeze. I wish more people knew about it!

itowlson:
Fixes #696.

This will always give the standard resource confirmation, but if the namespace contains other resources, or is the `default` namespace, then it requests additional confirmation.  Also if the user deletes the current namespace then we try to switch back to the `default` namespace so as not to be left with no valid active namespace in the kubeconfig.

cc @bradygaster

### DIFF
--- a/package.json
+++ b/package.json
@@ -467,7 +467,7 @@
                 {
                     "command": "extension.vsKubernetesDelete",
                     "group": "2@2",
-                    "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource(?!\\.namespace).*/i"
+                    "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource.*/i"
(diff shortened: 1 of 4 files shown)


bradygaster:
Awesomesauce! Thanks so much, @itowlson and team! I'll make sure to set up some time to show you the use-case I'd envisioned.

Issue #699, opened 2020-02-11
Snippet warning on load in latest build


itowlson:
Current `master` gives the warning:

_[ms-kubernetes-tools.vscode-kubernetes-tools]: One or more snippets from the extension 'vscode-kubernetes-tools' very likely confuse snippet-variables and snippet-placeholders (see https://code.visualstudio.com/docs/editor/userdefinedsnippets#_snippet-syntax for more details)_

This may be a result of the snippet changes since last release e.g. #676 - I need to check and tweak before we do another release.

itowlson:
Confirmed that this issue started after I merged #676, but I can't see what in that changeset it's complaining about: it all looks good to me!  @amirschw do you see this warning?  Can you spot what might be causing it?

itowlson:
Ah, I think it's the regeneration of the `range-*` snippet JSON.  E.g. previously `helm.json` line 192 read `"{{- range \\$i, \\$val := ${1:list} }}"` but now that last bit is `${1:$list} }}` (note the extra `$` in front of `list`).  I suspect this error occurred before and was 'fixed' by editing the generated `helm.json` - and when @amirschw regenerated it, that kludge got overwritten.  So I need to fix this at source in the `rawsnippets/*.yaml` files.

itowlson:
Closed via #701.

PR #701, opened 2020-02-11
Fix snippet syntax warning

itowlson:
We had a minor syntax error (text getting treated as variable references) in some of our Helm template snippets, which had been fixed in the generated `snippets/helm.json` but had not been fixed in the templates from which the JSON was generated.  This meant that when someone did things right and regenerated the JSON, which they did in #676, the error reappeared.

This PR fixes the error at source, so that future regenerations should not cause this to recur.


itowlson:
Fixes #699.

Issue #593, opened 2019-07-30
Follow logs webview not autoscrolling


itowlson:
Per https://github.com/Azure/vscode-kubernetes-tools/issues/460#issuecomment-516176958, the Follow Logs webview is not autoscrolling as new logs arrive.

cc @brendandburns - I am happy to tackle this if you want though

itowlson:
@finsterdexter I'm looking at this issue now and would value thoughts on the desired behaviour.  Presumably if you have found an interesting log and scrolled to it, you do _not_ want the next log message in Follow mode to cause the view to scroll back to the bottom?  So should the rule be (in Follow mode):

1. If the user has not scrolled up, keep the view tracking the most recent log messages.
2. If the user has scrolled up, do not scroll automatically - continue to append logs but leave the scroll position where it is.
3...? (maybe) If the user has scrolled up but has since scrolled back to the end, resume tracking?

Thoughts?

finsterdexter:
For my use case, what you outline there is exactly what happens in iterm and the vscode terminal, so that makes sense to me.

bfitzpat:
May I propose this simple solution we came up with locally? It may not be the cleanest solution but may offer a route for you to explore. https://github.com/Azure/vscode-kubernetes-tools/pull/704

theoharr:
Any update on this one? The follow logs function isn't really usable without the autoscrolling.

lstocchi:
Closing it as https://github.com/Azure/vscode-kubernetes-tools/pull/704 was merged and it is now part of the new version 1.2.1. Feel free to reopen this if you still face any problem with the autoscrolling :)

PR #704, opened 2020-02-13
issue #593 - adding simple auto-scroll checkbox to enable and disable

bfitzpat:
Signed-off-by: bfitzpat@redhat.com <bfitzpat@redhat.com>

### DIFF
--- a/src/components/logs/logsWebview.ts
+++ b/src/components/logs/logsWebview.ts
@@ -64,6 +64,8 @@ export class LogsPanel extends WebPanel {
                 </select>
                 <span style='position: absolute; left: 240px'>Match expression</span>
                 <input style='left:350px; position: absolute' type='text' id='regexp' onkeyup='eval()' placeholder='Filter' size='25'/>
+                <span style='position: absolute; left: 600px'>Auto-scroll</span>
+                <input style='left: 675px; position: absolute' type='checkbox' id='autoscroll' onkeyup='eval()' checked="true"/>
             </div>
             <div style='position: absolute; top: 55px; bottom: 10px; width: 97%'>
               <div style="overflow-y: scroll; height: 100%">
@@ -85,13 +87,13 @@ export class LogsPanel extends WebPanel {
                 return filter(logsText, true);
               }
 
-              const filter = (text, isNewLog) => {   
+              const filter = (text, isNewLog) => {
                 const regexp = document.getElementById('regexp').value;
                 const mode = document.getElementById('mode').value;
                 let content;
                 if (regexp.length > 0 && mode !== 'all') {
-                    const regex = new RegExp(regexp);   
-                    switch (mode) {                        
+                    const regex = new RegExp(regexp);
+                    switch (mode) {
                         case 'include':
                             content = text.filter((line) => regex.test(line));
                             break;
@@ -100,7 +102,7 @@ export class LogsPanel extends WebPanel {
                             break;
                         case 'before':
                             content = [];
-                            if (!isNewLog) { 
+                            if (!isNewLog) {
                                 for (const line of text) {
                                     if (regex.test(line)) {
                                         break;
@@ -117,7 +119,7 @@ export class LogsPanel extends WebPanel {
                                     return regex.test(line)
                                 });
                                 content = text.slice(i+1);
-                            }                           
+                            }
                             break;
                         default:
                             content = []
@@ -126,7 +128,7 @@ export class LogsPanel extends WebPanel {
                 } else {
                     content = text;
                 }
-                
+
                 return content;
               };
 
@@ -137,11 +139,11 @@ export class LogsPanel extends WebPanel {
                 return beautifyLines(contentLines);
               }
 
-              const beautifyLines = (contentLines) => {                
+              const beautifyLines = (contentLines) => {
                 let content = contentLines.join('\\n');
                 if (content) {
                     content = content.match(/\\n$/) ? content : content + '\\n';
-                }                
+                }
                 return content;
               };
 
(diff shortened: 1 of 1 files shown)


bfitzpat:
Though this may not be the best solution to this issue, it provides a simple solution that may be enough to address the scrolling problem. It's something we came up with locally and thought you might find useful.

src/components/logs/logsWebview.ts
```diff
@@ -80,6 +82,12 @@ export class LogsPanel extends WebPanel {
                     });
                     // TODO: need to apply filters here!
                     elt.appendChild(document.createTextNode(message.text));
+
+                    // handle auto-scroll on/off
+                    var checkBox = document.getElementById("autoscroll");
+                    if (checkBox.checked == true){
+                        document.getElementById('bottom').scrollIntoView();
```
itowlson: This element doesn't seem to be present in the HTML.  I found if I added a `<p id='bottom'></p>` after the `<code>...</code>` around line 59-62 then it worked, but that was only really simple testing...

src/components/logs/logsWebview.ts
```diff
@@ -51,6 +51,8 @@ export class LogsPanel extends WebPanel {
                 </select>
                 <span style='position: absolute; left: 240px'>Match expression</span>
                 <input style='left:350px; position: absolute' type='text' id='regexp' onkeyup='eval()' placeholder='Filter' size='25'/>
+                <span style='position: absolute; left: 600px'>Auto-scroll</span>
+                <input style='left: 675px; position: absolute' type='checkbox' id='autoscroll' onkeyup='eval()' checked="true"/>
```
itowlson: This makes the checkbox appear even in Show Logs (non-tailing) mode.  I'm not sure this makes sense.  Could we hide the caption and check box in that mode?  (The way it's done at the moment, the panel doesn't know whether the user is showing or following, so we'd need the controlling command to pass that info in I guess.)

src/components/logs/logsWebview.ts
bfitzpat: Sorry about missing that in my first pass, but I have added it back in.

src/components/logs/logsWebview.ts
bfitzpat: We just have it showing up in all cases currently. There wasn't a clear way to identify when we were coming in via show or follow.

bfitzpat:
If the log is extremely long, it seems to get hung up sometimes (I think due to the code in eval() it has issues processing super-long logs at times), but otherwise it works great as a very simple solution.

lstocchi:
Played with it and it worked fine for me but i have a doubt...by reading https://github.com/Azure/vscode-kubernetes-tools/issues/593#issuecomment-516672587 it looks like this doesn't cover what we really need. If you scroll up and a new log is written it takes you to the bottom. It can be quite frustrating.
Why don't we listen to `scroll`? Has it been tested before? Are there any issues?
My idea is this.

Have a variable that stores a boolean which specify if the user is at the end of the page. If true the window need to stick to the bottom when new logs are added, if false it means user scroll up.
```
let isToBottom = true;   
if (isToBottom) document.getElementById('bottom').scrollIntoView();
```

To set the variable we listen to the scroll event
```
<div id="logPanel" style="overflow-y: scroll; height: 100%">
```
```
const elem = document.getElementById('logPanel');
let lastScrollTop = 0;
const toBottom = debounce(function() {
        const st = elem.scrollTop;
        if (st > lastScrollTop){
             // scroll down
             isToBottom = (elem.scrollTop + window.innerHeight) >= elem.scrollHeight;
        } else {
             // scroll up
            isToBottom = false;
         }
            lastScrollTop = st <= 0 ? 0 : st; 
  }, 250);

elem.addEventListener("scroll", toBottom);
```
where the debounce function is taken from https://davidwalsh.name/javascript-debounce-function 
so we don't always fire the function.

```
// Returns a function, that, as long as it continues to be invoked, will not
// be triggered. The function will be called after it stops being called for
// N milliseconds. If `immediate` is passed, trigger the function on the
// leading edge, instead of the trailing.
function debounce(func, wait, immediate) {
	var timeout;
	return function() {
		var context = this, args = arguments;
		var later = function() {
			timeout = null;
			if (!immediate) func.apply(context, args);
		};
		var callNow = immediate && !timeout;
		clearTimeout(timeout);
		timeout = setTimeout(later, wait);
		if (callNow) func.apply(context, args);
	};
};
```

If this all works we shouldn't need the checkbox but we should have a `go to the bottom` button so if there are many logs one doesn't have to scroll down for hours...

itowlson:
@lstocchi This sounds like a good approach - are you able to take it on, or @bfitzpat do you want to take a crack at Luca's suggestion?

bfitzpat:
User cases discussed in meeting:
1. User-specified path to binary. (don't have currently)
2. System Path-specified binary.
3. If neither 1 or 2, download latest released version from github.

bfitzpat:
> @lstocchi This sounds like a good approach - are you able to take it on, or @bfitzpat do you want to take a crack at Luca's suggestion?

I can take a look.

lstocchi:
Hey @bfitzpat, i noticed a couple of errors.
First is `window.addEventListener("scroll", toBottom);`. It should be `elem.addEventListener("scroll", toBottom)`. The scroll is inside that div, not in the window.
The second (my bad) when creating the `toBottom` function it is saved inside a constant `const toBottom = debounce(function() {` but the debouce function returns a function so it must be `let toBottom = debounce(function() {`
After these changes it should be fine.

bfitzpat:
@lstocchi that's awesome. Those two fixes and we were off to the races. Thanks! 

I'll see about adding a button that scrolls to the bottom from the tools area at the top of the log window next!

lstocchi:
I guess this is done. I tested it and it works, i couldn't test it with all different use cases i had in mind because there are other PRs still opened related to the logView but this looks good to me. @itowlson WDYT?

I don't know why travis still shows it's pending but it's finished

lstocchi:
@itowlson Tested it with the new release. Works fine!! Thank you.

Thanks @bfitzpat for your contribution :+1:

PR #712, opened 2020-03-13
add new api to let external extensions have access to watch feature

### LINKED ISSUE #711: Add new api to support Watch
After merging #686 users can now watch resources and see their tree refreshed automatically. As discussed in the PR, It would be nice if the extension also allowed external extensions that extend this one to have access to this feature.

lstocchi:
it fixes #711 

@itowlson i don't know if you were already working on this or if you need to add other stuff to the new api. If this is the case feel free to close it. 
I did not update the doc, i am waiting for your comment and if everything is ok i will work on that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ Minikube tools to be installed and available on your PATH.
        * `vs-kubernetes.knownKubeconfigs` - An array of file paths of kubeconfig files that you want to be able to quickly switch between using the Set Kubeconfig command.
        * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created deployment and associated pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up if you wanted to view pod logs, etc.
        * `vs-kubernetes.outputFormat` - The output format that you prefer to view Kubernetes manifests in. One of "yaml" or "json". Defaults to "yaml".
-       * `vs-kubernetes.resources-to-watch` - List of resources to be watched. To identify a resource the extension uses the label displayed in the cluster explorer. I.e ["Pods", "Services", "Namespaces"]. 
+       * `vs-kubernetes.resources-to-watch` - List of resources to be watched. To identify a resource the extension uses the label displayed in the cluster explorer. E.g. ["Pods", "Services", "Namespaces"]. 
        * `logsDisplay` - Where and how to display Kubernetes logs.  One of `webview` (display in a filterable HTML view) and `terminal` (run the command in the VS Code terminal)
    * `vsdocker.imageUser` - Image prefix for the container images e.g. 'docker.io/brendanburns'
    * `checkForMinikubeUpgrade` - On extension startup, notify if a minikube upgrade is available. Defaults to true.
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Features include:
 * Create Helm charts using scaffolding and snippets.
 * Bootstrap applications using Draft, and rapidly deploy and debug them to speed up
   the development loop.
+* Watch resources in the cluster explorer and get live updates as they change
 
 **What's new in this version?**  See the [change log](CHANGELOG.md) to find out!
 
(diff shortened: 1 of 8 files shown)


itowlson:
@lstocchi I'm not working on it - thanks for thinking of this!

Unfortunately the way I put together the API files makes it hard for me to see the diff here... doh!  (This is me not thinking ahead, not a gripe at you!)  I saw you added an optional `apiName` to the resource node source - was there anything else?  If not, we should maybe consider making this `V1_1` instead of `V2`, as it remains backward compatible with existing V1 code - what do you think?

lstocchi:
@itowlson No there was just the `apiName` to add. At least for the part related to the `Watch` feature. I created this as `V2` because in the last PR you mentioned it. Not sure if you had something else in mind to add to it. If it's just this, we can definitely change it as `V1_1`

itowlson:
Great, thanks!  No, I wasn't thinking of anything else, just the way I set this up made it hard for me to be sure I had spotted all the diffs.  Something I need to think about for future versions!

lstocchi:
@itowlson doc updated. Give it a look whenever you can. Thank you.

README.md
```diff
@@ -205,6 +208,7 @@ Minikube tools to be installed and available on your PATH.
        * `vs-kubernetes.knownKubeconfigs` - An array of file paths of kubeconfig files that you want to be able to quickly switch between using the Set Kubeconfig command.
        * `vs-kubernetes.autoCleanupOnDebugTerminate` - The flag to control whether to auto cleanup the created deployment and associated pod by the command "Kubernetes: Debug (Launch)". The cleanup action occurs when it failed to start debug session or debug session terminated. If not specified, the extension will prompt for whether to clean up or not. You might choose not to clean up if you wanted to view pod logs, etc.
        * `vs-kubernetes.outputFormat` - The output format that you prefer to view Kubernetes manifests in. One of "yaml" or "json". Defaults to "yaml".
+       * `vs-kubernetes.resources-to-watch` - List of resources to be watched. To identify a resource the extension uses the label displayed in the cluster explorer. I.e ["Pods", "Services", "Namespaces"].
```
itowlson: Super pedantic but `e.g.` not `i.e.` for examples... sorry!

```suggestion
       * `vs-kubernetes.resources-to-watch` - List of resources to be watched. To identify a resource the extension uses the label displayed in the cluster explorer. E.g. ["Pods", "Services", "Namespaces"]. 
```

docs/extending/clusterexplorernodes.md
```diff
@@ -193,6 +193,34 @@ a `NodeSource`, then use its `at` method to specify where to attach to the tree.
 The `at` method may specify an existing grouping folder (specified by display name), or
 `undefined` to appear directly under the context tree node.
 
+**`API v1_1`**
```
itowlson: Switching to an "API v1_1" heading here feels disorienting to me.  I suggest we leave the original content and flow, except for updating the code sample to include the `apiName` as you have done, and then add a note that the `apiName` argument is not available in v1.

The principle I'm thinking of it that the docs should reflect how _someone writing an extension now_ would use the API, with backward compatibility notes for people maintaining extensions that they don't want to port to a new API version.

That would mean we should shift the 'how to get the API object' text to do `await k8s.extension.clusterExplorer.v1_1` instead of `.v1`.  What do you reckon?

docs/extending/README.md
```diff
@@ -70,7 +70,7 @@ snippet. In terms of the TypeScript typings from the NPM package, `activate` ret
   * Versions: `v1`
 * [Cluster Explorer API](clusterexplorer.md) - used for working with the Clusters tree
   * ID: `clusterexplorer`
-  * Versions: `v1`
+  * Versions: `v1`, `v1_1`
```
itowlson: Wondering if this should be `v1.1` instead of `v1_1`.  We have to use the underscore in the JavaScript names because the dot has special meaning, but the way people would normally refer to it is `1.1`.

README.md
lstocchi: Shame on me ... i can't even use the excuse i'm italian because they are Latin abbreviation, damn :grin:

docs/extending/clusterexplorernodes.md
lstocchi: Yeah, sounds good. I also changed the sentence where we explain what the `apiName` arg is used for. It's clearer now.

src/api/contract/cluster-explorer/v1_1.ts
```diff
@@ -0,0 +1,107 @@
+// This module is contractual and should not be changed after release.
+// It should be in sync with vscode-kubernetes-tools-api/ts/explorer-tree/v2.ts
```
itowlson, line 2:
```suggestion
// It should be in sync with vscode-kubernetes-tools-api/ts/explorer-tree/v1_1.ts
```

lstocchi:
We need to rememeber to also release new version of https://github.com/Azure/vscode-kubernetes-tools-api.

PR #714, opened 2020-03-17
Refactoring kubectl invocation to better support silence

itowlson:
The current `kubectl` invocation infrastructure has UI interactions such as 'binary not found', 'install dependencies' etc. deep in the fundamental path.  This can result in situations where the extension accidentally activates (e.g. because the user is working with a non-Kubernetes YAML file) and the user gets spammed with repeated notifications and prompts.

This PR refactors invocation to better support silent invocation, and to give the application level finer grained control over when to engage in UI interactions.  It does not necessarily reduce noise _right now_ because the main purpose is to restructure invocation with minimal risk, but it is necessary infrastructure for reducing noise in the forthcoming sequel.


itowlson:
@bnookala @squillace (and anyone else) it would be great if you could have a play with this and let me know if you see any regressions.  It's a bit of a beast and probably tricky to review unfortunately...

src/docker/dockerUtils.ts
```diff
@@ -14,7 +14,7 @@ import { Dictionary } from "../utils/dictionary";
 export async function resolveKubernetesDockerEnv(kubectl: Kubectl): Promise<{}> {
     const dockerEnv = Dictionary.of<string | number>();
     dockerEnv["DOCKER_API_VERSION"] = await dockerApiVersion();
-    const currentCluster = await getCurrentClusterConfig(kubectl);
+    const currentCluster = await getCurrentClusterConfig(kubectl, {});  // TODO: should this be silent
```
lstocchi, line 17: TODO here. Should be {silent: true}

src/components/installer/installdependencies.ts
```diff
@@ -0,0 +1,54 @@
...
+    const installPromises = [
+        installDependency("kubectl", gotKubectl, installKubectl),
+        installDependency("Helm", gotHelm, (sh) => installHelm(sh, warn)),
+        installDependency("Draft", gotDraft, installDraft),
+    ];
+
+    if (!config.getUseWsl()) {
+        // TODO: Install Win32 Minikube
+        installPromises.push(
+            installDependency("Minikube", gotMinikube, (shell: Shell) => {
```
lstocchi, line 35: To be coherent with the other lines better having `(sh) => installMinikube(sh, null)` ??

src/binutilplusplus.ts
```diff
@@ -0,0 +1,371 @@
...
+            linesSubject.next(line);
+        }
+    });
+    return { lines: linesSubject };
+}
+
+export async function invokeBackground(context: Context, args: string[]): Promise<BackgroundExecResult> {
+    const fbr = await findBinary(context);
+    if (!fbr.found) {
+        return { resultKind: 'exec-bin-not-found', execProgram: context.binary, command: args.join(' '), findResult: fbr };
```
lstocchi, line 286: Wouldn't it be better to have constants instead of writing strings every time? (exec-failed, exec-errorer, exec-succeeded...)

src/components/kubectl/logs.ts
```diff
@@ -113,19 +115,20 @@ async function getLogsForContainer(
...
-        cmd = `${cmd} -f`;
-        kubectl.invokeAsync(cmd, undefined, (proc: ChildProcess) => {
-            proc.stdout.on('data', (data: string) => {
-                panel.addContent(data);
-            });
-        });
+        const lines = await kubectl.observeCommand(args);
+
+        lines.subscribe(
+            (line) => { if (line) { panel.addContent(line + '\n'); } },
```
lstocchi, line 121: `${line} \n`

src/binutilplusplus.ts
itowlson: I did this for something else on another branch and didn't feel it was terribly worth the effort - because these are not general-purpose strings but statically checked instances of the `exec-bin-not-found` (or whatever) _type_, you get compiler checking anyway.  But yeah, would be easier to change if they were defined as consts of the relevant type.

itowlson:
@lstocchi Thanks!  I addressed the comments except for the one about named constants, where I am not sure about the benefits vs folderol of the change, given that the compiler verifies the values anyway...?

lstocchi:
> @lstocchi Thanks! I addressed the comments except for the one about named constants, where I am not sure about the benefits vs folderol of the change, given that the compiler verifies the values anyway...?

I was thinking about constants just to prevent any typos when writing new code or, if in future we need to change any of those string, we don't need to go all over the code. But it's not a requirement at all. Just to make you think on this :+1:

src/components/kubectl/logs.ts
```diff
@@ -113,19 +115,20 @@ async function getLogsForContainer(
...
-        cmd = `${cmd} -f`;
-        kubectl.invokeAsync(cmd, undefined, (proc: ChildProcess) => {
-            proc.stdout.on('data', (data: string) => {
-                panel.addContent(data);
-            });
-        });
+        const lines = await kubectl.observeCommand(args);
+
+        lines.subscribe(
+            (line) => { if (line) { panel.addContent(`${line}\n`); } },
```
lstocchi, line 121: Here we have the same issue we have now. The process remains alive even if the log panel is disposed and it keeps calling the `addContent` method. What worries me is not the error itself because this PR is not related to that one and it can be fixed in a second moment but i am not sure the observeCommand gives us a reference to keep track of the childProcess created (???)

itowlson: Good catch - thank you.  I'll update `observeCommand` to track the child process.  I expect your fix to merge first, so will need to update the logs stuff in this PR once I rebase.

itowlson: Grr.  `spawnrx` just gives us the output, not the process object.  But the docs say that unsubscribing from the output stream terminates the process, so I think we can mimic this closely enough.

src/binutilplusplus.ts
```diff
@@ -0,0 +1,371 @@
...
+            resultKind: execResult.resultKind,
+            execProgram: execResult.execProgram,
+            command: execResult.command,
+            result: fn(execResult.stdout.trim())
+        };
+    }
+
+    export function failureMessage(execResult: FailedExecResult, options: FailureMessageOptions): string {
+        const err = ExecResult.tryMap(execResult, (s) => s);
+        if (Errorable.failed(err)) {
```
lstocchi, line 103: I think this check can be removed. We are dealing with `FailedExecResult` so no need to verify if they are failed.

src/helm.exec.ts
```diff
@@ -4,22 +4,24 @@ import { helm as logger } from './logger';
 import * as YAML from 'yamljs';
 import * as _ from 'lodash';
 import * as tmp from 'tmp';
-import * as extension from './extension';
 import * as helmrepoexplorer from './helm.repoExplorer';
 import * as helm from './helm';
 import { showWorkspaceFolderPick } from './hostutils';
-import { shell as sh, ShellResult, ExecCallback } from './shell';
+import { shell as sh, ShellResult, ExecCallback, shell } from './shell';
```
lstocchi, line 10: Double import here -> `shell as sh` and `shell`

src/helm.exec.ts
```diff
@@ -570,6 +572,32 @@ export async function helmExecAsync(args: string): Promise<ShellResult | undefin
...
+const HELM_CONTEXT: Context = {
+    host: host,
+    fs: shellfs,
+    shell: shell,
+    pathfinder: undefined,
+    binary: HELM_BINARY,
+    status: undefined,
+};
+
+export async function helmInvokeCommand(command: string): Promise<ExecResult> {
```
lstocchi, line 591: This function is never called. Can it be deleted?

src/extension.ts
```diff
@@ -755,22 +770,23 @@ function loadKubernetesCore(namespace: string | null, value: string) {
     (err) => vscode.window.showErrorMessage(`Error loading document: ${err}`));
 }
 
-function exposeKubernetes() {
-    findKindNameOrPrompt(kuberesources.exposableKinds, 'expose', { nameOptional: false}, (kindName: string) => {
-        if (!kindName) {
-            vscode.window.showErrorMessage('couldn\'t find a relevant type to expose.');
```
lstocchi, line 761: This error message has been removed. Not sure if it's correct.

src/binutilplusplus.ts
itowlson: We could convert it to an assertion, but the compiler can't prove that passing a `FailedExecResult` to `tryMap` will result in a `Failed` Errorable, so we need _something_ to tell the compiler that it can use the error value of the Errorable.  I could write `as Failed` but the `if` check just seemed a bit more belt and braces.

src/helm.exec.ts
itowlson: Oof, good catch.  Thanks!

src/helm.exec.ts
itowlson: It could be, but many of the other Helm functions will be ported over to use it (I think I must have started doing that and then backed them out because I was doing too much at once).

src/extension.ts
itowlson: I believe the removal is correct.  `findKindNameOrPrompt` should return falsy only if we prompted _and_ the user cancelled.  And of course cancellation shouldn't result in a notification.

lstocchi:
@itowlson i finally finished to look at everything. Sorry it tooks so long but i was quite busy and this PR is huge. All the things i commented previously are the only ones i found. For the rest it looks good for me :+1:

itowlson:
Thanks heaps @lstocchi - appreciate all your time and attention on this!

Issue #725, opened 2020-04-03
Error when opening "show log" webview after having already opened and closed it


lstocchi:
I found an error when trying to open the `show logs` webview for the second time after having already opened and closed it. 

![image](https://user-images.githubusercontent.com/49404737/78385261-de888d00-75db-11ea-8513-06728fdf4e61.png)

The issue sits in the `currentPanels` map. When the new webview is created it is added to the map with a key `namespace/resource/name`. When the panel is closed, the code tries to delete the key `resource/name`, so it fails and when you try to reopen the webview, there is a hit in the map (like the webview is still opened) and it tries to update the old webview which was disposed.

lstocchi:
Closing it as https://github.com/Azure/vscode-kubernetes-tools/pull/726 was merged

PR #726, opened 2020-04-06
fixed disposed error logpanel + kill useless process

lstocchi:
This PR addresses a couple of issues i found with the logPanel. First is #725. Second is related to the listener created to append new content when following logs. The process created was never killed.

### DIFF
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -116,6 +116,7 @@ async function getLogsForContainer(
 
     if (displayMode === LogsDisplayMode.Follow) {
         const followProcess = await kubectl.observeCommand(args);
+        panel.setAppendContentProcess(followProcess);
 
         // TODO: during rebase, we will need to also provide the followProcess.terminate method to the viewer
         followProcess.lines.subscribe(
@@ -131,7 +132,8 @@ async function getLogsForContainer(
         if (ExecResult.failed(result)) {
             kubectl.reportFailure(result, { whatFailed: 'Error reading logs' });
         } else {
-            panel.setInfo(result.stdout, containerResource.kindName);
+            panel.deleteAppendContentProcess();
+            panel.setContent(result.stdout);
         }
     } catch (err) {
         vscode.window.showErrorMessage(`Error reading logs ${err}`);
--- a/src/components/logs/logsWebview.ts
+++ b/src/components/logs/logsWebview.ts
@@ -1,9 +1,12 @@
 import * as vscode from 'vscode';
 import { WebPanel } from '../webpanel/webpanel';
+import { RunningProcess } from '../../binutilplusplus';
 
 export class LogsPanel extends WebPanel {
     public static readonly viewType = 'vscodeKubernetesLogs';
     public static currentPanels = new Map<string, LogsPanel>();
+
+    public appendContentProcess: RunningProcess | undefined;
 
     public static createOrShow(content: string, resource: string): LogsPanel {
         const fn = (panel: vscode.WebviewPanel, content: string, resource: string): LogsPanel => {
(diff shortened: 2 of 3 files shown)


src/components/webpanel/webpanel.ts
```diff
@@ -1,19 +1,18 @@
 import * as vscode from 'vscode';
+import { LogsPanel } from '../logs/logsWebview';
```
itowlson, line 2: `WebPanel` is also used by the Describe webview; it shouldn't depend on `LogsPanel`.

src/components/webpanel/webpanel.ts
```diff
@@ -1,19 +1,18 @@
...
 
     protected static createOrShowInternal<T extends WebPanel>(content: string, resource: string, viewType: string, title: string, currentPanels: Map<string, T>, fn: (p: vscode.WebviewPanel, content: string, resource: string) => T): T {
         const column = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.viewColumn : undefined;
 
         // If we already have a panel, show it.
         const currentPanel = currentPanels.get(resource);
         if (currentPanel) {
-            currentPanel.setInfo(content, resource);
-            currentPanel.update();
+            currentPanel.setContent(content);
```
itowlson, line 15: Not sure whether it's safe to remove the resource, or to remove the call to `currentPanel.update()` in the next line.  These might be required for `DescribePanel`.

src/components/kubectl/logs.ts
```diff
@@ -127,7 +128,8 @@ async function getLogsForContainer(
         if (!result || result.code !== 0) {
             vscode.window.showErrorMessage(`Error reading logs: ${result ? result.stderr : undefined}`);
         } else {
-            panel.setInfo(result.stdout, containerResource.kindName);
+            panel.deleteAppendContentProcess();
```
itowlson, line 131: Not sure what is going on here.  In this code path the `kubectl` call exits, so there is no child process to delete.  Does this relate to the case where the user was previously doing Follow Logs for a pod and has now done a Show Logs, and we are trying to reuse the webview so we have to clear out the previous follow-related state?  If so perhaps explain this in a comment?

I kind of feel that maybe this shouldn't be a client problem.  If the webview is hanging onto state, then the webview should be responsible for cleaning it up - perhaps the LogsPanel should have two distinct methods, one to start a follow and one to perform a show, and those methods should handle the necessary state management?

src/components/logs/logsWebview.ts
```diff
@@ -22,11 +25,21 @@ export class LogsPanel extends WebPanel {
 
     public addContent(content: string) {
         this.content += content;
-        if (this.canProcessMessages) {
-            this.panel.webview.postMessage({
-                command: 'content',
-                text: content,
-            });
+        this.panel.webview.postMessage({
```
itowlson, line 28: Was the `canProcessMessages` check redundant?

src/components/webpanel/webpanel.ts
lstocchi: The currentPanel.update() is called inside the `setContent`. It was already there when it was `setInfo`. I just deleted the second call which is useless.

src/components/kubectl/logs.ts
lstocchi: The use case you described it's exactly what's happening here. Going from a `Follow Logs` to a `Show Logs` and reuse the webview.

I don't follow what you're saying about the two distinct method to handle the state management. There is no much to handle here. If it's a `Follow Logs` we just need to keep a reference of the child process created (so it can be killed later), if it's a `Show Logs` we need to kill the old child process if it exists and this is made by calling `setAppendContentProcess` and `deleteAppendContentProcess`.

src/components/logs/logsWebview.ts
lstocchi: The `canProcessMessages` only checked if the panel was alive (= it was opened). It was useful because the childProcess didn't get killed so without this check the process would have kept writing in the panel which was already disposed and it would be resulted in an error. Because we now kill the process when the panel is disposed we don't need to check if it's "alive" or not.

lstocchi:
@itowlson i only covered the first comment. I replied on the others with my thoughts. Let me know what you think

itowlson:
Ugh, I merged the kubectl refactoring prematurely and now this needs updating.  I'll tackle this; sorry, I should have waited and merged this first.

Issue #683, opened 2020-01-10
Follow Logs with a Match Expression receives updates that do not match


mipnw:
- In VSCode Kubernetes extension, click on a Pod and select "Follow Logs"
- A VS code tab opens with the selected log
- Set Show logs: "that don't match" 
- Set Match Expression "an expression"
- At that point note the tab is redrawn to match the "does not match an expression". So far so good.
- Do something on your pod that would log "an expression"
- Witness that new log entry appears in followed log tab.

itowlson:
Thanks for flagging this - we'll take a look.

cc @brendandburns

theoharr:
I can confirm this is still an issue in the current version

```
ms-kubernetes-tools.vscode-kubernetes-tools@1.1.1
```

lstocchi:
Closing it as https://github.com/Azure/vscode-kubernetes-tools/pull/727 was merged. Feel free to reopen it if you still face this issue once this fix will be published. Thanks

PR #727, opened 2020-04-06
Apply filters when new logs are appended by using Follow Logs

lstocchi:
This is related to #683.

No filters were applied when new logs were appended. I found a `TODO: need to apply filters here` so i guess it was already planned but it was left uncomplete

### DIFF
--- a/src/components/logs/logsWebview.ts
+++ b/src/components/logs/logsWebview.ts
@@ -61,13 +61,68 @@ export class LogsPanel extends WebPanel {
                 </div>
             </div>
             <script>
-              var lastMode = '';
-              var lastRegexp = '';
-              var renderNonce = 0;
+              let renderNonce = 0;
+              let orig = \`${this.content}\`.split('\\n');
 
-              var orig = \`${this.content}\`.split('\\n');
+              const getFilteredContent = (textToAppend) => {   
+                const regexp = document.getElementById('regexp').value;
+                const mode = document.getElementById('mode').value;                         
+                const rawContent = textToAppend ? textToAppend : orig;
+                let content;
+                if (regexp.length > 0 && mode !== 'all') {
+                    const regex = new RegExp(regexp);   
+                    switch (mode) {                        
+                        case 'include':
+                            content = rawContent.filter((line) => regex.test(line));
+                            break;
+                        case 'exclude':
+                            content = rawContent.filter((line) => !regex.test(line));
+                            break;
+                        case 'before':
+                            content = [];
+                            if (!textToAppend) { 
+                                for (const line of orig) {
+                                    if (regex.test(line)) {
+                                        break;
+                                    }
+                                    content.push(line);
+                                }
+                            }
+                            break;
+                        case 'after':
+                            if (textToAppend) {
+                                content = rawContent;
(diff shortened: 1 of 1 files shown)


src/components/logs/logsWebview.ts
```diff
@@ -79,82 +133,47 @@ export class LogsPanel extends WebPanel {
                         }
                     });
                     // TODO: need to apply filters here!
-                    elt.appendChild(document.createTextNode(message.text));
+                    const content = getBeautifiedContent(getFilteredContent(text));
```
itowlson, line 136: Is the TODO on the preceding line still correct?

src/components/logs/logsWebview.ts
```diff
@@ -61,11 +61,65 @@ export class LogsPanel extends WebPanel {
...
+                            break;
+                    }
+                } else {
+                    content = rawContent;
+                }
+                
+                return content;
+              };
+
+              const getBeautifiedContent = (contentAsArray, ix, end) => {
```
itowlson, line 113: suggest name such as `contentLines`

src/components/logs/logsWebview.ts
```diff
@@ -61,11 +61,65 @@ export class LogsPanel extends WebPanel {
...
+                if (ix && end) {
+                    contentAsArray = content.slice(ix, end);
```
itowlson, line 115: Is it really your intent to overwrite the input here?  So if the indices are passed then it ignores the passed  `contentArray` altogether?  Feels like we maybe have two distinct functions here e.g. `beautifyContentLineRange(ix, end)` (works on global `content`) and `beautifyLines(lines)` (works on the passed-in array of lines), where the first one performs the slice then calls down to the second.

src/components/logs/logsWebview.ts
```diff
@@ -61,11 +61,65 @@ export class LogsPanel extends WebPanel {
...
-              var lastRegexp = '';
-              var renderNonce = 0;
+              let renderNonce = 0;
+              let orig = \`${this.content}\`.split('\\n');
 
-              var orig = \`${this.content}\`.split('\\n');
+              const getFilteredContent = (textToAppend) => {   
+                const regexp = document.getElementById('regexp').value;
+                const mode = document.getElementById('mode').value;                         
+                const rawContent = textToAppend ? textToAppend : orig;
```
itowlson, line 70: So if `textToAppend` is passed then it _only_ applies to the new text, but otherwise it applies a new filter to the existing set of logs?  I find this combination of behaviours a bit hard to keep straight in my head... what if we had two functions `filterAll()` (operates on global `content`, for use when applying a new filter) and `filterNewLogs(logsText)` (operates on passed text, for use when appending new logs with an existing filter).  Both could call down to a common function e.g. `filter` but would differ in how they prepared the input to `filter`.

src/components/logs/logsWebview.ts
lstocchi: I seriously missed that one. It still worked bc the eval is triggered when a new regex or a new filter mode is picked and it works with the full content ... good catch :+1: fixed that. I also split the function to make the code more readable.

PR #733, opened 2020-04-15
Don't emit a camelCase warning for V1_1

itowlson:
Fixes an eslint warning on these two files.


lstocchi:
Sorry, my bad. I didn't check with lint. I usually have one single command for everything like `build: "npm run lint && npm run compile"` and i run that one. My suggestion is to run the eslint check also in travis, it enforces to merge cleaner code.

Regarding this PR, i also have a camelcase error inside `vscode-kubernetes-tools/src/api/implementation/cluster-explorer/versions.ts`

PR #738, opened 2020-04-21
Fixed issue with host item when checking minikube status

### LINKED ISSUE #732: Kubernetes: Start minikube fails
Error that pops up on VS Code:
`Command 'Kubernetes: Start minikube' resulted in an error (Cannot read property 'toLowerCase' of undefined)`

Within my `settings.json` I have the following (`<user>` is just a placeholder for this code snippet):
```json
    "vs-kubernetes": {
        "vs-kubernetes.minikube-path.windows": "C:\\Users\\<user>\\.vs-kubernetes\\tools\\minikube\\windows-amd64\\minikube.exe",
        "vs-kubernetes.helm-path.windows": "C:\\Users\\<user>\\.vs-kubernetes\\tools\\helm\\windows-amd64\\helm.exe",
        "vs-kubernetes.draft-path.windows": "C:\\Users\\<user>\\.vs-kubernetes\\tools\\draft\\windows-amd64\\draft.exe"
    },
```
But if I call `minikube start` from terminal it works, using the path found in `settings.json`:
```sh
$ C:\\Users\\<user>\\.vs-kubernetes\\tools\\minikube\\windows-amd64\\minikube.exe start
😄  minikube v1.9.2 on Microsoft Windows 10 Pro 10.0.19041 Build 19041
✨  Automatically selected the docker driver
👍  Starting control plane node m01 in cluster minikube
🚜  Pulling base image ...
💾  Downloading Kubernetes v1.18.0 preload ...
    > preloaded-images-k8s-v2-v1.18.0-docker-overlay2-amd64.tar.lz4: 542.91 MiB
🔥  Creating Kubernetes in docker container with (CPUs=2) (2 available), Memory=8100MB (12236MB available) ...
🐳  Preparing Kubernetes v1.18.0 on Docker 19.03.2 ...
    ▪ kubeadm.pod-network-cidr=10.244.0.0/16
🌟  Enabling addons: default-storageclass, storage-provisioner
🏄  Done! kubectl is now configured to use "minikube"

❗  C:\Program Files\Docker\Docker\resources\bin\kubectl.exe is v1.15.5, which may be incompatible with Kubernetes v1.18.0.
💡  You can also use 'minikube kubectl -- get pods' to invoke a matching version
```

image

lstocchi:
it fixes #732

### DIFF
--- a/src/components/clusterprovider/minikube/minikube.ts
+++ b/src/components/clusterprovider/minikube/minikube.ts
@@ -229,7 +229,11 @@ async function minikubeStatus(context: Context): Promise<MinikubeInfo> {
     const result = await context.shell.exec(`"${context.binPath}" status`);
 
     if (result && result.stderr.length === 0) {
-        const hostStatus = result.stdout.split('\n')[0].split(': ')[1].toLowerCase();
+        const hostItem = result.stdout.split('\n').find((status) => status.includes('host'));
+        if (!hostItem) {
+            throw new Error(`Failed to get host status: Unable to run Minikube`);
+        }
+        const hostStatus = hostItem.split(': ')[1].toLowerCase();
         return {
             running: 'stopped' !== hostStatus,
             message: `${result.stdout}`


src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -229,7 +229,11 @@ async function minikubeStatus(context: Context): Promise<MinikubeInfo> {
     const result = await context.shell.exec(`"${context.binPath}" status`);
 
     if (result && result.stderr.length === 0) {
-        const hostStatus = result.stdout.split('\n')[0].split(': ')[1].toLowerCase();
+        const hostItem = result.stdout.split('\n').find((status) => status.includes('host'));
+        if (!hostItem) {
+            throw new Error(`Failed to get host status: Unable to run Minikube`);
```
itowlson, line 234: I don't think we can put this message into the status function - it is used in places other than the start function.  Also not sure about throwing rather than returning it through the `MinikubeInfo.message` property.

src/components/clusterprovider/minikube/minikube.ts
```diff
@@ -229,7 +229,11 @@ async function minikubeStatus(context: Context): Promise<MinikubeInfo> {
     const result = await context.shell.exec(`"${context.binPath}" status`);
 
     if (result && result.stderr.length === 0) {
-        const hostStatus = result.stdout.split('\n')[0].split(': ')[1].toLowerCase();
+        const hostItem = result.stdout.split('\n').find((status) => status.includes('host'));
```
itowlson, line 232: When I run `minikube status` on my home machine, I get:

```
C:\Users\ivan> minikube.exe status
�🤷  There is no local cluster named "minikube"
�👉  To fix this, run: minikube start
```

This doesn't contain a `host` line, but it feels like it should return `running: false` rather than failing?

lstocchi:
> I'm not quite sure what to do about this. It feels like we should be controlling the output of `minikube status` via the `-f` flag or `-o=json` but I think we had that in the past and it didn't work reliably because Minikube changed its format string contract (dunno if we tried the `-o` flag). But there is also the case where there is no Minikube cluster _at all_ and it prints something totally different (and that is not affected by format flags or output flags so good luck parsing it as JSON _even if that's what you asked for_). And (pet peeve) the CLI loads the output with cutesy emoticons which it doesn't seem possible to turn off.
> 
> I feel like `minikube status` is so hostile that maybe we should just skip calling it from the `start` function, and instead just try running `minikube start` and interpret the exit code and output. But I'm not sure. What do you reckon?

I'm not sure. Dealing with the output of `minikube start` seems to be even more troublesome for me because of the multiple versions/vm-driver options and we cannot test everything. And even if we remove the `status` call when starting minikube we should solve this issue anyway because of the "get minishift status" command.
It seems that the previous code worked until now, nobody complained. So it makes me think that only the newest version of minikube changed the way their status is displayed. Is it too nasty to check if the host key is present (for newest versions) and if it fails to use the code as before?

itowlson:
(after email discussion with @lstocchi)  Okay, for now, we're going to merge this and get it out to people, then we can review whether we can actually support Minikube in a way that's portable across versions and architectures, or if it's just too hostile and too prone to breaking changes/cute 'improvements'.  But this will hopefully unblock the common case!

PR #746, opened 2020-04-29
Improve Portforwarding

paroga:


src/components/kubectl/port-forward.ts
```diff
@@ -105,62 +97,102 @@ export async function portForwardKubernetes(kubectl: Kubectl, explorerNode?: any
...
+    });
+    return ports;
 }
 
 /**
  * Prompts the user on what port to port-forward to, and validates numeric input.
  * @returns An array of PortMapping objects.
  */
-async function promptForPort(kubectl?: Kubectl, podName?: string, namespace?: string): Promise<PortMapping[]> {
+async function promptForPort(kubectl: Kubectl, kind: string, resourceName?: string, namespace?: string): Promise<PortMapping[]> {
```
lstocchi, line 162: Optional can be removed from `resourceName` as the calling method doesn't have this as optional and it's the only one which is using `prompForPort`

src/components/kubectl/port-forward.ts
```diff
@@ -105,62 +97,102 @@ export async function portForwardKubernetes(kubectl: Kubectl, explorerNode?: any
...
+                extractedPorts = extractDeploymentPorts(result.stdout);
             }
         } catch (err) {
             console.log(err);
         }
     }
 
+    let defaultValue: string | undefined = undefined;
+    if (extractedPorts.length > 0) {
+        const portPairs = extractedPorts.map(({name, port}) => `${port}:${name}`);
```
lstocchi, line 185: `name` can be undefined so i'd suggest to check its value or we can end up showing `8080:undefined`

lstocchi:
@itowlson can you give it a look pls? Maybe i missed something. 

I also noticed that when you write the port mapping in the inputbox you can write something like `5000 4000 3000:8000` and it does 1+ port-forwarding (in this case 3). Is this behavior correct or it's a bug? I think it just check if it contains a `:` and numbers

paroga:
> Is this behavior correct or it's a bug?

The current implementation should match the behavior of `kubectl port-forward`. Did I miss something or what did you expect?

It supports three variants:
* `localPort:remotePort` will map `localPort` to `remotePort`
* `:remotePort` will map a free local port to `remotePort`
* `remotePort` will map the local port with the same port number as `remotePort` to `remotePort`

lstocchi:
@paroga I was referring to the fact of writing multiple ports in the inputBox and if it was documented but i just read that the inputbox description clearly states `Separate multiple port mappings with spaces` so no issue. Just my tired eyes that didn't catch that :+1:

paroga:
any updates on this PR?

itowlson:
@paroga Sorry for the delay, I have been knocked out with other responsibilities.  Doing a big catch-up this morning and looking at it now.

src/components/kubectl/port-forward.ts
```diff
@@ -176,31 +208,33 @@ async function promptForPort(kubectl?: Kubectl, podName?: string, namespace?: st
...
  * Validates the user supplied port mapping(s)
  * @param portMapping The portMapping string captured from an input field
+ * @param validPorts List of valid named ports
  * @returns A ValidationResult object describing the first error found.
  */
-function validatePortMapping(portMapping: string): ValidationResult | undefined {
+function validatePortMapping(portMapping: string, validPorts: ExtractedPort[] = []): ValidationResult | undefined {
     const portPairs = portMapping.split(' ');
-    const validationResults: ValidationResult[] = portPairs.map(validatePortPair);
+    const validationResults: ValidationResult[] = portPairs.map(validatePortPair.bind(null, validPorts));
```
itowlson, line 222: Rather than using `.bind` could we use a lambda here?
```suggestion
    const validationResults: ValidationResult[] = portPairs.map((pair) => validatePortPair(validPorts, pair));
```

It may be personal taste but I find this easier to read than `bind` with its irksome pseudo-`this` argument... *grin*

itowlson: This might allow you to remove the `: ValidationResult[]` explicit type annotation too.

src/components/kubectl/port-forward.ts
```diff
@@ -234,26 +266,45 @@ function validatePortPair(portPair: string): ValidationResult {
...
 }
 
+/**
+ * Validates if the port is a named port or withing the valid range
+ * @param validPorts List of valid named ports
+ * @param port The port to validate
+ * @param nullable Can the port be empty or zero
+ * @returns Boolean identifying if the port is valid
+ */
+function isPortValid(validPorts: ExtractedPort[], port: string, nullable?: boolean): boolean {
```
itowlson, line 276: I would like the `nullable` argument to be mandatory because it's not obvious which way the default would be.  Even better might be to make it an enum e.g. `PortSpecifier.{ Required | AllowEmpty }` as this will be more self-documenting at the call site.

src/components/kubectl/port-forward.ts
```diff
@@ -234,26 +266,45 @@ function validatePortPair(portPair: string): ValidationResult {
...
  * Builds and returns multiple PortMapping objects
  * @param portString A validated, user provided string containing the port mappings
+ * @param validPorts List of valid named ports
  * @returns An array containing the requested PortMappings
  */
-export function buildPortMapping(portString: string): PortMapping[] {
+export function buildPortMapping(portString: string, validPorts: ExtractedPort[] = []): PortMapping[] {
     const portPairs = portString.split(' ');
-    return portPairs.map(buildPortPair);
+    return portPairs.map(buildPortPair.bind(null, validPorts));
```
itowlson, line 294: Again I prefer lambdas to bind but it's a personal taste nit, not a big deal.

src/components/kubectl/port-forward.ts
```diff
@@ -264,11 +315,38 @@ function buildPortPair(portPair: string): PortMapping {
...
+}
+
+/**
+ * Builds a single numberic port for a PortMapping object from the captured user input
+ * @param validPorts List of valid named ports
+ * @param portString The port provided by the user
+ * @returns numberic port number
+ */
+function buildPort(validPorts: ExtractedPort[], port: string): number {
+    const found = validPorts.filter(({name}) => name === port);
```
itowlson, line 343: Consider using `find`:

```suggestion
    const validPort = validPorts.find(({name}) => name === port);
```

(and then `if (validPort)`)

src/components/kubectl/port-forward.ts
```diff
@@ -234,26 +266,45 @@ function validatePortPair(portPair: string): ValidationResult {
...
+}
+
 /**
  * Builds and returns multiple PortMapping objects
  * @param portString A validated, user provided string containing the port mappings
+ * @param validPorts List of valid named ports
  * @returns An array containing the requested PortMappings
  */
-export function buildPortMapping(portString: string): PortMapping[] {
+export function buildPortMapping(portString: string, validPorts: ExtractedPort[] = []): PortMapping[] {
```
itowlson, line 292: Is `validPorts` the right name through all this?  It feels like other ports could be perfectly 'valid' but I think these are specifically ports that are given names in the spec.  Is that right?  If so consider renaming to `namedPorts` or `extractedPorts` or something.

On the other hand if these really are the only valid ports and anything else will be rejected then stick with the current name.

paroga:
@itowlson ping

itowlson (approved):
Looks good - thanks and apologies for the delay!

Issue #748, opened 2020-05-02
Feature Request: Helm Context Menus


shreyu86:
It would be great if there are context menus on Helm Resources. 
Currently `Helm Releases` lists helm releases it would be really useful to have context menus like `Status`, `History`, `Rollback` `Uninstall` in the context menu of a helm release. 

Happy to submit a PR for this if this is something that would be considered helpful. 
If there is interest in adding the context menus, what would be a good place to start?

itowlson:
I would love to see a PR for this - thanks!

In terms of where to start, not sure if you are asking about which command(s) to start with, or the technical aspects.  You've already done a lot of work on this extension already, so I'm guessing the former, because you probably know your way around the technical side!  (But happy to offer guidance if you do want it.)

I agree it would be best to PR just one or two initially, and then do the rest when we had worked through any design considerations.  My suggestions would be Rollback and Uninstall because those probably require less output than Status or History so you don't need to get tangled up in webviews or stuff...?

shreyu86:
Thanks for the pointers @itowlson I will start with uninstall as the first context menu to add which gets me up to speed on the extension code base again once I gather feedback on the uninstall release function I can add rollback (I have a WIP for that struggling with `showQuickPick`)

shreyu86:
opened https://github.com/Azure/vscode-kubernetes-tools/pull/749 as the first PR for the Helm context menus

shreyu86:
For Helm release rollback I experimented with two implementation.
1.[Implementation with `showQuickPick` ](https://github.com/shreyu86/vscode-kubernetes-tools/tree/helm-rollback)
2. [Implementation with Helm history as a new node (child of Helm release node)](https://github.com/shreyu86/vscode-kubernetes-tools/tree/exp2)

Here is how the implementation1 look like: 

![impl1-lr](https://user-images.githubusercontent.com/311217/81524722-171c3480-9307-11ea-9636-06bd86ce0d13.gif)

and here is implementation 2 looks like:
![impl2-lr](https://user-images.githubusercontent.com/311217/81524742-2c915e80-9307-11ea-8783-f4b08b76797e.gif)

wanted to get early feedback before I open an MR so that I can polish the implementation of choice.

lstocchi:
Difficult choice as they both make sense to me. 
[...]

PR #749, opened 2020-05-04
Helm Uninstall Context Menu

shreyu86:
* adds `uninstall` context menu to Helm release explorer. 
* have included an confirmation dialog to avoid any accidents. 
* added option to copy release name as only having `Uninstall` in the context menu could lead to accidentally clicking on that option only.
part of the feature request #748


package.json
```diff
@@ -585,6 +586,11 @@
                     "group": "0@2",
                     "when": "view == extension.vsKubernetesHelmRepoExplorer && viewItem =~ /vsKubernetes\\.((chart)|(chartversion))/i"
                 },
+                {
+                    "command": "extension.helmUninstall",
+                    "group": "3",
```
shreyu86, line 591: was unsure here so have added a new group, happy to follow the conventions for groups if new group is not appropriate.

src/helm.exec.ts
```diff
@@ -274,6 +274,31 @@ export function helmGet(resourceNode?: ClusterExplorerNode) {
...
+export function helmUninstall(resourceNode?: ClusterExplorerNode) {
+    if (!resourceNode) {
+        return;
+    }
+    if (resourceNode.nodeType !== NODE_TYPES.helm.release) {
+        return;
+    }
+    const releaseName = resourceNode.releaseName;
+    logger.log("⎈⎈⎈ Uninstalling " + releaseName);
+    vscode.window.showWarningMessage(`This will uninstall ${releaseName}, this cannot be undone.`, 'Uninstall').then((opt) => {
```
lstocchi, line 286: I would change the message... maybe `You are about to uninstall ${releaseName}. This action cannot be undone.`

shreyu86:
@lstocchi thank you for your comments. I have incorporated your feedback. Also noticed I had a couple of linting issues so fixed those too.

src/helm.exec.ts
shreyu86: Thank you, improved the message.

lstocchi:
Thanks @shreyu86. It looks good to me. I saw that you also want to add other features. Do you want to keep this PR opened and add more stuff here or this can be merged?

shreyu86:
@lstocchi I am still experimenting with rollback. I think if it's possible this would be phase 1 and can be merged. I'll open other MR soon.

lstocchi:
Ok, merging this then. Thanks again @shreyu86

shreyu86:
Awesome! Thank you! Looking forward to additional context menus for helm 😊

PR #751, opened 2020-05-09
Fix attach to node.js

aelij:
On my machine I kept getting the error: "Unable to set Node.js in the container to inspect mode".
After debugging the extensions, I saw the error from `kubectl` was an unterminated string after `-c`. Changing to double quotes fixed it for me.


lstocchi:
@aelij thank you for this patch. One thing before merging this. Can you please give me more infos about your environment? E.g your OS? are you using minikube?

aelij:
Windows, connecting to AKS.

PR #753, opened 2020-05-13
Helm Context Menu - Rollback

shreyu86:
Implementation for Helm release rollback to a previous version.

Part of the enhancements as per #748 

* Added new nodetype for the helm releases
* On each of the releases have the option to rollback
* I tried to figure out if there is option of not having the rollback context menu on the current `deployed` release but my research came up short, so I have added an informational message to handle this case.
* If this PR is okay I will need some help to understand how to display this release information using `helmget` and to show the release specific values. Similar function to [this](https://github.com/Azure/vscode-kubernetes-tools/blob/c34d7d11aaccaad55fcffe126caa14366d053d90/src/components/clusterexplorer/node.helmrelease.ts#L20)  

This is one of the two implementations I was thinking of https://github.com/Azure/vscode-kubernetes-tools/issues/748#issuecomment-626466066


src/helm.exec.ts
```diff
@@ -263,7 +275,7 @@ export function helmGet(resourceNode?: ClusterExplorerNode) {
     if (!resourceNode) {
         return;
     }
-    if (resourceNode.nodeType !== NODE_TYPES.helm.release) {
+    if ((resourceNode.nodeType !== NODE_TYPES.helm.release || NODE_TYPES.helm.history)) {
```
lstocchi, line 278: There is an issue here. It's interpreted like  (resourceNode.nodeType !== NODE_TYPES.helm.release || true) as NODE_TYPES.helm.history has a value. The helm history treeItem command is never executed bc it gets in this if.

src/components/clusterexplorer/node.helmrelease.ts
```diff
@@ -1,21 +1,55 @@
...
 import { Host } from '../../host';
-import { ClusterExplorerNode, ClusterExplorerNodeImpl, ClusterExplorerHelmReleaseNode } from './node';
+import { Kubectl } from '../../kubectl';
 import { NODE_TYPES } from './explorer';
+import { ClusterExplorerHelmHistoryNode, ClusterExplorerHelmReleaseNode, ClusterExplorerNode, ClusterExplorerNodeImpl } from './node';
+import { MessageNode } from './node.message';
+import moment = require('moment');
+
+export class HelmHistoryNode extends ClusterExplorerNodeImpl implements ClusterExplorerHelmHistoryNode {
+    constructor(readonly releaseName: string, readonly revision: number, readonly updated:  string, readonly status: string, readonly description: string) {
```
lstocchi, line 13: Why do you need the description here? It looks like it's never used or i'm missing something?

src/helm.exec.ts
```diff
@@ -301,6 +313,47 @@ export function helmUninstall(resourceNode?: ClusterExplorerNode) {
...
+    }
+    return { succeeded: true, result: releases.reverse() };
+}
+
+export async function helmRollback(resourceNode?: HelmHistoryNode) {
+    if (!resourceNode) {
+        return;
+    }
+    if (resourceNode.status === "deployed") {
+        vscode.window.showInformationMessage('this is the currently deployed release');
```
lstocchi, line 336: First letter uppercase `This`

shreyu86:
Thanks for the comments @lstocchi and no worries about the delay. I'll address the comments today.

src/helm.exec.ts
shreyu86: Fixed

src/components/clusterexplorer/node.helmrelease.ts
shreyu86: Removed this, was thinking of using this but deferred to not use this

src/helm.exec.ts
shreyu86: Fixed this, thanks for catching this

src/helm.exec.ts
```diff
@@ -275,7 +275,7 @@ export function helmGet(resourceNode?: ClusterExplorerNode) {
     if (!resourceNode) {
         return;
     }
-    if ((resourceNode.nodeType !== NODE_TYPES.helm.release || NODE_TYPES.helm.history)) {
+    if (resourceNode.nodeType !== (NODE_TYPES.helm.history ||NODE_TYPES.helm.release)) {
```
lstocchi, line 278: We still have a problem here because it will first check the result inside the brackets `(NODE_TYPES.helm.history || NODE_TYPES.helm.release)` and it will return a value based on the OR expression. 

The OR will return the first value of the expression if its value is true and because NODE_TYPES.helm.history will be always converted to true because it is always defined, it's like you're just doing `resourceNode.nodeType !== NODE_TYPES.helm.history`. If you try it you will see that it now works for heml.history nodes but it fails for heml.release nodes.

I'd suggest to simply convert it to `resourceNode.nodeType !== NODE_TYPES.helm.history && resourceNode.nodeType !== NODE_TYPES.helm.release`

shreyu86: Sorry my bad I missed this. Fixing this condition

shreyu86:
@lstocchi thank you for your time and help. Agreed on the documentation part once we this merged can add documentation for rollback and uninstall features. Thanks again really appreciate your help.

lstocchi:
Thank you @shreyu86 !!!

shreyu86:
Thanks for merging this @lstocchi

PR #736, opened 2020-04-19
#Issue 724: breakpoints in vscode not working as expected when debugging in docker container 

mattjanson:
Added the possibility to define a sourceFileMap
entry to the dotNetDebugProvider in order to fix an issue
with non working breakpoints for the Kubernetes: Debug (Attach)
operation

PR #754, opened 2020-05-14
Fix for Pullrequest #736: replacing all occurences of backslash with regex

ltsbkuster:
For paths nested deeper in filesystem it should use the regex-version of `replace` to replace all occurences of the backslash.

I also updated the sample docs about `vs-kubernetes.dotnet-source-file-map` because it was pretty unclear to me that a string to the path is expected.

@mattjanson do you agree with the description?


mattjanson:
@ltsbkuster Yes, I totally agree. I was on the Track the replace function does replace all occurances :(

PR #756, opened 2020-05-17
Adding support for Debug(Launch) to NodeJS project

lstocchi:
This is related to https://github.com/Azure/vscode-kubernetes-tools/issues/183 , https://github.com/Azure/vscode-kubernetes-tools/issues/155 and https://github.com/Azure/vscode-kubernetes-tools/issues/154 

@itowlson i tested it on minikube and it works fine. I guess we need to add a new section to the documentation with a gif/small video to show how easy it is to make it work .... but the code is ready to be reviewed if you want to give it a look.


itowlson:
Sorry for the delay @lstocchi - looking now.

src/debug/nodejsDebugProvider.ts
```diff
@@ -103,4 +118,15 @@ export class NodejsDebugProvider implements IDebugProvider {
...
         return true;
     }
+
+    public async getDebugArgs(): Promise<string | undefined> {
+        const debugCommand = await vscode.window.showInputBox({
+            prompt: 'Debug command for your container:',
+            placeHolder: 'Example: node --inspect app.js'
+        });
+        if (!debugCommand) {
+            return undefined;
```
itowlson, line 128: Feels like there might be a quirk here: if the user cancels out of the input box (by hitting Esc), it looks like the deployment operation will go ahead but with no additional arguments.  This may not be what the user expects from cancellation - they may expect to cancel the entire operation, not just the command gathering part.

src/debug/nodejsDebugProvider.ts
```diff
@@ -103,4 +118,15 @@ export class NodejsDebugProvider implements IDebugProvider {
     public isPortRequired(): boolean {
         return true;
     }
+
+    public async getDebugArgs(): Promise<string | undefined> {
+        const debugCommand = await vscode.window.showInputBox({
+            prompt: 'Debug command for your container:',
```
itowlson, line 124: Not quite sure what "Debug command for your container" means... but not sure what a better prompt would be either!

src/debug/nodejsDebugProvider.ts
lstocchi: Good catch :+1: Not sure about which naming to use here. I used { succeeded, command }, even if succeeded doesn't satisfy me completely. If you have a better idea, let me know

src/debug/nodejsDebugProvider.ts
lstocchi: I used this as it was already used in the other debug command and because i didn't find anything better .... but i agree with your concern. I changed it. Let me know if you like it more.

lstocchi:
@itowlson I forgot to add a bit of documentation for debug launch. Please give a look at it too. Thanks :)

src/debug/nodejsDebugProvider.ts
itowlson: In some of my projects I have a `Cancellable<T>` type - e.g. https://github.com/deislabs/porter-vscode/blob/master/src/utils/cancellable.ts.  Would that be appropriate here perhaps?

Issue #770, opened 2020-06-17
Kubernetes command failed


gildas:
I use kubectl on my Linux Mint box via snap.
I run microk8s (still via snap) locally as well as using remote k8s clusters (Azure, Docker for Mac remotely, etc).
I run helm also via snap.
So, my snap list looks like (snippet):
```console
Name                   Version                     Rev   Tracking       Publisher              Notes
code                   a5d1cc28                    34    latest/stable  vscode✓                classic
core                   16-2.45                     9289  latest/stable  canonical✓             core
docker                 19.03.11                    471   latest/stable  canonical✓             -
helm                   3.2.3                       244   latest/stable  snapcrafters           classic
kubectl                1.18.3                      1534  latest/stable  canonical✓             classic
microk8s               v1.18.3                     1443  latest/stable  canonical✓             classic
```
Note that I have a `~/.kube/config` with several clusters, all work fine from the shell.

I decided to give `vscode-kubernetes-tools` a spin.
After installation when you click on the extension logo in the Activity Bar, the list of clusters, helm repos, and clouds stay empty (I saw the little progress bar running for each for a few seconds); a toaster popped up saying only: `Kubectl command failed:` without any explanation whatsover.

I looked at the `Output` view in the `Log (Extension Host)` and nothing either.

Not sure where to look for information and reasons why the extension believes it cannot run `kubectl`

lstocchi:
Thanks for opening this issue @gildas. We received other issues similar to this and it seems that VSCode installed with snap comes with some permissions problem. I wasn't able to replicate it before but i never tried with Linux Mint. I'll let you know if i find a solution to this.

lstocchi:
@gildas I was able to replicate it and i think i have a fix (#771) . Would you please try it and see if it works for you?

You can install the extension by using a vsix file.
I created it (download link: https://github.com/lstocchi/vscode-kubernetes-tools/raw/master/vscode-kubernetes-tools-1.2.1.vsix ) or you can build it by yourself (you can use this branch https://github.com/lstocchi/vscode-kubernetes-tools/tree/i770 ).

Once installed you should keep seeing the error as before `Kubectl command failed:`. Now if you go in the settings page and set "vs-kubernetes.enable-snap-flag" to true the extension should start working normally (CMD/CTRL + SHIFT + P -> search for `Open Setting` -> click on `Preferences: Open Settings (JSON)` -> add `"vs-kubernetes.enable-snap-flag": true` and save).

[...]

Issue #522, opened 2019-04-15
Kubernetes extension does not work with snap version of VS Code


dholbach:
I have the latest version of VS Code install through `snap install code --classic` on Ubuntu.

When starting the Kubernetes extension (latest, installed through VSIX), I see the following error messages in `/var/log/syslog`:

```
Apr 15 11:59:50 reef kernel: [13150.164921] audit: type=1400 audit(1555322390.646:498): apparmor="DENIED" operation="file_inherit" profile="/snap/core/6673/usr/lib/snapd/snap-confine" pid=18791 comm="snap-confine" family="unix" sock_type="stream" protocol=0 requested_mask="send receive" denied_mask="send receive" addr=none peer_addr=none
```

In another round of testing, I saw:

```
[69083.590392] audit: type=1400 audit(1554802886.556:723): apparmor="DENIED" operation="ptrace" profile="/usr/sbin/libvirtd" pid=1305 comm="libvirtd" requested_mask="read" denied_mask="read" peer="snap.vscode.vscode"
```

I turned there for help as I couldn't figure out why the extension simply didn't work for me with a local `minikube` cluster up and running.

It appears `snap` confinement blocks access to resources the Kubernetes extension need to figure out what's happening.

itowlson:
Hmm, okay.  I'm not familiar with `snap` but will take a look.  Are you able to work around this by installing VS Code via another mechanism?  Thanks for letting us know!

dholbach:
The [.deb version](https://code.visualstudio.com/docs/setup/linux#_debian-and-ubuntu-based-distributions) works fine.

iliatages:
+1

4ndt3s:
Same problem here.

lstocchi:
Hello everyone,
I was able to replicate it and i think i have a fix (#771) . Would you please try it and see if it works for you?

You can install the extension by using the vsix file.
I created it (download link: https://github.com/lstocchi/vscode-kubernetes-tools/raw/master/vscode-kubernetes-tools-1.2.1.vsix ) or you can build it by yourself (you can use this branch https://github.com/lstocchi/vscode-kubernetes-tools/tree/i770 ).

[...]

PR #771, opened 2020-06-18
Added a hack to make kubectl works if installed with snap

lstocchi:
Finally i was able to replicate the error users have by installing stuff with snap. #770 #522 
Did i find a fix? Apparently yes. Did i find out the reason why it doesn't work? No, absolutely not lol. 

This is what i got if i try to execute the command without the hack
error snap

In Unix systems this solution should always work (i guess in Windows it fails as there is no `cat` cmd) but i added an extra setting so snap users can enable this because i am not sure if i'm forgetting something. This way the extension will keep working as before for others.

### DIFF
--- a/package.json
+++ b/package.json
@@ -266,7 +266,7 @@
                         },
                         "vs-kubernetes.enable-snap-flag": {
                             "type": "boolean",
-                            "description": "If true will enable the flag to run kubectl commands using kubectl installed by snap."
+                            "description": "Enables compatibility with instances of VS Code that were installed using snap."
                         }
                     },
                     "default": {
--- a/package.json
+++ b/package.json
@@ -263,6 +263,10 @@
                         "vs-kubernetes.resources-to-watch": {
                             "type": "array",
                             "description": "List of resources to be watched."
+                        },
+                        "vs-kubernetes.enable-snap-flag": {
+                            "type": "boolean",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -263,6 +263,10 @@
                         "vs-kubernetes.resources-to-watch": {
                             "type": "array",
                             "description": "List of resources to be watched."
+                        },
+                        "vs-kubernetes.enable-snap-flag": {
+                            "type": "boolean",
+                            "description": "If true will enable the flag to run kubectl commands using kubectl installed by snap."
```
itowlson, line 269:
```suggestion
                            "description": "Enables compatibility with instances of VS Code that were installed using snap."
```

itowlson: Might be worth adding something to the read-me, or starting a new Troubleshooting or FAQ page that we can link from the read-me.

lstocchi:
@itowlson Updated the readme. Can you give it a look? Feel free to edit it.

I tried to reproduce the error in #522 but i cannot replicate it and i don't have any idea how to fix that (i will wait someone else to give me some hint) .... but I think we can merge this PR as it cover some use cases (#770). WDYT?

itowlson:
Agree, let's ship it to help out the folks for whom it works, but keep tracking with users who are still experiencing problems.

Issue #765, opened 2020-06-05
Add option to limit log output to previous x mins


adamrushuk:
Follow Logs and Show logs are great options, but sometimes there are too many logs, so it would be useful to limit the output to previous 5 mins for example.

Basically passing `--since=5m` to kubectl, eg: `kubectl logs -f deployment/nginx --all-containers=true --since=5m`

itowlson:
@adamrushuk Some discussion of possible UI going on over at @Devjeel's PR #773 - we'd love your input if you haven't seen it.

adamrushuk:
Great to see action on this already; thanks all :)

lstocchi:
Closing it as new log viewer has been merged and released. Please open new issues if you face any problem or have some ideas how to enhance it. Thank you!!

PR #773, opened 2020-06-22
Add limit logs support 

Devjeel:
- [✔] Add limit logs support

- [✔] Update README

- [  ] (Help needed) --since value provided explicitly. eg `--since=5m` . Any need to make it user inputtable ? 

Fix to : #765

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ If you want to skip TLS verification for a particular cluster, you can edit your
    * `Kubernetes: Get` - Get the status for a specific resource.
    * `Kubernetes: Follow Logs` - Follow logs for a pod in an output window.
    * `Kubernetes: Show Logs` - Show logs for a pod in an output window.
+   * `Kubernetes: Limit Logs` - Limit logs to only return last X minute or seconds logs in an output window. (Defult to 5 min.)
    * `Kubernetes: Follow Events` - Follow events on a selected namespace.
    * `Kubernetes: Show Events` - Show events on a selected namespace.
    * `Kubernetes: Watch` - Watch a specific resource or all resources of that object type, and update the cluster explorer as they change
--- a/src/components/kubectl/logs.ts
+++ b/src/components/kubectl/logs.ts
@@ -13,7 +13,8 @@ import { ExecResult } from '../../binutilplusplus';
 
 export enum LogsDisplayMode {
     Show,
-    Follow
+    Follow,
+    Limit
 }
 
 /**
@@ -97,6 +98,10 @@ async function getLogsForContainer(
 
     if (displayMode === LogsDisplayMode.Follow) {
         args.push('-f');
+    }
+
+    if (displayMode === LogsDisplayMode.Limit) {
+        args.push('--since=5m');
     }
 
     const cmd = args.join(' ');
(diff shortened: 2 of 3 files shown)


Devjeel:
@lstocchi Yes, you are right. hard coded values are not good. 

Please share ideas on how to allow user to input value. 🚀
@itowlson

itowlson:
I did wonder if we could add a drop-down to the logs webview to specify the 'since' interval, but I don't think that works well unless we default to limiting - otherwise the webview displays everything first, and then you get to say "uh no only the last 30 seconds" which isn't a great user experience!

But a possible development of that would be to do something like this:

* Get rid of separate Show, Follow and Limit Logs commands; just have a Logs command
* When the webview starts up, it does _not_ display logs immediately.  Instead it displays only the toolbar, with options like:
  * Show or Follow
  * Time period to show (e.g. 1/2/5/10/20/30 mins/1/2/6/24/48 hours/All)
  * Filter options (as present)
* The toolbar also has a Go button
* And we fetch the logs when the user presses the Go button

Thoughts?

One issue with this is for folks who prefer their logs in the terminal.  I guess we could have another option in the toolbar to send the logs to the terminal (which would disable the Filter options... unless anyone fancies constructing the right grep incantation and executing a piped command), but not sure that's the right experience.

If we don't want something that radical then yes, a VS Code quick pick is the way to go to select the time interval.  But I'm not sure we want to force users to go through that quick pick every time they want to view logs...

itowlson:
Thinking about this I am tempted by the Radical Webview option because it gives us somewhere to put future knobs too, rather than requiring more commands and more quick picks for more and more log options.

Also, the "which container do you want to see logs for" quick pick (for multi-container pods) could go in the toolbar too.

adamrushuk:
> When the webview starts up, it does _not_ display logs immediately. Instead it displays only the toolbar, with options like:
> * Show or Follow
> * Time period to show (e.g. 1/2/5/10/20/30 mins/1/2/6/24/48 hours/All)
> * Filter options (as present)

This sounds like the best way forward to me 👍 

> I guess we could have another option in the toolbar to send the logs to the terminal

I also really like this idea.

However it's implemented, having a way to limit log output will be a huge win for this extension.

lstocchi:
> Thinking about this I am tempted by the Radical Webview option because it gives us somewhere to put future knobs too, rather than requiring more commands and more quick picks for more and more log options.
> 
> Also, the "which container do you want to see logs for" quick pick (for multi-container pods) could go in the toolbar too.

I agree with this. It could be quite handy to have a unique webview to manage/add other features in future. 

> If we don't want something that radical then yes, a VS Code quick pick is the way to go to select the time interval. But I'm not sure we want to force users to go through that quick pick every time they want to view logs...

It's a radical solution but if we don't do it now I have the feeling we will end up with this discussion again in the future. So, better to tackle it now. 
I also don't like much showing a quick pick before to show logs but this was the only idea i had. If we have a better solution, we should go for that.

> But a possible development of that would be to do something like this:
> 
> * Get rid of separate Show, Follow and Limit Logs commands; just have a Logs command
> * When the webview starts up, it does _not_ display logs immediately.  Instead it displays only the toolbar, with options like:
>   
>   * Show or Follow
>   * Time period to show (e.g. 1/2/5/10/20/30 mins/1/2/6/24/48 hours/All)
>   * Filter options (as present)
> * The toolbar also has a Go button
> * And we fetch the logs when the user presses the Go button
> 
> Thoughts?

+1

itowlson:
@Devjeel Sorry, I had not thought about these complexities when I tagged it as a good first issue!  I hope you're still keen to work on it?!

Devjeel:
@itowlson this sounds great ;) . Let me give it a try. If you can suggest how to implement webview/toolbar type for logs option ?

itowlson:
@Devjeel The toolbar is already in there with the filter functionality and a "scroll to bottom" button.  You should be able to add new controls to that existing toolbar.  The tricky bit will, I think, be deferring the fetch of the logs until the user hits the Go button.

itowlson:
@Devjeel Forgot to say, thanks for doing this!

lstocchi:
Any news on this PR @Devjeel ? Do you want us to handle it?

Devjeel:
> 
> 
> Any news on this PR @Devjeel ? Do you want us to handle it?

Hi @lstocchi , got caught in my own stuff. Please assign it to someone else. Thanks.

lstocchi:
Ok thank you @Devjeel !!

Issue #772, opened 2020-06-22
Provide extensions recommendation based on installed CRD 


evidolob:
`vscode-kubernetes-tools` can provide recommendation to install extensions base on current k8s  cluster CRS's.
This mostly the same as [Workspace recommended extensions](https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions), but instead of letting user configure that we can provide recommendation list.

Requirements:

- Flow and behavior is mostly the same as `Workspace recommended extensions`, first we show dialog that we have recommendations for current cluster, if `Show` button clicked, we show tree view with list of recommended extensions.
- Recommendations can be turn off in preferences
- Dialog shows only once when extension connects to new cluster
- Dialog not appears when there are all recommended extensions installed

We can have that recommendation list in JSON file, format example:

```json
{
    "crdApiToExtensions": {
        "CRD.API.value": [
            "some.vscode-extenion-id"
        ],
    },
    "extensions": [
        {
            "id": "some.vscode-extenion-id",
            "label": "Cool  Extensions",
            "description": "Description to the Cool Extensions",
            "icon": "https://path/to/icon.png"
        }
    ]
}
```

I have made a prototype of this:
![ezgif com-video-to-gif (15)](https://user-images.githubusercontent.com/929743/83023536-b3e40f00-a035-11ea-88b8-085d4903f24c.gif)
so I can provide a PR with this feature.

lstocchi:
What do you think @itowlson ?

itowlson:
This is a really interesting idea.  We'd need to maintain a (user-curated) mapping of CRDs to extension recommendations, and figure out how to make it Not Annoying.

I'm actually wondering if this is something that could be done using the API, external to the core extension.  That might make it easier to maintain and would mean it could ship on its own cadence as users contributed new recommendations.  But it would also make it less easy to discover.

cc @squillace for Giant Marketing Brain

squillace:
OK, originally I read this and started thinking, which required a bunch of drinking first, so now I'm back. 

I think, in so far as CRDs are "cluster role services", it makes some sense to enabling this kind of functionality. The user story is that they:

- connect to a cluster
- the extension sees "known" CRDs and prompts for related lookups
- user has configurations to control this

Does that sound about right?

evidolob:
@squillace Yes it does.

squillace:
[...]

PR #780, opened 2020-07-01
Extensions recommendation based on installed CRD

evidolob:
This is draft implementation.

Todo:

- [ ] Detect cluster change to update recommendation list
- [ ] Add preference settings to disable recommendations

Current demo: 
ezgif com-video-to-gif (23)

Fix: #772

### DIFF
--- a/src/components/extexplorer/explorer.ts
+++ b/src/components/extexplorer/explorer.ts
@@ -1,0 +1,156 @@
+
+import * as vscode from 'vscode';
+import * as fs from 'fs';
+import * as path from 'path';
+import { Kubectl } from '../../kubectl';
+
+const RECOMMENDED_EXT_CONTEXT = 'kubernetes:recommendedExtensions';
+
+enum VSCodeCommands {
+    SetContext = 'setContext',
+    Open = 'vscode.open'
+}
+
+interface RecommendedExtension {
+    id: string;
+    label: string;
+    description: string;
+    icon: string;
+}
+interface Recommendations {
+    crdApiToExtensions: { [crdApi: string]: string[] };
+    extensions: RecommendedExtension[];
+}
+
+class ExtensionTreeItem extends vscode.TreeItem {
+
+    constructor(private ext: RecommendedExtension) {
+        super(ext.label, vscode.TreeItemCollapsibleState.None);
+    }
+
+    get description(): string {
+        return this.ext.description;
+    }
+
+    get iconPath(): vscode.Uri {
+        return vscode.Uri.parse(this.ext.icon);
+    }
+
+    get command(): vscode.Command {
+        return { command: VSCodeCommands.Open, arguments: [vscode.Uri.parse(`vscode:extension/${this.ext.id}`)] } as vscode.Command;
+    }
+
+}
+
+class RecommendedExtensionsView implements vscode.TreeDataProvider<RecommendedExtension>{
+
+    private extensions: RecommendedExtension[] | undefined;
+    private treeView: vscode.TreeView<RecommendedExtension>;
+    private onDidChangeTreeDataEmitter = new vscode.EventEmitter<RecommendedExtension | undefined>();
+    readonly onDidChangeTreeData: vscode.Event<RecommendedExtension | undefined> = this.onDidChangeTreeDataEmitter.event;
+
+    constructor() {
+        this.treeView = vscode.window.createTreeView('kubernetes.RecommendedExtensions', { treeDataProvider: this });
+    }
+
+    dispose(): void {
+        this.treeView.dispose();
+    }
+
+    getTreeItem(element: RecommendedExtension): vscode.TreeItem | Thenable<vscode.TreeItem> {
+        return new ExtensionTreeItem(element);
+    }
+    getChildren(element?: RecommendedExtension): vscode.ProviderResult<RecommendedExtension[]> {
+        if (!element && this.extensions) {
+            return this.extensions;
+        }
+        return undefined;
+    }
+    getParent?(): vscode.ProviderResult<RecommendedExtension> {
+        return undefined; // we have plain list, so no parents
+    }
+
+    setExtensions(extensions: RecommendedExtension[]): void {
+        this.extensions = extensions;
+    }
+
+}
+
+async function recommendExtensionsToUser(extensions: RecommendedExtension[], subscription: vscode.Disposable[]): Promise<void> {
+    if (extensions.length > 0) { // do not prompt if we cannot recommend anything
+        const treeView = new RecommendedExtensionsView();
+        subscription.push(treeView);
+        treeView.setExtensions(extensions);
+        vscode.commands.executeCommand(VSCodeCommands.SetContext, RECOMMENDED_EXT_CONTEXT, true);
+    }
+
+}
+
+async function getCRDApiGroups(kubectl: Kubectl): Promise<string[]> {
+    const result = new Set<string>();
+    const invokeResult = await kubectl.invokeCommand('get crd -o json');
+    if (invokeResult.resultKind === 'exec-succeeded') {
+        try {
+            const res = JSON.parse(invokeResult.stdout);
+            res.items.forEach((i: any) => result.add(i.spec.group));
+        } catch (err) {
+            // ignore
+        }
+    }
+    return Array.from(result);
+}
+
+async function readRecommendation(extPath: string): Promise<Recommendations> {
+    return new Promise((resolve, reject) => {
+        fs.readFile(path.join(extPath, 'recommendation', 'recommended-extensions.json'), (err, content) => {
+            if (err) {
+                reject(err);
+            }
+            const fileContent = content.toString('utf8');
+            try {
+                resolve(JSON.parse(fileContent));
+            } catch (error) {
+                reject(error);
+            }
+        });
+    });
+}
+
+export async function recommendExtensions(kubectl: Kubectl, context: vscode.ExtensionContext): Promise<void> {
+    vscode.commands.executeCommand(VSCodeCommands.SetContext, RECOMMENDED_EXT_CONTEXT, false);
+    const crdApiGroups = await getCRDApiGroups(kubectl);
+
+    const extensionToPromote = new Set<string>();
+    const recommendation = await readRecommendation(context.extensionPath);
+    const recommendationsMap = recommendation.crdApiToExtensions;
+    // collect extension id to promote based on installed operators
+    for (const crdApi of crdApiGroups) {
+        for (const key in recommendationsMap) {
+            if (recommendationsMap.hasOwnProperty(key)) {
+                const value = recommendationsMap[key];
+                if (crdApi.startsWith(key)) {
+                    value.forEach((val: string) => extensionToPromote.add(val));
+                }
+            }
+        }
+    }
+
+    // remove already installed extensions
+    if (extensionToPromote.size > 0) {
+        for (const ext of vscode.extensions.all) {
+            if (extensionToPromote.has(ext.id)) {
+                extensionToPromote.delete(ext.id);
+            }
+        }
+    }
+
+    // get needed extensions description from json
+    const extDescription: RecommendedExtension[] = [];
+    recommendation.extensions.forEach((ext) => {
+        if (extensionToPromote.has(ext.id)) {
+            extDescription.push(ext);
+        }
+    });
+
+    recommendExtensionsToUser(extDescription, context.subscriptions);
+}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -86,6 +86,7 @@ import { ClusterExplorerNode, ClusterExplorerConfigurationValueNode, ClusterExpl
 import { create as activeContextTrackerCreate } from './components/contextmanager/active-context-tracker';
 import { WatchManager } from './components/kubectl/watch';
 import { ExecResult } from './binutilplusplus';
+import { recommendExtensions } from './components/extexplorer/explorer';
 
 let explainActive = false;
 let swaggerSpecPromise: Promise<explainer.SwaggerModel | undefined> | null = null;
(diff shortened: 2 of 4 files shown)


evidolob:
@lstocchi @itowlson  Can you look on this?
I can prototype more with new vscode webview api.

evidolob:
cc @squillace

squillace:
thanks for tagging me, @evidolob, this looks great. I like the experience, but we'll have to design how we detect and enable CRDs properly and fairly for all. I seem to recall we had a conversation like this. @itowlson do you recall?

itowlson:
@evidolob Apologies for the slow response on this one.  I think this is super interesting!  This feature really helps discoverability, and helps authors of CRD-based tools provide a truly end to end experience for their users.

I am not quite sure, however, what the user experience should be.  What you have will certainly do the job, but it doesn't yet feel consistent with the normal VS Code recommendation UX - the "there are recommendations for..." notification with a button that opens the Extensions tab on a marketplace search (which then brings in all the associated UI such as extension previews).  But I don't know if consistency with the VS Code recommendation UX is possible or practical, given that 1. we might want to present a number of specific extensions rather than a category search and 2. we don't want to have to reimplement a bunch of "don't show me again" style logic etc. that the built-in recommendation system provides for itself.

We also need to consider how to evolve the recommendations database.  I don't want to become responsible for evaluating, say, multiple Tekton extensions to decide which is "recommended."  Is our bar going to be "anyone can put in a PR recommendingy any extension and as long as it's not obviously evil then we'll give it our blessing"?  This feels like it risks our reputation (consider a malicious actor who creates a plausible extension, gets us to recommend it, and then modifies it to steal data).  A plain "search for extensions for CRD X" based on a keyword is more neutral than "we recommend" and would require no curation - but might provide correspondingly less value.

These are deep issues and I won't pretend that they're just nitpicks.  But this is a great first step in investigating those issues and I would be really eager to work out how we can address them and deliver this to users and CRD authors, because there is _so much goodness_ here just waiting to be unlocked.

evidolob:
@itowlson @squillace 
Sorry I was disrupted by other work.
I change a way how to recommended extensions shows. Now we have a CRD mapping to extension Tags and with recommendation I just perform search on marketplace with tahat tags:
ezgif com-gif-maker (14)

gorkem:
@evidolob @itowlson What do we need to do to move this PR forward? @squillace

squillace:
Hi @gorkem, I don't know, which is a fancy way of saying I think there aren't any blockers. 

Two last questions:
- What if the VS Code K8s extension were submitted to a publicly governed foundation? Would you as a user be comfy with the way we "suggest" things in that scenario?
- What about non-marketplace extensions? Can someone add a GH repo-based suggestion?

Apologies for this late questioning, as we're all swamped but the ux is just lovely....

gorkem:
> * What if the VS Code K8s extension were submitted to a publicly governed foundation? Would you as a user be comfy with the way we "suggest" things in that scenario?

I think open governance would make this even better. Ultimately this would be a PR to the `recommended-extensions.json` and would go through the same rules as discussions as any other PRs. We are probably trusting the project maintainers with even more critical features on this process so I am not worried.  

> * What about non-marketplace extensions? Can someone add a GH repo-based suggestion?
> 

Depends on the installation experience. If there is a way to provide a similar installation experience for GH repo-based extensions or alternate marketplaces I would be open to supporting those methods and adding them. However I do not think this should become a way to show information that is not backed up with an experience.

squillace:
Good responses! @itowlson whatchya think to that response? It's _almost as if he's thought of everything....._

itowlson:
I agree with @gorkem's view on how the list would be maintained and I agree that this would align really well with an open governance structure.  And I don't think there's any need to support non-marketplace extensions at this point - let's build it and see who comes.

We would want to document the process for getting your extension onto the list, and we would want to have some sense of how we ensure the list remains decent quality (like what if a 'recommended' extension is turned to evil, or simply falls out of date and no longer works well with current versions).

@gorkem @evidolob this PR is still marked as draft - are we now at the stage where we feel it's ready for review?

itowlson:
To be clear, I don't think the documentation and process stuff is a blocker - to some extent we can work that out as we go.  I'd love to get this in if we feel is it ready to review.

squillace:
I concur. I haven't DONE anything yet about concurring, but I continue to concur.

squillace:
whenever you're ready, @gorkem, let us know.

src/components/extexplorer/explorer.ts
```diff
@@ -0,0 +1,73 @@
...
+        } catch (err) {
+            // ignore
+        }
+    }
+    return Array.from(result);
+}
+
+async function readRecommendation(extPath: string): Promise<Recommendations> {
+    return new Promise((resolve, reject) => {
+        fs.readFile(path.join(extPath, 'recommendation', 'recommended-extensions.json'), (err, content) => {
```
itowlson, line 37: Cosmetic: I think we have some proper async `fs` functions around and tend to use those rather than manually constructing promises.

src/components/extexplorer/explorer.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (vscode.workspace.getConfiguration('extensions').get('ignoreRecommendations')) {
+        return;
+    }
+    const crdApiGroups = await getCRDApiGroups(kubectl);
+
+    const extensionToPromote = new Set<string>();
+    const recommendation = await readRecommendation(context.extensionPath);
+    // collect extension id to promote based on installed operators
+    for (const crdApi of crdApiGroups) {
+        for (const key in recommendation) {
```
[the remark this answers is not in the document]
itowlson: Oh this way gives you the type safety?

src/extension.ts
```diff
@@ -145,6 +146,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<APIBro
     const dependenciesProvider = new HelmDependencyDocumentProvider();
     const helmSymbolProvider = new HelmDocumentSymbolProvider();
     const completionProvider = new HelmTemplateCompletionProvider();
+    recommendExtensions(kubectl, context);
```
itowlson, line 149: Cosmetic: not sure this is the most natural place for this - this part of the code is mostly wiring up config.

src/extension.ts
```diff
@@ -83,6 +83,7 @@ import { ClusterExplorerNode, ClusterExplorerConfigurationValueNode, ClusterExpl
 import { create as activeContextTrackerCreate } from './components/contextmanager/active-context-tracker';
 import { WatchManager } from './components/kubectl/watch';
 import { ExecResult } from './binutilplusplus';
+import { recommendExtensions } from './components/extexplorer/explorer';
```
itowlson, line 86: Suggest something like `recommender` rather than `extexplorer`?

src/components/extexplorer/explorer.ts
```diff
@@ -0,0 +1,73 @@
...
+            if (recommendation.hasOwnProperty(key)) {
+                const value = recommendation[key];
+                if (crdApi.startsWith(key)) {
+                    value.forEach((val: string) => extensionToPromote.add(val));
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
itowlson: Oh... because you are passing this to extension search, you're okay with already-installed extensions appearing, because they'll be shown as installed.  And that's why you don't need to deduplicate either.

itowlson (commented):
This looks good - a few nitpicks/discussion points but nothing very substantive.

src/components/extexplorer/explorer.ts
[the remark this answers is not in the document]
evidolob: Basically it TS(vscode) snippet for `for in` loop:
```typescript
for (const key in object) {
    if (Object.prototype.hasOwnProperty.call(object, key)) {
        const element = object[key];
        
    }
}
```

src/components/extexplorer/explorer.ts
```diff
@@ -0,0 +1,73 @@
...
+
+interface Recommendations {
+    [crdApi: string]: string[];
+}
+
+async function recommendExtensionsToUser(tags: string[]): Promise<void> {
+    if (tags.length > 0) { // do not prompt if we cannot recommend anything
+        const answer = await vscode.window.showInformationMessage('Do you want to view recommended extensions for your Kubernetes cluster?', 'Ignore', 'Show Recommendation');
+        if (answer === 'Show Recommendation') {
+            const searchValue = tags.map((it) => `@tag:${it}`).join(' ');
```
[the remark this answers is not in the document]
evidolob: Yes, that search by tag.
In https://github.com/Azure/vscode-kubernetes-tools/pull/780#issuecomment-778855475 ,
You propose to provide recommendation with search, instead of recommending specific extensions:

>We also need to consider how to evolve the recommendations database. I don't want to become responsible for evaluating, say, multiple Tekton extensions to decide which is "recommended." Is our bar going to be "anyone can put in a PR recommendingy any extension and as long as it's not obviously evil then we'll give it our blessing"? This feels like it risks our reputation (consider a malicious actor who creates a plausible extension, gets us to recommend it, and then modifies it to steal data). A plain "search for extensions for CRD X" based on a keyword is more neutral than "we recommend" and would require no curation - but might provide correspondingly less value.

Maybe I miss understand your point.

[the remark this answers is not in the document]
itowlson: No, I had forgotten this discussion I'm afraid!  Thanks for reminding me!  I agree with past Ivan - tag search avoids maintenance and evaluation issues.  But in that case we _must_ say something like "*search for* extensions for your CRDs" not "recommended extensions" - recommended implies endorsement and curation, search implies only "we'll save you typing in the keyword yourself."

So, having been reminded of those thoughts - I now agree with what you are doing, but would like to see the UI at least changed from "show recommended extensions" to "search for extensions".

With that change we will be good to go I think!

src/components/recommender/extension-recommender.ts
```diff
@@ -0,0 +1,67 @@
...
+            res.forEach((i: any) => result.add(i));
+        } catch (err) {
+            // ignore
+        }
+    }
+    return Array.from(result);
+}
+
+async function readRecommendation(extPath: string): Promise<Recommendations> {
+    const fileContent = await fs.readFile(path.join(extPath, 'recommendation', 'recommended-tags.json'), 'utf8');
```
apupier, line 38: not sure in this case, but isn't preferable to use the vscode.fs.readFile API to have better chance to have it working in a remote environment?

apupier (commented):
is it possible to add indication in contributing (or readme) guide that there is this cool feature and how to fill `recommended-tags.json`?

src/components/extexplorer/explorer.ts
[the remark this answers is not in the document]
lstocchi: @evidolob you missed this point. We also need to change the statement `'Do you want to view recommended extensions for your Kubernetes cluster?'`. By citing Ivan "recommended implies endorsement and curation".
Maybe `Do you want to search for extensions based on CRDs that exist on your kubernetes cluster?"

src/components/recommender/extension-recommender.ts
evidolob: Done

src/components/extexplorer/explorer.ts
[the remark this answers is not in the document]
evidolob: OK, I change that message. Thx @lstocchi !

src/components/recommender/extension-recommender.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+async function getCRDApiGroups(kubectl: Kubectl): Promise<string[]> {
+    const result = new Set<string>();
+    const invokeResult = await kubectl.invokeCommand('api-versions');
+    if (invokeResult.resultKind === 'exec-succeeded') {
+        try {
+            const res = invokeResult.stdout.split('\n');
+            res.forEach((i: any) => result.add(i));
```
lstocchi, line 28: There is no need of having this i guess, split already returns an array and api-versions doesn't have duplicate (right?) so no need to use a set. We should add a verification that there are no empty values (eg last '\n'). 
Something like 
```
try {
 return invoke.....split('\n').filter(s => s not empty)
} catch .... {}
return [];
```

recommendation/recommended-tags.json
```diff
@@ -0,0 +1,11 @@
+{
+    "tekton.dev": [
+        "tekton"
+    ],
+    "operator.knative.dev": [
```
lstocchi, line 5: This doesn't cover all use cases. Atleast in my minikube knative is not discovered correctly. Below what i have. Maybe we can think of using only `knative.dev` or if this syntax can create any problem to add something more like `serving.knative` or `sources` ...

image

src/components/recommender/extension-recommender.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+    try {
+        const crdApiGroups = await getCRDApiGroups(kubectl);
+
+        const extensionToPromote = new Set<string>();
+        const recommendation = await readRecommendation(context.extensionPath);
+        // collect extension id to promote based on installed operators
+        for (const crdApi of crdApiGroups) {
+            for (const key in recommendation) {
+                if (recommendation.hasOwnProperty(key)) {
```
lstocchi, line 53: Is this necessary? You are already looping over recommendation keys, wouldn't this be always true?

lstocchi (changes requested):

recommendation/recommended-tags.json
evidolob: hm, yes that can be a problem, as CRD's can differs, it is depends on how you install them(with operator or not)

lstocchi:
Thanks @evidolob :pray:  @itowlson WDYT about it? Ready to be merged? We can then refine the documentation and discuss about "how to get an extension onto the list" in a new PR

lstocchi (approved):
@itowlson yes, that was fixed. LGTM :rocket:

PR #800, opened 2020-07-22
Describe shows a list of possible options

### LINKED ISSUE #793: [suggestion] "describe" should show a list of possible options
Similar to `>Kubernetes: Follow Logs` showing a list of possible pods to select from, it would be nice if `>Kubernetes: Describe` could show a list of possible resources to select from (e.g. `Deployments`, `Pods` ...etc) and further show a list of more possible options for a selected resource.

Example:

- User runs `>Kubernetes: Describe`
- Following list is displayed:
  ```
  Nodes
  Deployments
  Pods
  ... etc
  ```
- User selects `Deployments`
- Following list is displayed:
  ```
  foo
  bar
  hello-world
  ```
- If user selects a deployment (e.g. `foo`), then results for `describe deployment foo` appears in a searchable window
- If user does not select a deployment, then results for `describe deployment` appears in a searchable window

lstocchi:
It resolves #793 and it is built over #797 

I really liked the suggestion made by the user as the workflow we have now is far from being intuitive. 

In the current implementation, when the `describe` action is executed (without any node selected and/or any editor opened), users are shown an input box they need to fill in with the object they want to work with (e.g pods or pod/podName). 
And if an empty string is entered they are shown a quickpick with a limited set of options.  

image
image
image

With this PR the workflow changes a bit and skips the first step, showing directly the quickpick with all possible kinds user has access to (get, list verbs are checked)
image
image

If we agree at using this workflow we can also change the `delete`, `load` and `get` actions.


src/extension.ts
```diff
@@ -1279,9 +1283,12 @@ async function describeKubernetes(explorerNode?: ClusterExplorerResourceNode) {
             const nsarg = explorerNode.namespace ? `--namespace ${explorerNode.namespace}` : '';
             return { cmd: `describe ${explorerNode.kindName} ${nsarg}`, resourceKindName: explorerNode.kindName };
         } else {
-            const value = await findKindNameOrPrompt(kuberesources.commonKinds, 'describe', { nameOptional: true });
-            if (value) {
-                return { cmd: `describe ${value}`, resourceKindName: value };
+            const clusterKinds = await kubectlUtils.clusterResources(kubectl, ['get', 'list']);
+            if (clusterKinds.succeeded) {
+                const value = await findKindNameOrPrompt(clusterKinds.result, 'describe', { nameOptional: true, skipPrompt: true });
```
itowlson, line 1288: I'm not sure users always want to wait for us to load a list of resource kinds followed by a list of resources.  With the current workflow, if I know which resource I want to describe and am confident with the kubectl kind/name format (and the names are short!), I can just enter it.  The two queries to the cluster slow me down here.  And where I want the guided prompts, once I am used to the "hit enter" idiom, the overhead of that is pretty low (though the initial cognitive overhead is significant),

This is an area where I wish VS Code had a richer widget set - if we could have something where we could allow the user to type but still display a list of suggestions, that would solve so many problems.  Oh well!

I guess I am okay with it but maybe we should consider a setting to restore the old behaviour... though that adds more complexity and more documentation effort...

src/kubectlUtils.ts
```diff
@@ -459,6 +460,19 @@ export async function namespaceResources(kubectl: Kubectl, ns: string): Promise<
...
 
+export async function clusterResources(kubectl: Kubectl, verbs: string[]): Promise<Errorable<ResourceKind[]>> {
+    const arresult = await kubectl.readTable('api-resources -o wide');
+    if (ExecResult.failed(arresult)) {
+        return { succeeded: false, error: [ExecResult.failureMessage(arresult, {})] };
+    }
+
+    const resourceKinds: ResourceKind[] = arresult.result.filter((r) => verbs.every((verb) => r.verbs.includes(verb)))
+                                         .map((r) =>
+                                            new ResourceKind(r.name, r.name, r.kind, r.shortnames === '' ? r.kind : r.shortnames));
```
itowlson, line 471: This refers to 'shortnames' plural.  If a thing can have more than one shortname, we should be a bit careful about what we populate here.  It may not be an issue for the current use case, but once this function goes in, people will use it for all sorts of things and won't know about any potential shortname issues.  Mark my words...!

src/extension.ts
lstocchi: I was looking for a way to show the quickpick and also allow users to type something else (a mix between quickpick and inputbox). I guess we can make something similar as i found https://github.com/microsoft/vscode/issues/89601 ... but then i thought we would still have the multiple queries to the cluster so i opted for the setting as you suggested.

src/kubectlUtils.ts
lstocchi: I am still unsure how to handle this. I replaced the shortname with the name. I guess this should solve future potential issue.. ???

squillace:
@lstocchi lemme pull and workthis over this week. FINALLY. The work you've done here to find the right way is outstanding, btw.

itowlson:
@lstocchi Would you be able to rebase this please?  Happy to go with it, though it might need to queue up behind some infrastructure PRs...

itowlson (approved):
Looks good, thanks!

Issue #801, opened 2020-07-23
Do not throw error from callbacks of registered contributors for Yaml schema support


apupier:
see https://github.com/redhat-developer/vscode-yaml/issues/323

currently it is preventing other registered contributors to be handled. After improved in VS Code Yaml, there will be an error message.

If VS Code Kubernetes extension is unable to know if the yaml is a kubernetes file or not, it should return undefined, not throw an error.

PR #802, opened 2020-07-23
Do not throw error for yaml schema resolution #801

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>


apupier:
> i would also put a try catch around the `parse` call so if the function `ensureCache` is (or ever will be) called somewhere else inside the k8s extension we are covered. If an error is caught there, the cache for that file should be reset.
> 
> And i think we should also notify the repo owner of `node-yaml-parser` lib.

agree that you're proposed solution is better.

> If you don't have time for this, just let us know and we will handle it.

I will let you handle it. Thank you.

lstocchi:
Closing it as we have a fix at https://github.com/Azure/vscode-kubernetes-tools/pull/804

PR #806, opened 2020-08-17
add status bar items for viewing/changing current context and namespace

### LINKED ISSUE #777: Showing active namespace in status bar.
It would be good to show active namespace in the status bar as it is a root extension for `vscode-tekton`
and showing active namespace in the status bar will help us to know if we switch the namespace and current namespace which we are using.

for more detail Link: https://github.com/redhat-developer/vscode-tekton/pull/320#issuecomment-647353738

cc: @gorkem @evidolob @lstocchi @siamaksade

lstocchi:
it resolves #777 

This PR adds two items to the status bar: one which displays the current context and another one which displays the current namespace. (in the example below minikube is the context, tektontest the namespace)
image

By clicking on the context, the user can switch it (a quickpick shows up with a list of possible contexts).... the same for the namespace item.
Users can hide one or both items by using the optional settings in the extension.


lstocchi:
@sudhirverma i added a description how this patch works. Is it enough for your use case? Did you need anything else?

sudhirverma:
Thanks @lstocchi That would be enough for my use case for showing current namespace in the status bar.

cc: @siamaksade and @evidolob

squillace:
this looks nice and useful. @itowlson shall I merge, or would you like to peruse?

bnookala (approved):

PR #818, opened 2020-09-17
Adding cicd configure option 

kanika1894:
1. Adding the menu option for setting up ci/cd and browse pipeline
image

2. If the extension isn't installed, a warning message will appear and will open the window to install 'Deploy to Azure' extension
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -622,14 +622,14 @@
                     "group": "1"
                 },
                 {
-                    "command": "extension.ConfigureKubernetesPipeline",
+                    "command": "extension.ConfigueCICDPipeline",
                     "group": "7@2",
-                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes\\.providesKubeconfig/i"
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes/i"
                 },
                 {
-                    "command": "extension.BrowseKubernetesPipeline",
+                    "command": "extension.BrowseCICDPipeline",
                     "group": "7@3",
-                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes\\.providesKubeconfig/i"
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes/i"
                 }
             ],
             "commandPalette": [
@@ -718,11 +718,11 @@
                     "when": ""
                 },
                 {
-                    "command": "extension.ConfigureKubernetesPipeline",
+                    "command": "extension.ConfigueCICDPipeline",
                     "when": "never"
                 },
                 {
-                    "command": "extension.BrowseKubernetesPipeline",
+                    "command": "extension.BrowseCICDPipeline",
                     "when": "never"
                 }
             ]
@@ -1133,14 +1133,14 @@
                 "category": "Kubernetes"
             },
             {
-                "command": "extension.ConfigureKubernetesPipeline",
+                "command": "extension.ConfigueCICDPipeline",
                 "title": "Configure CI/CD",
-                "category": "Azure App Service"
+                "category": "Deploy to Azure"
             },
             {
-                "command": "extension.BrowseKubernetesPipeline",
+                "command": "extension.BrowseCICDPipeline",
                 "title": "Browse All Pipelines",
-                "category": "Azure App Service"
+                "category": "Deploy to Azure"
             }
         ],
         "keybindings": [
--- a/package.json
+++ b/package.json
@@ -1356,7 +1356,7 @@
     },
     "devDependencies": {
         "@types/mocha": "^2.2.32",
-        "@types/node": "^10.17.35",
+        "@types/node": "^10.2.0",
         "@types/request": "^2.48.1",
         "@types/unzipper": "^0.10.0",
         "@typescript-eslint/eslint-plugin": "^2.19.2",
--- a/package.json
+++ b/package.json
@@ -608,18 +608,23 @@
                 },
                 {
                     "command": "kubernetes.cloudExplorer.mergeIntoKubeconfig",
-                    "group": "7",
+                    "group": "7@1",
                     "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes\\.providesKubeconfig/i"
                 },
                 {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -608,18 +608,28 @@
...
                 },
                 {
                     "command": "kubernetes.cloudExplorer.findProviders",
                     "when": "view == kubernetes.cloudExplorer && viewItem == kubernetes.noProviders",
                     "group": "1"
+                },
+                {
+                    "command": "extension.ConfigureKubernetesPipeline",
+                    "group": "7@2",
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes\\.providesKubeconfig/i"
```
bishal-pdMSFT, line 627: do have any dependency on `.providesKubeconfig`? If no, then it should not be in the condition. On the other hand, should there be a check for Azure cloud provider?

package.json
```diff
@@ -608,18 +608,28 @@
...
+                    "group": "7@4",
                     "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes\\.providesKubeconfig/i"
                 },
                 {
                     "command": "kubernetes.cloudExplorer.findProviders",
                     "when": "view == kubernetes.cloudExplorer && viewItem == kubernetes.noProviders",
                     "group": "1"
+                },
+                {
+                    "command": "extension.ConfigureKubernetesPipeline",
```
bishal-pdMSFT, line 625: name should be `ConfigueCICDPipeline`

package.json
```diff
@@ -608,18 +608,28 @@
...
+                    "group": "7@2",
+                    "when": "view == kubernetes.cloudExplorer && viewItem =~ /kubernetes\\.providesKubeconfig/i"
+                },
+                {
+                    "command": "extension.BrowseKubernetesPipeline",
```
bishal-pdMSFT, line 630: name should be `BrowseCICDPipeline`

package.json
```diff
@@ -1113,6 +1131,16 @@
                 "command": "kubernetes.portForwarding.showSessions",
                 "title": "Show Port Forwarding Sessions",
                 "category": "Kubernetes"
+            },
+            {
+                "command": "extension.ConfigureKubernetesPipeline",
+                "title": "Configure CI/CD",
+                "category": "Azure App Service"
```
bishal-pdMSFT, line 1138: category is wrong

package.json
```diff
@@ -1113,6 +1131,16 @@
...
+            },
+            {
+                "command": "extension.BrowseKubernetesPipeline",
+                "title": "Browse All Pipelines",
+                "category": "Azure App Service"
```
bishal-pdMSFT, line 1143: here as well

package.json
kanika1894: what should be the category?

src/extension.ts
```diff
@@ -98,6 +99,9 @@ const clusterProviderRegistry = clusterproviderregistry.get();
 const configMapProvider = new configmaps.ConfigMapTextProvider(kubectl);
 const git = new Git(shell);
 const activeContextTracker = activeContextTrackerCreate(kubectl);
+const azurePipelinesExtensionId = 'ms-vscode-deploy-azure.azure-deploy';
```
bishal-pdMSFT, line 102: `deployToAzureExtensionId`

src/extension.ts
```diff
@@ -535,6 +541,40 @@ async function explain(obj: any, field: string): Promise<string | null> {
     });
 }
 
+async function browsePipeline(node: CloudExplorerTreeNode): Promise<void> {
+    if (await isAzurePipelinesExtensionInstalled()) {
```
bishal-pdMSFT, line 545: `isDeployToAzureExtensionInstalled`

src/extension.ts
```diff
@@ -535,6 +541,40 @@ async function explain(obj: any, field: string): Promise<string | null> {
...
+        await executeAzurePipelineExtensionCommand(browsePipelineCommand, node);
```
bishal-pdMSFT, line 546: `executeDeployToAzureExtensionInstalled`

src/extension.ts
```diff
@@ -535,6 +541,40 @@ async function explain(obj: any, field: string): Promise<string | null> {
...
+
+async function isAzurePipelinesExtensionInstalled(): Promise<boolean> {
+    const pipelinesExtension = vscode.extensions.getExtension(azurePipelinesExtensionId);
+    if (!pipelinesExtension) {
+        try {
+            vscode.window.showWarningMessage('Please install/enable `Deploy to Azure` extension and then retry.');
+        } catch (err) {
+            return false;
+        }
+        await vscode.commands.executeCommand('extension.open', azurePipelinesExtensionId);
```
bishal-pdMSFT, line 564: what does this do? will it open extension's marketplace page?

package.json
bishal-pdMSFT: Not really sure. where does this category show up to user?

bishal-pdMSFT: Ah this would show up in command paletter. But we are hiding this command from command palette, so it should not matter.
May be we can use "Deploy to Azure" category. That is what we do currently in our extension.

src/extension.ts
kanika1894: It opens the following page
image

package.json
kanika1894: Removed the dependency on kubeConfig. Not sure, how to put a condition on cloud provider.
Tried
``` kubernetes.cloudExplorer.provider.cloudName == 'Azure' ``` and some other options as well, didn't work.
Will have to check for the option

squillace:
OK, so this was different than our expectations, I think. The idea I recall discussing was to modify the AKS Cloud Provider extension (because then we know you have AKS interests) and then layer in a custom extension on top of the core VS K8s Extension API. 

This seems to modify the core extension, which we will not be doing. Do I misunderstand? The functionality's good, but I think it's attached to the wrong extension.

Tatsinnit:
💡 I think this functionality id shipped under this PR: https://github.com/Azure/vscode-aks-tools/pull/42 so might be this PR can be closed now. Thank you. cc: @bishal-pdMSFT & @kanika1894

kanika1894:
yes, @Tatsinnit , we can close this.

PR #833, opened 2020-11-05
Update debug-on-kubernetes.md

willemodendaal:
Just a little typo fix in the documentation.


itowlson (approved):
Good catch and thanks for taking the time to fix!

PR #850, opened 2020-12-31
Restrictive installer permission for kubectl.

Tatsinnit:
* Restrictive installer permission for kubectl. 

Probably more conformance to something like this: https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/components/installer/installer.ts#L137 

 Thanks, and cc: @itowlson and @squillace for eyeballing and thoughts. 🙏 💡


squillace:
@itowlson this looks like the right perm set to me.

itowlson (approved):
Thanks!

PR #851, opened 2020-12-31
Implementation for new Local Tunnel Debugger API

amsoedal:
Following up on this PR: https://github.com/Azure/vscode-kubernetes-tools-api/pull/31
New API PR: https://github.com/Azure/vscode-kubernetes-tools-api/pull/33

This PR includes the implementation for the new Local Tunnel Debugger API which adds the Debug (Local Tunnel) command and allows users to redirect traffic from their cluster to their machine when developing, allowing the benefits of remote and local debugging to be combined.


src/api/contract/localtunneldebugger/v1.ts
```diff
@@ -0,0 +1,15 @@
+// This module is contractual and should not be changed after release.
+// It should be in sync with vscode-kubernetes-tools-api/ts/localtunneldebugger/v1.ts
+// at all times.
+
+export interface LocalTunnelDebuggerV1 {
+    registerLocalTunnelDebugProvider(localRedirectionDebugger: LocalTunnelDebuggerV1.LocalTunnelDebugProvider): void;
```
daniv-msft, line 6: Nit: could you please rename this one?

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,122 @@
...
+with the keyword `kubernetes-extension-local-tunnel-provider` in `package.json`.
+For example:
+
+```json
+{
+    "keywords": [
+        "kubernetes",
+        "contoso",
+        "megafloof",
+        "kubernetes-extension-local-tunnel-provider"
```
daniv-msft, line 119: Should the tag be `kubernetes-extension-local-tunnel-debug-provider`? This seems to be the name used in the rest of the file.

daniv-msft: ...or `kubernetes-extension-local-tunnel-debugger`? Seems like we're using this name as well.
Are there any conceptual differences between both?

src/components/config/config.ts
```diff
@@ -262,3 +262,8 @@ export function isContextStatusBarDisabled(): boolean {
 export function isNamespaceStatusBarDisabled(): boolean {
     return vscode.workspace.getConfiguration(EXTENSION_CONFIG_KEY)['vs-kubernetes.disable-namespace-info-status-bar'];
 }
+
+// gets the default local tunnel debug provider to use
```
daniv-msft, line 266: Nit: Is this actually the `default`? Should it be more the `debug provider defined in settings` or something of that kind?

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,46 @@
...
+            .then((selection: string | undefined) => {
+                if (selection === browseExtensions) {
+                    vscode.commands.executeCommand('extension.vsKubernetesFindLocalTunnelDebugProviders');
+                }
+            });
+            return;
+        }
+
+        let providerName: string = getLocalTunnelDebugProvider();
+        if (providerName === "") {
```
daniv-msft, line 28: Is it possible that this `providerName` is `undefined`, if the user manually remove this field from the settings? Or do we have the guarantee that it will always be set to an empty string if missing?

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    register(provider: LocalTunnelDebugProvider): void {
+        console.log(`Registered local tunnel debugger type ${provider.id}`);
+        this.providers.push(provider);
+    }
+
+    startLocalTunnelDebugProvider(target?: any): void {
+        const browseExtensions = "Find Providers on Marketplace";
+        if (this.providers.length === 0)
+        {
```
daniv-msft, line 17: Nit: could you please put this bracket on the above line, to respect the convention in the rest of the file?

src/extension.ts
```diff
@@ -2324,7 +2328,15 @@ async function showInfoMessage(message: string) {
 }
 
 function kubernetesFindCloudProviders() {
-    const searchUrl = 'https://marketplace.visualstudio.com/search?term=kubernetes-extension-cloud-provider&target=VSCode&category=All%20categories&sortBy=Relevance';
+    searchMarketPlace("kubernetes-extension-cloud-provider");
+}
+
+function kubernetesFindLocalTunnelDebugProviders() {
+    searchMarketPlace("kubernetes-extension-local-tunnel-provider");
+;}
```
daniv-msft, line 2336: Nit: extra semicolon

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,122 @@
+# Adding Providers to the Local Tunnel Debugger
+
+The Local Tunnel Debugger is intended to provide a local debugging option that includes
+the benefits of remote debugging. Traffic from the cluster is rerouted to the version of the service
```
daniv-msft, line 4: Nit: maybe `application` instead of `services`? Users might not only debug services

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,46 @@
+import * as vscode from 'vscode';
+
+import { LocalTunnelDebugProvider } from './localtunneldebugger.extension';
+import { getLocalTunnelDebugProvider } from '../config/config';
+
+export class LocalTunnelDebugger {
+    private readonly providers = Array.of<LocalTunnelDebugProvider>();
```
Tatsinnit, line 7: 💡 adding some thoughts, looks like cool feature. Is there any chances of this being `undefined`? or there is a default value fallback for this?

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,46 @@
...
+            return;
+        }
+
+        let providerName: string = getLocalTunnelDebugProvider();
+        if (providerName === "") {
+            // If no provider is configured in the settings, take the first one that's registered
+            providerName = this.providers.map((p) => p.id).sort()[0];
+        }
+        const providerToUse: LocalTunnelDebugProvider | undefined = this.providers.find((p) => p.id === providerName);
+        if (providerToUse === undefined) {
```
Tatsinnit, line 33: 💡 is the value of `LocalTunnelDebugProvider` always guaranteed, aka in case of empty array this `if` can be bypassed.

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export class LocalTunnelDebugger {
+    private readonly providers = Array.of<LocalTunnelDebugProvider>();
+
+    register(provider: LocalTunnelDebugProvider): void {
+        console.log(`Registered local tunnel debugger type ${provider.id}`);
+        this.providers.push(provider);
+    }
+
+    startLocalTunnelDebugProvider(target?: any): void {
```
Tatsinnit, line 14: 💡 Idea\thought share: possible thought process behind below is: bail early if none of the conditions satisfy.

Given below 3 major if conditions, if:

* `providers.length` is `non-empty` and 
* `providename` if `non-empty` and 
*  `providerToUse` is `not undefined`, there is nothing which probably gets run?

In that case should we do all the existence check upfront and then bail with some message that you have got all necessary stuff and nothing actually happened?

Tatsinnit (commented):
I added few idea shares, as this PR feature looks a cool idea. mainly all my thoughts\points below is around philosophy of possibly `succeed early if all conditions satisfy` or `bail early is some don't`

src/components/localtunneldebugger/localtunneldebugger.ts
Tatsinnit: * Minor idea addition: we could compartmentalize these conditions into its own small `functs` units, and call it from main function as well.

docs/extending/localtunneldebugger.md
amsoedal: Yeah, the **debug provider** would be the extension that actually provides the debugging functionality, while the **debugger** is the layer that keeps track of the debug providers and connects the Debug (Local Tunnel) command to the provider's startDebugging API. (I tried to copy the way the CloudExplorer/CloudProvider interfaces interact.) In line with this, I'll go with "kubernetes-extension-local-tunnel-debug-provider" :)

src/api/contract/localtunneldebugger/v1.ts
amsoedal: Yes, thanks for catching this!

src/components/config/config.ts
amsoedal: Yeah, it's the one defined in settings. Will update

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: Hmm, good question. I did set the default value in package.json, and from the [docs](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) it looks like even if they remove it from settings.json the fallback value should get resolved? However, this is an easy case to handle so I'll updating to include the `undefined` case just in case. I see some of the other configurations in this file have varying ways of handling this

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: Thanks!

src/extension.ts
amsoedal: Thanks1

docs/extending/localtunneldebugger.md
amsoedal: Sure!

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: @Tatsinnit  thank you so much for adding these comments! Dan told me you had messaged him and mentioned forking this PR. Perhaps we can sync up later today? I've just addressed Dan's feedback, hopefully it won't cause any conflicts. :)

Tatsinnit: Thank you so much @amsoedal, Yeah I dint wanted to just comment :) was planning to possibly add a PR to your Fork and see if that helps. Yes, please feel free to ping me internally in teams please. I think Dan will have my alias, since its git I will not ask your internal alias here.

src/components/localtunneldebugger/localtunneldebugger.extension.ts
```diff
@@ -0,0 +1,4 @@
+export interface LocalTunnelDebugProvider {
```
Tatsinnit, line 1: 💡 Idea share for consistency and what others think? since this is the `interface`, we could rename this file to `localTunnelDebuggerProvider` as the interface name is **Or** rename the interface and file as `localtunneldebuggerinterface` 

* Might be worth to at least rename file same as the class name aka: rename from `localtunneldebugger.extension.ts` to `localtunneldebuggerprovider`

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        let providerName: string | undefined = getLocalTunnelDebugProvider();
+        if (!providerName) {
+            // If no provider is configured in the settings, take the first one that's registered
+            providerName = this.providers.map((p) => p.id).sort()[0];
+        }
+        const providerToUse: LocalTunnelDebugProvider | undefined = this.providers.find((p) => p.id === providerName);
+        if (providerToUse === undefined) {
+            vscode.window.showWarningMessage(`You have configured VSCode to use Local Tunnel debugger '${providerName}', but it is not installed.`, browseExtensions)
```
Tatsinnit, line 33: 💡 See line 17, just wondering why this is `WarningMessage` vs `InformationMessage`? reason being this rest of the calling mechanism for both UI are same, so we can easily extract this to common static funct.

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,45 @@
...
+            .then((selection: string | undefined) => {
+                if (selection === browseExtensions) {
+                    vscode.commands.executeCommand('extension.vsKubernetesFindLocalTunnelDebugProviders');
+                }
+            });
+            return;
```
Tatsinnit, line 39: 💡 Idea share: this `returns undefined` but main signature returns `void` as type annotation i.e. `startLocalTunnelDebugProvider(target?: any): void `

src/components/localtunneldebugger/localtunneldebugger.extension.ts
amsoedal: That's a good point! I will rename the interface & the file to "localTunnelDebuggerProvider" and after I merge your PR. Thanks for pointing that out!

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: Hey @Tatsinnit, would you mind clarifying this one this a little more? :) I was trying to follow the way cloudexplorer keeps track of its providers, but if I'm doing something unusual here please do let me know!

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: I see you consolidated these in the PR, it looks good to me!

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: Ah I see -- Thanks for fixing in your PR! I want to make sure I'm following the conventions. :)

src/components/localtunneldebugger/localtunneldebugger.ts
Tatsinnit: Thanks and sounds good in that case (if things look good) you can merge my PR and then give it a test please. (It should flow through this PR automatically) :)

src/components/localtunneldebugger/localtunneldebugger.ts
Tatsinnit: I think the PR for the fort below covered some parts of this: 

* Just for other readers: https://github.com/Azure/vscode-kubernetes-tools/pull/851#discussion_r552996801

src/components/localtunneldebugger/localtunneldebugger.extension.ts
amsoedal: @Tatsinnit @daniv-msft  I noticed that the "normal" debugging providers in the same repo (e.g. `dotnetdebugprovider.ts`) use this "debug provider" naming. I think "debug provider" sounds more colloquial than "debugger provider", even though the latter would technically be more consistent. But I'm open to changing it as I want to make sure I'm following the conventions of the repo & keeping things orderly!

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: @Tatsinnit I merged the PR, will resolve any comments I think were addressed by it, but please feel free to reopen them or open new ones if you have more thoughts. I had to tweak a couple things (the version of tslint used in the CI build was unhappy about the use of "??", and I updated the error conditions as the latter message was always getting used). Otherwise I definitely think the code looks cleaner, really appreciate the help on that!

src/components/localtunneldebugger/localtunneldebugger.extension.ts
daniv-msft: Sounds good, thanks @amsoedal. Yes, if `Xdebugprovider.ts` is already in use in the codebase, I agree we should follow the existing convention and stick to "debug provider".

amsoedal: Cool, thanks for the feedback. I renamed the `localtunneldebugger.extension.ts` file to `localtunneldebugprovider` like you suggested @Tatsinnit, and added the debug option to the deployments like you mentioned offline @daniv-msft. Let me know if there's more changes you guys would like to see or if it looks OK.

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,122 @@
...
+tunnel debugging capabilities can register themselves as a Local Tunnel Debug provider in order
+to hook into the Kubernetes extension's API.
+
+## Elements of a debug provider
+
+An object that provides content for the Local Tunnel Debugger is called a _local tunnel debug provider_.
+Providers must be hosted within a Visual Studio Code extension.  This table summarises what
+your providers and their hosting extension need to do; the rest of this article goes into detail.
+
+| Component            | Responsibilities                                                     |
```
Tatsinnit, line 18: 💡  You can achieve this representation: 

image

From this code:

```
| Component            | Responsibilities                                                     |
|----------------------|----------------------------------------------------------------------|
| Your extension       | <ul><li>Activate when the Cluster Explorer is displayed or when the Debug (Local Tunnel) option is selected in the command palette</li><li>Register debug providers with Kubernetes extension</li><li>Tag itself as a local tunnel debug provider in the Visual Studio Marketplace</li></ul>        |
| Local Tunnel Debug provider| <ul><li>Implement the local tunnel debug provider interface</li><li>Allow users to configure which debug provider to use</li></ul>  |
| Kubernetes extension | Display the Debug (Local Tunnel) command in the palette or when interacting with supported resources in the Cluster Explorer                                 |
```

Tatsinnit (commented):
Thank you so much @amsoedal Looks cool, the Final PR eye balling and key final take \ decision of how this command is hooked can be done by @itowlson , Thank you to both of you for time and letting me contribute to this PR. 🙏:)

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,122 @@
...
+interface LocalTunnelDebugProvider {
+    readonly id: string;
+    startDebugging(target?: any): void;
+}
+```
+
+### Implementing the metadata
+
+The `id` should be a user-friendly name for your debug provider, such as `Contoso Trampoline Debugger`.  
+The id will allow users to configure a default local tunnel debugger to use.
```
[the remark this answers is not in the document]
amsoedal: Good catch! Thanks!

docs/extending/localtunneldebugger.md
amsoedal: Thanks so much! It looks so much better now :) I've updated the table items to be more descriptive as well.

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,122 @@
...
+
+In your extension's `activate` function, you must register your provider(s) using the
+Kubernetes extension API.  The following sample shows how to do this using the NPM
+helper package; if you don't use the helper then the process of requesting the API is
+more manual but the registration is the same.
+
+```javascript
+const MY_PROVIDER = {
+    id: "Contoso Trampoline Debugger",
+    startDebugging: (target?: any) => startDebugSession(target)
```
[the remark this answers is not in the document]
amsoedal: Oo, good question, that probably is confusing. I've added a sample implementation to localtunneldebugger.md, but should I take the extra step of renaming this method as well?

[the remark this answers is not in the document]
Tatsinnit: 💡 Sounds like a great idea. I think worth spending time to think right `funct` name to resonate right idea for end user. I will do a heavy +1 for this idea for making thinks more simpler and readable \ understandable. Thank you.

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,143 @@
...
+}
+```
+
+## Registering the debug provider
+
+In order to hook into the Debug (Local Tunnel) option, a local debug provider must be _registered_
+with the Kubernetes extension. This is the responsibility of the VS Code extension that hosts
+the debug provider. To do this, the extension must:
+
+* Activate in response to the `kubernetes.cloudExplorer` view
```
itowlson, line 80: I think you indicated Cluster Explorer rather than Cloud Explorer before?

In any case, is this actually required or could an extension register only when the user issues the command?  (Perhaps there are timing issues registering in response to the command - if so we should consider those.)

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,143 @@
...
+* Activate in response to the `Debug (Local Tunnel)` command
+* Request the Kubernetes extension's Local Tunnel Debug Provider API
+* Call the `register` method for each cloud provider it wants to display
```
itowlson, line 83: "cloud provider"?

package.json
```diff
@@ -275,6 +277,10 @@
                         "vs-kubernetes.disable-namespace-info-status-bar": {
                             "type": "boolean",
                             "description": "Disable displaying your current Kubernetes namespace in VS Code's status bar."
+                        },
+                        "vs-kubernetes.local-tunnel-debug-provider": {
+                            "type": "string",
+                            "description": "Configure the Local Tunnel debug provider you want to use by default."
```
itowlson, line 283:
```suggestion
                            "description": "The Local Tunnel debug provider you want to use by default."
```

docs/extending/localtunneldebugger.md
```diff
@@ -0,0 +1,143 @@
...
+    startDebugging: (target?: any) => startDebugSession(target)
+};
+
+function startDebugSession(target?: any) {
+    const clusterExplorer = await k8s.extension.clusterExplorer.v1;
+    const commandTarget = clusterExplorer.resolveCommandTarget(target);
+
+    if (commandTarget) {
+        if (commandTarget.nodeType === 'resource') {
+            if (commandTarget.resourceKind.manifestKind === 'Pod') {
```
itowlson, line 58: It would be good to mention in this section what kinds of resources a provider implementation should handle.  But I wonder if we need to put that burden on providers - e.g. will every provider need the same logic and UI for going from a Deployment to a Pod, and if so can we provide this in the core?

src/api/contract/localtunneldebugger/v1.ts
```diff
@@ -0,0 +1,15 @@
+// This module is contractual and should not be changed after release.
+// It should be in sync with vscode-kubernetes-tools-api/ts/localtunneldebugger/v1.ts
+// at all times.
+
+export interface LocalTunnelDebuggerV1 {
+    registerLocalTunnelDebugProvider(localTunnelDebugProvider: LocalTunnelDebuggerV1.LocalTunnelDebugProvider): void;
+    startLocalTunnelDebugProvider(target?: any): void;
```
itowlson, line 7: Is the idea of this to allow an extension to trigger local tunnel debugging?  If so, maybe drop `Provider` from the name?

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,48 @@
...
+            message = `You have configured VSCode to use Local Tunnel debugger '${providerName}', but it is not installed.`;
+        }
+        else if (!this.providers.length) {
+            message = 'You do not have a Local Tunnel Debug Provider installed.';
+        }
+
+        LocalTunnelDebugger.displayUIandExecuteCommand(message);
+    };
+
+    static displayUIandExecuteCommand(message: string) {
```
itowlson, line 36: Suggest e.g. `promptToFindOnMarketplace`

src/components/localtunneldebugger/localtunneldebugger.ts
```diff
@@ -0,0 +1,48 @@
...
+        this.providers.push(provider);
+    }
+
+    startLocalTunnelDebugProvider(target?: any) {
+        const providerSetting: string | undefined = getLocalTunnelDebugProvider();
+        const providerName: string | undefined = providerSetting ? providerSetting : this.providers.map((p) => p.id).sort()[0];
+        const providerToUse: LocalTunnelDebugProvider | undefined = this.providers.find((p) => p.id === providerName);
+
+        // On success start early.
+        if (this.providers.length && providerToUse) {
```
itowlson, line 19: Do you need `this.providers.length`?  If `providerToUse` is set then I think that's all we need.

src/extension.ts
```diff
@@ -2324,7 +2328,15 @@ async function showInfoMessage(message: string) {
...
 function kubernetesFindCloudProviders() {
-    const searchUrl = 'https://marketplace.visualstudio.com/search?term=kubernetes-extension-cloud-provider&target=VSCode&category=All%20categories&sortBy=Relevance';
+    searchMarketPlace("kubernetes-extension-cloud-provider");
+}
+
+function kubernetesFindLocalTunnelDebugProviders() {
+    searchMarketPlace("kubernetes-extension-local-tunnel-debug-provider");
+}
+
+function searchMarketPlace(searchTerm: string) {
```
itowlson, line 2338: super nitpick sorry
```suggestion
function searchMarketplace(searchTerm: string) {
```

itowlson (changes requested):
This looks terrific - thank you @amsoedal!  Most of the comments are nitpicks or typoes, but some are in docs where it would be good to clear them up.  Apart from that I think just one question about activation, another about the 'other' API function which I'm not sure of the intended use of, and an open-ended muse about whether we can/should do more setup in the core prior to calling into the provider.  But this is looking great - I'm really excited to see this in action with Bridge to Kubernetes.

src/api/contract/localtunneldebugger/v1.ts
amsoedal: Oh I see, e.g. "startLocalTunnelDebugging"? Yes, the idea is this actually starts the debug provider's debugging method.

src/components/localtunneldebugger/localtunneldebugger.ts
amsoedal: Makes sense!

src/extension.ts
amsoedal: Hehe thanks for catching that!

src/api/contract/localtunneldebugger/v1.ts
amsoedal: I went with "startLocalTunnelDebugSession", but I'm open to changes ofc.

docs/extending/localtunneldebugger.md
amsoedal: Yes, you're right, thank you for catching that. Regarding the timing, I agree that it shouldn't be a requirement for the extension to activate unless the actual command is getting executed. Perhaps I can make this bullet "suggested" rather than a "must"

docs/extending/localtunneldebugger.md
amsoedal: Good catch!

docs/extending/localtunneldebugger.md
amsoedal: Ooh, good point. I will definitely add which resources the provider implementation is expected to handle. Regarding whether or not to provide this in the core, let me sync offline about @daniv-msft about that, but if in practice every provider would be reimplementing the same logic to resolve the pod behind a deployment then that sounds reasonable to provide it up front.

docs/extending/localtunneldebugger.md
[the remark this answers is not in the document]
amsoedal: Makes sense. I changed it to `startLocalTunnelDebugging` but I'm open to suggestions. :)

src/api/contract/localtunneldebugger/v1.ts
itowlson: If this starts the provider's session shouldn't it be a method on the provider rather than a method exposed by the Kubernetes extension?

I feel I don't have this straight in my mind yet.  Could you explain _who_ calls this method (the k8s extension, the provider, the extension containing the provider, or a third-party extension) and _when_?  Thanks and apologies for not getting it.

amsoedal: No worries at all, it could definitely be a naming issue or that I'm not using the correct pattern. Basically, in extension.ts we bind the Debug (Local Tunnel) command to the method exposed by the k8s extension: 

```
registerCommand('extension.vsKubernetesDebugLocalTunnel', (target?: any) => { localTunnelDebugger.startLocalTunnelDebugSession(target); }),
```
And then inside the implementation of the method in component/localtunneldebugger.ts, we see if a provider is installed, do some tie-breaking if multiple providers are available, and then call into the provider that we chose:

```
    startLocalTunnelDebugSession(target?: any) {
        const providerSetting: string | undefined = getLocalTunnelDebugProvider();
        const providerName: string | undefined = providerSetting ? providerSetting : this.providers.map((p) => p.id).sort()[0];
        const providerToUse: LocalTunnelDebugProvider | undefined = this.providers.find((p) => p.id === providerName);

        // On success start early.
        if (providerToUse) {
            providerToUse.startLocalTunnelDebugging(target);
            return;
        }
```
Let me know if that makes sense, or if I can flatten this in a better way!

itowlson: It makes sense, but that makes it sound like this is a method that is called only from _within_ the extension (from `extension.ts`), not by external consumers of the API.  In which case it should not be part of the API, just a plain exported function from the implementation module.  Unless the _provider_ extension needs to call it, there's no need to put it in that part of the API.

(More generally, unless a method gets called across extension boundaries - core calling into provider or provider calling into core - then it doesn't need to be in the API at all.  But this interface is specifically one that the provider uses to call into the core.)

Hope this makes sense - very happy to tee up a call if that would be a more efficient way to get us on the same page.

amsoedal: Hey Ivan, that does make sense! Thanks for clarifying that. I don't know why I was confusing what needed to be in the API vs. what simply needs to be exported. :) I've removed the method from the interface and I tested to make sure everything was still able to snap together like it should, but don't hesitate to let me know if there's anything else I can fix, stylistically or otherwise.

docs/extending/localtunneldebugger.md
amsoedal: I synced with Dan offline, and we felt that at the moment it might not make sense to offer this in the core of the extension, as different providers could have different preferences regarding how to resolve a pod from a deployment. However this is definitely up for discussion, if you or the team have different thoughts about it!

amsoedal:
@itowlson I've addressed the feedback, thank you all again for the good comments & cooperation! If you want to see the feature in action, I'd be happy to schedule a call sometime this week and do a demo. Otherwise, please don't hesitate to let me know if there's more changes you'd like to see.

itowlson (approved):
This looks good - thanks so much for driving this through and for your patience with my niggling.

We already have a 1.2.4 release in progress, and to avoid muddying the water, I'm going to avoid merging this until after that is complete (hopefully tomorrow).  After that I'd like to do a 1.3.0 soon after (end of this week or early next week) accompanied by an updated SDK (which will also include some Configuration API changes) so that you can launch the fully integrated experience.  I'll keep you in the loop but please chase me if things are not happening as quickly as you'd like.

Issue #728, opened 2020-04-07
Automate release process


itowlson:
Currently the release process for the extension is:

* I go "what ho, I think I'll do a release"
* I create a publish-x.x.x branch to update the version and the changelog (and any other docs etc. that didn't get updated in the course of the feature work)
* I build a VSIX locally and do some smoke testing
* I create a PR and merge the publish branch
* I pull upstream/master, build a VSIX locally and do another smoke test
* I create a release using the GitHub UI and drag the VSIX onto the release form
* I run `vsce publish` to push the VSIX to the marketplace

Yeah, I know.  Don't judge me.

This is, to say the least, highly manual and highly dependent on me not borking my local environment.  It would be good if we could automate this to something like:

* "What ho"
* Create publish-x.x.x branch, test locally, create PR [still manual]
* Possibly GH runs a few additional sanity checks e.g. making sure that `package.json` and `changelog.md` are updated **[automated]**
* Merge pull request [still manual]
* GitHub builds the VSIX **[automated]**
* GitHub creates a release using the version number from `package.json` **[automated]**
* GitHub pushes VSIX to marketplace **[automated]**

Longer term we should figure out how to smoke test as part of the automated checks on PRs, but this will be tough given the amount of interactivity and dependencies...

dcermak:
>     * "What ho"

Not sure if this can be automated :wink:

> 
>     * Create publish-x.x.x branch, test locally, create PR [still manual]
> 
>     * Possibly GH runs a few additional sanity checks e.g. making sure that `package.json` and `changelog.md` are updated **[automated]**
> 
>     * Merge pull request [still manual]
> 
>     * GitHub builds the VSIX **[automated]**

A standard github action should be able to do this.

> 
>     * GitHub creates a release using the version number from `package.json` **[automated]**

This could be automated via https://github.com/actions/create-release and https://github.com/actions/upload-release-asset

> 
>     * GitHub pushes VSIX to marketplace **[automated]**

This can be achieved using this github action: https://github.com/HaaLeo/publish-vscode-extension

lstocchi:
Closing it as we have already automated the release process.

PR #854, opened 2021-01-18
Automate release process (#728)

lstocchi:
This is related to https://github.com/Azure/vscode-kubernetes-tools/issues/728

The workflow has to be triggered manually and it does almost all the job needed to publish the extension to the marketplace.

- Build the extension
- Check that version used in the `package.json` file is mentioned in `changelog.md`
- Create a new release (if another one with the same version exists, it fails)
- Publish it to the marketplace

The only secret required is the marketplace token.

The only edit we need to make inside the project is about the changelog, a version should be written like `## [x.x.x]` because the action follows this convention.

By having this workflow in place we can also make a release when Ivan is busy on other projects. We just need to agree when a new release can be made.

Only thing missing is the testing process, not sure if we want to write some unit/integration tests or there was some reason to not adding them.

Any idea how this can be improved? @itowlson @squillace @gorkem

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Build & Publish
 on: [workflow_dispatch]
 
 jobs:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,55 @@
+name: CI
+on: [workflow_dispatch]
+
+jobs:
+  build:
+    name: release
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: 14
+    # Run install dependencies
+    - name: Install dependencies
+      run: npm install
+    # Run tests
+    - name: Build 
+      run: npm run compile
+    - name: Get current package version
+      id: package_version
+      uses: martinbeentjes/npm-get-version-action@v1.1.0
+    - name: Check version is mentioned in Changelog
+      uses: mindsers/changelog-reader-action@v2.0.0
+      with:
+        version: ${{ steps.package_version.outputs.current-version }}
+        path: 'CHANGELOG.md'
+    - name: Create a Release
+      id: create_release
+      uses: actions/create-release@v1
+      env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
+      with:
+        tag_name : ${{ steps.package_version.outputs.current-version}}
+        release_name: ${{ steps.package_version.outputs.current-version}}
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+        pat: ${{ secrets.VS_MARKETPLACE_TOKEN }}
+        registryUrl: https://marketplace.visualstudio.com
+    - name: Attach vsix to release
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ steps.create_release.outputs.upload_url }}
+        asset_path: ${{ steps.create_vsix.outputs.vsixPath}}
+        asset_name: ${{ steps.create_vsix.outputs.vsixPath}}
```
jeffmaury, line 52: Won't it display the whole path on the release page ?

lstocchi: This is the outcome https://github.com/lstocchi/vscode-automate-release/releases/tag/0.0.7 . The vsix is linked to the release as it is done in the previous releases here.

itowlson:
This is great!

One thing we should do along with this is write some maintainer documentation, even if it's just a "How to Release" paragraph that says:

* Create a branch `publish-x.y,z`
* Update `package.json` with the new version
* Add a section to `CHANGELOG.md` with the header `## <new version>`
* Create PR, get approval, merge
* Run the `CI` (or whatever) workflow manually from the GH Actions tab

Yes, it would be really good if we could write some smoke tests.  These used to exist, but what they tested was not very useful, and they fell into disrepair.  I am not quite sure how we would test in a meaningful way given how much of the extension is either UI or needs a k8s cluster or both!

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
+name: CI
```
itowlson, line 1: As this actually releases/publishes a build, perhaps name it "Publish" or something?

lstocchi:
> One thing we should do along with this is write some maintainer documentation, even if it's just a "How to Release" paragraph that says:

Done, give it a look please. 

> Yes, it would be really good if we could write some smoke tests. These used to exist, but what they tested was not very useful, and they fell into disrepair. I am not quite sure how we would test in a meaningful way given how much of the extension is either UI or needs a k8s cluster or both!

Yes i understand what you mean, but even if we add few tests they are better than nothing imo. Also a simple "when you click `Debug: Attach` the quickpick displays these options" can be helpful so we are aware if something breaks up and we are a bit less nervous when we make a release :laughing: . I'll try to setup something using kind  so we can start from that one.

itowlson:
Docs look good and I've created the marketplace token.  This is awesome - thanks so much!

:shipit: :shipit: :shipit:

PR #862, opened 2021-01-25
Publish 1.2.3

itowlson:
This is basically the bundling change, plus a couple of nonfunctional changes (documentation fixes).  I'm suggesting we publish this as its own release despite the lack of new features, because that way we can easily detect the impact of this change in isolation, and roll back without withdrawing features if it turns out to be a disaster!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## [1.2.3]
+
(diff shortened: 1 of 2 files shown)


itowlson:
@lstocchi I really want to do the automated release one immediately after this one!

lstocchi (approved):
LGTM :rocket:

squillace (approved):

PR #869, opened 2021-01-29
Run smoke test in CI workflow

itowlson:
Sets up infrastructure for running tests on PRs.

**NOTE:** This is currently only a PR so that I can verify whether I have tests running okay in the GitHub workflow.  The tests are **not** real and this should **not** be merged.


itowlson:
@lstocchi This now looks like it is running tests on PRs - do you have any thoughts on what we could put in as starters for our test suite?

lstocchi:
@itowlson i would say to start writing those tests that you usually do manually before a release. What do you test? You look at the PR merged and test those things or do some other basic tests? 

Something useful could be to create a workflow that set up a kubernetes cluster with kind and we execute our code on that one. So we are sure everything works if kubectl changes.

lstocchi:
What's the status here? Can this be merged? Today i was working on #895 and i thought that it would be great to have some tests that check the url created and if it really exists.

itowlson:
Yes, this can be merged -- with the understanding that it does not actually introduce any tests!  I'll remove the dummy test and ship.

PR #872, opened 2021-02-01
Centralise Cluster Explorer API implementation

itowlson:
There is a lot of shared behaviour in the Cluster Explorer V1 and V1.1 API implementations (reflecting their significant overlap in the API surface).  We have a new API version coming down the line for a new tree layout, and I am hoping to add some new node sources before too long, which would be another API version.  So it would be good to centralise so that we can more easily review only what has changed rather than reviewing implementations files that are marked entirely as new code.

### DIFF
--- a/src/api/implementation/cluster-explorer/common.ts
+++ b/src/api/implementation/cluster-explorer/common.ts
@@ -1,0 +1,167 @@
+/* eslint-disable camelcase */
+
+import * as vscode from 'vscode';
+// import { KubernetesExplorer, KUBERNETES_EXPLORER_NODE_CATEGORY } from "../../../components/clusterexplorer/explorer";
+import { ExplorerExtender /* , ExplorerUICustomizer */ } from "../../../components/clusterexplorer/explorer.extension";
+import { /* CustomGroupingFolderNodeSource, CustomResourceFolderNodeSource, */ NodeSourceImpl } from "../../../components/clusterexplorer/extension.nodesources";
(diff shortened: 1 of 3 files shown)


src/api/implementation/cluster-explorer/common.ts
```diff
@@ -0,0 +1,198 @@
+/* eslint-disable camelcase */
+
+import * as vscode from 'vscode';
+// import { KubernetesExplorer, KUBERNETES_EXPLORER_NODE_CATEGORY } from "../../../components/clusterexplorer/explorer";
```
lstocchi, line 4: this can be removed

lstocchi (approved):
Good idea. LGTM.
I see that there is also other code that could be centralised. Any reason why you didn't do that? 
E.g 
```
resolveCommandTarget(target?: any): ClusterExplorerV1_1.ClusterExplorerNode | undefined {
        if (!target) {
            return undefined;
        }
        if (target.nodeCategory === KUBERNETES_EXPLORER_NODE_CATEGORY) {
            const implNode = target as ClusterExplorerNode;
            const apiNode = adaptKubernetesExplorerNode(implNode);
            return apiNode;
        }
        return undefined;
    }
```
we could create  a method in common
```
resolveTarget(explorer, target): ClusterExplorerV1 | V1_1 | undefined;
```
then call it from both impementations.

The same for `registerNodeContributor`, `registerNodeUICustomizer`, `nodeSources`

itowlson:
I did `resolveCommandTarget` and `nodeSources` - didn't bother with the other functions because they basically do all the work in common functions anyway and only the interactions with `this.explorer` remain in the class.

PR #883, opened 2021-02-08
Show chart and app version in Helm release history

### LINKED ISSUE #881: Feature: add chart version and other meta to helm releases view
Would you please add at least the cart version to the helm Releases (history) view, to get detailed information what release one might want to restore. The chart release ids which are just increasing number ate not helpful at al.

image
image

itowlson:
Fixes #881.

This has the info as a tooltip - is that going to work for you @cforce?  If not let's discuss good ways of presenting this extra info.

cc @shreyaskarnik as I don't seem to be able to add you as a reviewer

### DIFF
--- a/src/components/clusterexplorer/node.helmrelease.ts
+++ b/src/components/clusterexplorer/node.helmrelease.ts
@@ -20,7 +20,8 @@ export class HelmHistoryNode extends ClusterExplorerNodeImpl implements ClusterE
     readonly nodeType = NODE_TYPES.helm.history;
     getTreeItem(): vscode.TreeItem | Thenable<vscode.TreeItem> {
         const updatedTime = moment(this.release.updated).fromNow();
-        const treeItem = new vscode.TreeItem(`${this.release.revision} - ${this.release.status} (${updatedTime})`, vscode.TreeItemCollapsibleState.None);
+        const treeItem = new vscode.TreeItem(`${this.release.revision}`, vscode.TreeItemCollapsibleState.None);
+        treeItem.description = `${this.release.status} (${updatedTime})`;
         treeItem.command = {
             command: "extension.helmGet",
             title: "Get",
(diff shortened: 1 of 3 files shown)


shreyaskarnik:
@itowlson checking out the PR. Also on a sidenote apologies for https://github.com/Azure/vscode-kubernetes-tools/pull/877 did not realize I was changing contracts. :(  Thanks for catching and fixing.

shreyaskarnik:
@itowlson Works great!

shreyaskarnik:
In terms of the helm release tree there is this example 
image 

which I like from the docker extension where additional information can be displayed a little differently with font weights. Maybe the uptime can be displayed with a different font weight which can draw more attention to release number prompting the hover or click to gather more information.

itowlson:
I really like this idea.  That 'less prominent' effect is done via the `TreeItem.description` property so this should be a fairly easy change.  Do you suggest I roll it into this PR, or would you prefer to hold off and do it as a separate enhancement?

shreyaskarnik:
@itowlson either way would work fine., I want to be respectful of your time. Reason I did bring this example up was to potentially use it in other tree items.

itowlson:
@shreyaskarnik This is slightly off topic, but looking at the code, it looks like clicking a history node loads the _current_ release, not the one that was selected?  `helm get all` does have a `--revision` option - should we pass that?

shreyaskarnik:
@itowlson checking this out.

itowlson:
@shreyaskarnik Here is how it looks with the description - is this what you envisaged and if so are you happy with it?

image

shreyaskarnik:
@itowlson that looks great! I can repro the behaviour with clicking of helm history node and it shows only the last node. This is due to the fact that `getChildren` is empty and the command you proposed will fix that.

itowlson:
Thanks!  I don't want the 'load correct release' fix to get squashed in with the metadata and formatting changes, so I've raised #884 to track that.

lstocchi (approved):

PR #895, opened 2021-02-15
fix install dependencies is hardcoded to amd64

### LINKED ISSUE #889: Installing dependencies is hardcoded to amd64
When I attempted to run this extension on an instance that was running on an arm64 architecture, I got an "exec format error" after installing dependencies through the extension.

After further inspection, it turns out that the extension hardcodes `amd64` in several places as the platform for which to download binaries such as `kubectl`, `helm`, etc. See, for example, [here](https://github.com/Azure/vscode-kubernetes-tools/blob/929a87c101bbe46d5233f88c37c4868079b88846/src/components/installer/installer.ts#L35).

This code should automatically detect the relevant platform for which to download binaries.

lstocchi:
it resolves #889

### DIFF
--- a/src/components/installer/installationlayout.ts
+++ b/src/components/installer/installationlayout.ts
@@ -17,9 +17,21 @@ export function formatBin(tool: string, platform: Platform): string | null {
     if (!platformString) {
         return null;
     }
-    const toolPath = `${platformString}-amd64/${tool}`;
+    const platformArchString = platformArch(platformString);
+    const toolPath = `${platformString}-${platformArchString}/${tool}`;
     if (platform === Platform.Windows) {
         return toolPath + '.exe';
     }
     return toolPath;
+}
+
+export function platformArch(os: string) {
+    if (os !== 'linux') {
+        return 'amd64';
+    }
+    switch (process.arch) {
+        case 'arm': return 'arm';
+        case 'arm64': return 'arm64';
+        default: return 'amd64';
+    }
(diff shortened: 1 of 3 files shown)


src/components/installer/installationlayout.ts
```diff
@@ -17,9 +17,21 @@ export function formatBin(tool: string, platform: Platform): string | null {
...
+    const platformArchString = platformArch(platformString);
+    const toolPath = `${platformString}-${platformArchString}/${tool}`;
     if (platform === Platform.Windows) {
         return toolPath + '.exe';
     }
     return toolPath;
 }
+
+export function platformArch(os: string) {
+    if (os !== 'linux') {
```
itowlson, line 29: This might be a silly question, but is there any reason to only check the architecture on Linux?  Why not just check regardless of the OS?

itowlson (approved):
That was super quick - thanks!  I did have one minor question but happy to merge regardless.

src/components/installer/installationlayout.ts
lstocchi: I saw that there are tools that are only packaged for amd64. Kubectl only comes in amd64 for Darwin and Windows.

nilbot: @itowlson Could I resurrect this since `minikube`, `kubectl` etc now have darwin-arm64 releases? This also affects #889

PR #898, opened 2021-02-25
add webpack __dirname rule

### LINKED ISSUE #897: Error opening dashboard with minikube cluster
On a minikube cluster (1.17.0) i cannot open the dashboard with the "Open Dashboard" menu selection from the Kubernetes Extension for vscode since version 1.2.3 and above. I have this message / error : 

`Something went wrong when ensuring the Kubernetes API Proxy output stream existed`

It work fine with extension version 1.2.2... Any ideas ? 

Thanks

lstocchi:
it resolves #897 

@itowlson found this https://webpack.js.org/configuration/node/#node__dirname . Now the `__dirname` is resolved to `/` and it probably fails because it doesn't have permission to write the proxy file there. By adding this rule `__dirname` should be resolved  like before. Try it whenever you can.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -45,6 +45,9 @@ const config = {
           use: ['file-loader']
       }
     ]
+  },
+  node: {
+    __dirname: false
   }
 };
 module.exports = config;


itowlson (approved):
Looks good - thanks!

Issue #831, opened 2020-10-22
Add option for showing timestamps in logs


arsnyder16:
kubectl logs --timestamps

itowlson:
This shouldn't be too hard at the code level but we need to consider the user experience,  I think we now have a webview for logs?  So a possible UX could be a checkbox on the webview to show/hide timestamps, with a an option to configure the default.

Or we could just always provide timestamps - what do people think?

lstocchi:
Closing it. New log webview offers this.

Issue #828, opened 2020-10-15
Refactor logs view 


lstocchi:
Opening this as a remainder bc I closed https://github.com/Azure/vscode-kubernetes-tools/pull/773. The guy who opened that PR is not able to work on that anymore.

Description of the new logs view at https://github.com/Azure/vscode-kubernetes-tools/pull/773#issuecomment-648544485

tvvignesh:
@lstocchi @itowlson I guess adding colors to the log would add a lot of usability. Currently, its difficult to distinguish log levels or other messages during development cause of this in web view.

![image](https://user-images.githubusercontent.com/1165845/100925985-beadbe00-3508-11eb-9318-0f858046b1a0.png)

lstocchi:
@tvvignesh I agree. Hopefully i should have some free time in a week to start working on all the enhancements asked for the log webview. :crossed_fingers: ...and sorry for the lack of updates on this.

sdwerwed:
This freezes my VS Code because logs are too big, RAM and CPU are getting too high (and wouldn't blame the laptop, it is a good one).
Could we give it a bit higher priority as it is not just an enhancement, it will prevent issues like this. 

(I never coded in this typescript so not sure how much helpful I would be here)

lstocchi:
Closing it as new log viewer has been merged and released. Please open a new issue if you face this or other problems. Thank you!!

Issue #798, opened 2020-07-21
Logs output is not in color when using webview


mojo2012:
Hi guys,

awesome extension!

One thing that bugs me though, is that the log output does not interpret the ANSI color codes correctly when using the webview. In the terminal it works of course.

It would be awesome if you would add that feature, because using the webview/editor window is way more comfortable than the terminal.

Cheers

lstocchi:
Thanks for opening this @mojo2012 . We are in the process to refactor the log view https://github.com/Azure/vscode-kubernetes-tools/pull/773 so we will surely keep this into consideration. Thanks again for the suggestion.

lstocchi:
Closing it as new log viewer has been merged and released. Please open new issues if you face any problem or have some ideas how to enhance it. Thank you!!

PR #905, opened 2021-03-10
Refactor logs view

lstocchi:
Finally i had some time to look into the several issues/suggestions regarding our log webview.

I took inspirations from the Ivan's idea and the discussion we had in #773 + other issues found around.

AFAIK this patch is related to #773, #828, #831, #798, #737

The log webview opens up empty with only the toolbar which consists of two rows.
image

The **first row** contains all options that can be useful to catch logs
* Select the container (if only one available the select box is hidden)
* Show or follow logs
* Show timestamp
* Only return logs newer than a relative duration
* Lines of recent log file to display
* Redirect logs to the terminal

Once the run button is clicked the `stop` button (if in follow mode) and `clear` button are displayed.
The `reset` just reset all options to their default values.

The **second row** contains all options to work on the current logs.
* Filters (as before)
* Wrap lines (users were complaining about lines being too long)
* Scroll to bottom (as before)

To highlights logs i took the regex patterns (edited them a bit) from https://marketplace.visualstudio.com/items?itemName=emilast.LogFileHighlighter (as made by vscode itself https://code.visualstudio.com/updates/v1_20#_output-panel). 
Unfortunately i couldn't take advantage of the existing extension as that doesn't work with webviews. Maybe we can improve it in future releases (mentioning it here so we can give credits to that guy).

I guess that's all :exploding_head: Have fun :rocket: 

Example how this works
logview-1

cc @squillace @gorkem


squillace:
@lstocchi what the EFF cazzarolla! this is amazing

gorkem:
TBH. I was not a big fan of log views inside editor area so far but this made me change my mind. definite +1 from me.

.vscode/launch.json
```diff
@@ -16,7 +16,7 @@
             "outFiles": [
                 "${workspaceFolder}/dist/**/*.js"
             ],
-            "preLaunchTask": "npm: watch",
+            "preLaunchTask": "npm: test-compile",
```
itowlson, line 19: Not sure about this as part of this PR - if we want to do this let's break it out for further discussion.  We have direct experience of webpack introducing errors and I'd greatly prefer to pick those up at this stage of the cycle.

package.json
```diff
@@ -1240,7 +1220,9 @@
     "scripts": {
         "vscode:prepublish": "webpack --mode production",
         "lint": "eslint -c .eslintrc.js --ext .ts ./src",
-        "compile": "webpack --mode none",
+        "compile": "npm run compile:ext && npm run compile:view",
```
itowlson, line 1223: Again, let's break these out into a separate "Curse All Webpacks, Jeeves" PR.

schema/colors.json
```diff
@@ -0,0 +1,107 @@
+[
```
itowlson, line 1: Is this file specific to k8s logs and if so should we name it accordingly?

src/components/logs/app/index.html
```diff
@@ -0,0 +1,150 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Kubernetes logs</title>
+    <!-- meta http-equiv="Content-Security-Policy" -->
```
itowlson, line 6: ~What does it mean for this to be empty - most restrictive or least restrictive?  The VS Code folks do like us to have a reasonably restrictive CSP if possible.~ Never mind - I see it gets replaced at runtime.

src/components/logs/app/index.html
```diff
@@ -0,0 +1,150 @@
...
+                <vscode-option value='exclude'>that don't match</vscode-option>
+                <vscode-option value='after'>after match</vscode-option>
+                <vscode-option value='before'>before match</vscode-option>
+            </vscode-select>
+            <div style="width: 350px; display: inline-block">
+                <vscode-inputbox placeholder="Enter your key words" id='filter-input'></vscode-inputbox>
+            </div>
+            
+            <div class="tooltip ml-10">Wrap lines:
+                <span class="tooltiptext">Long words should be broken and wrap onto the next line.</span>
```
itowlson, line 133: The tooltip seems at odds with the caption.  Does this control whether lines wrap at all, or is the idea that lines _always_ wrap and this merely controls _where_ they are allowed to break?

itowlson (changes requested):
This looks really nice - thanks for sorting out the quagmire.  A couple of nitpicks about possible unrelated changes, otherwise looks great!

src/components/logs/app/index.html
lstocchi: I'm a bit confused by your question but by answering it it should be the second option.
Let me explain. This checkbox changes the css rule used to display logs, `white-space: pre` or `white-space: pre-wrap`. 
In both cases white-spaces and line-break are honoured. The difference is that in the first case (checkbox not checked and `white-space: pre` rule used) we don't care if the lines are wider than the parent (by simplifying the vscode webview in our case). We just follow the actual text, when there is a new-line, a new-line is printed otherwise everything will go in one row.
image

In the second case (checkbox checked and `white-space: pre-wrap` rule used), we still honour white-space and break-line from the original text but we also impose that a line will not be wider than the parent. In this way the logs are fully visible in our webview and one doesn't need to scroll horizontally.
image

So this checkbox just control where a line is allowed to break, somehow. Hopefully this answer your question.

By reading the tooltip maybe i got what your doubt, we don't break words but lines??
Feel free to edit it by yourself or to propose a better tooltip as i'm kinda lost here. Thank you!

.vscode/launch.json
lstocchi: I forgot to have changed it, i had some issue to debug it. Reverted :+1:

package.json
lstocchi: Done, i also noted i didn't edit the `vscode:prepublish` and i guess we need to take care of that as well.

schema/colors.json
lstocchi: Changed name and moved from the schema folder as this is not related to schemas. If you have a better idea for naming it please let me know, or feel free to edit it yourself :+1:

itowlson (approved):
Looks good - excited to get this out there! Thanks!

squillace:
doubleplusgood

PR #906, opened 2021-03-15
Updated ingress.yml snippet to use networking.k8s.io/v1

### LINKED ISSUE #901: Snippets: ingress.yaml should use apiVersion: networking.k8s.io/v1
Snippet [`ingress.yaml`](https://github.com/Azure/vscode-kubernetes-tools/blob/master/snippets/ingress.yaml) currently uses `apiVersion: extensions/v1beta1`, which is deprecated in v1.14+ (and will be removed in v1.22). It should therefore be replaced with `apiVersion: networking.k8s.io/v1`.

Proposed body for the snippet, in accordance with the [docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ingressservicebackend-v1-networking-k8s-io): 
```
apiVersion: networking.k8s.io/v1
  kind: Ingress
  metadata:
    name: ${1:myingress}
    labels:
        name: ${1:myingress}
  spec:
    rules:
    - host: ${2:<Host>}
      http:
        paths:
        - pathType: Prefix
          path: "/"
          backend:
            service:
              name: ${3:<Service>}
              port: 
                number: ${4:<Port>}
```

Persistent13:
This resolves #901

### DIFF
--- a/snippets/ingress.yaml
+++ b/snippets/ingress.yaml
@@ -2,21 +2,21 @@ name: ingress
 label: Kubernetes Ingress
 description: Kubernetes Ingress
 body: |2
-apiVersion: networking.k8s.io/v1
-  kind: Ingress
-  metadata:
-    name: ${1:myingress}
-    labels:
-        name: ${1:myingress}
-  spec:
-    rules:
-    - host: ${2:<Host>}
-      http:
-        paths:
-        - pathType: Prefix
-          path: "/"
-          backend:
-            service:
-              name: ${3:<Service>}
-              port: 
-                number: ${4:<Port>}
+  apiVersion: networking.k8s.io/v1
+    kind: Ingress
+    metadata:
+      name: ${1:myingress}
+      labels:
+          name: ${1:myingress}
+    spec:
+      rules:
+      - host: ${2:<Host>}
+        http:
+          paths:
+          - pathType: Prefix
+            path: "/"
+            backend:
+              service:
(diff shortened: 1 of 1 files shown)


snippets/ingress.yaml
```diff
@@ -1,8 +1,4 @@
-name: ingress
-label: Kubernetes Ingress
-description: Kubernetes Ingress
-body: |2
```
bacongobbler, line 4: Can you add these back? It looks like you not only changed the apiVersion but you also changed the snippet.

bacongobbler (changes requested):

snippets/ingress.yaml
Persistent13: Oops! Sorry about that

snippets/ingress.yaml
```diff
@@ -1,8 +1,8 @@
 name: ingress
 label: Kubernetes Ingress
 description: Kubernetes Ingress
-body: |2 
-  apiVersion: extensions/v1beta1
+body: |2
+apiVersion: networking.k8s.io/v1
```
bacongobbler, line 5: needs to be indented 2 spaces to match.

bacongobbler (changes requested):

itowlson:
@bacongobbler is this okay to merge now?

bacongobbler (approved):

PR #920, opened 2021-04-13
Add stronger check on binary result

lstocchi:
Long story short, if you use an old kubectl binary with a new kubernetes cluster (e.g kubectl 1.15 on kubernetes cluster 1.20.2) the `config view` command may fail

Screenshot from 2021-04-13 11-37-19

which makes the extension behave in a wrong way (empty tree and a generic json parse error displayed)

This fix allows to show the correct messages to the user
Screenshot from 2021-04-13 12-37-25

### DIFF
--- a/src/binutilplusplus.ts
+++ b/src/binutilplusplus.ts
@@ -247,7 +247,7 @@ export async function invokeForResult(context: Context, command: string, stdin: 
         return { resultKind: 'exec-failed', execProgram: context.binary, command };
     }
 
-    if (sr.code === 0) {
+    if (sr.code === 0 && sr.stderr === '') {
         return { resultKind: 'exec-succeeded', execProgram: context.binary, command, stdout: sr.stdout };
     }
 


itowlson:
This is a slightly tricky one - not helped by my poor memory I fear.  In the olden days we used to check that stderr was empty before returning success, just as you propose.  But - and this is where my memory lets me down - we _did_ see cases where the command succeeded _but some additional information was written to stderr_.

Now we could say "Ivan's vague memories are not a rigorous basis for programming decisions" and I would sympathise with that.  So we could merge this, and if we got any regressions, well, we could fix them.  But that does make me feel jittery.

How practical would it be to capture stderr on success results, and make the decision of how to interpret 'code 0 with error' at a higher level?

lstocchi:
> This is a slightly tricky one - not helped by my poor memory I fear. In the olden days we used to check that stderr was empty before returning success, just as you propose. But - and this is where my memory lets me down - we _did_ see cases where the command succeeded _but some additional information was written to stderr_.
> 
> Now we could say "Ivan's vague memories are not a rigorous basis for programming decisions" and I would sympathise with that. So we could merge this, and if we got any regressions, well, we could fix them. But that does make me feel jittery.
> 
> How practical would it be to capture stderr on success results, and make the decision of how to interpret 'code 0 with error' at a higher level?

I agree about trying to handle this in a different way if you remember that. There are so many combinations about kubectl/kubernetes-based products that I have no doubt that a similar situation (command succeed with stderr not empty) could happen. 

So what you're proposing is to create a new function within `execResult` that accepts a shellResult and returns the correct type (`exec-failed` or `exec-succeeded`)? Or you want to create a new `execResult` subtype that represent a `code 0 with error` response? In this case wouldn't it be less practical to handle this new subtype everywhere?

itowlson:
I guess at the `ExecResult` level I am maybe thinking just add a `stderr` field to the Succeeded case.  I don't think we need to create a distinct enum case for it.

PR #922, opened 2021-04-14
Binaries installed with wrong permissions should be fixed

itowlson:
We had an issue where binaries we installed were too broadly writeable.  That is now fixed, but users who allowed old versions of the extension to do an install will be left with poorly configured files.

This fix checks the extension's tool installation directory at startup and fixes any executable files to be user-writeable only.  To avoid performing this check on every startup, it records if there has been a successful pass and skips the directory walk if so.


Tatsinnit (approved):
:shipit:  ❤️  looks nice and having a record of the legacy-writable binaries is awesome. Thank you.

lstocchi (approved):

PR #924, opened 2021-04-17
Fetch custom crd schemas from active cluster

### LINKED ISSUE #917: How to have completion for Custom Resource Definitions
based on https://github.com/Azure/vscode-kubernetes-tools/pull/583#issue-298696614 i would expect to have completion in this case:
```
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: bitcoin-to-kafka
spec:
  
```
The kameletBinding CRD deployed 
Screenshot from 2021-03-30 15-14-19

I would expect the "integration" completion https://github.com/apache/camel-k/blob/9e384110be106b4970af2e967e28e0877fa78359/config/crd/bases/camel.apache.org_kameletbindings.yaml#L66 but it is not the case.

I also tried to associate the "kubernetes" schema to the file, but in this case I have this error on the "kind" value:
```
Value is not accepted. Valid values: "StorageVersion", "DaemonSet", "Deployment", "ReplicaSet", "StatefulSet", "TokenRequest", "TokenReview", "LocalSubjectAccessReview", "SelfSubjectAccessReview", "SelfSubjectRulesReview", "SubjectAccessReview", "HorizontalPodAutoscaler", "Scale", "Job", "CronJob", "CertificateSigningRequest", "Lease", "LimitRange", "Namespace", "Node", "PersistentVolume", "PersistentVolumeClaim", "Pod", "ReplicationController", "ResourceQuota", "Service", "Ingress", "FlowSchema", "PriorityLevelConfiguration", "IngressClass", "NetworkPolicy", "RuntimeClass", "PodDisruptionBudget", "PodSecurityPolicy", "CSIDriver", "CSINode", "VolumeAttachment", "CustomResourceDefinition", "APIService".yaml-schema: https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.20.5-standalone-strict/_definitions.json(1)
```

What am i missing to have the completion/validation/hover available?

lstocchi:
This resolves #917 

Until now the plugin only downloaded the swagger json from the active cluster which contains minimal schemas for custom crds. An example is the camel-k operator as mentioned in #917. The minimal schema for the kameletbinding type is 
```
{
  "id": "camel.apache.org/v1alpha1@KameletBinding",
  "apiVersion": "camel.apache.org/v1alpha1",
  "kind": "KameletBinding",
  "type": "object",
  "name": "org.apache.camel.v1alpha1.KameletBinding"
}
```
while the crd on the cluster contains the actual schema 
```
"versions": [
            {
                "additionalPrinterColumns": [
                    {
                        "description": "The Kamelet Binding phase",
                        "jsonPath": ".status.phase",
                        "name": "Phase",
                        "type": "string"
                    }
                ],
                "name": "v1alpha1",
                "schema": {
                    "openAPIV3Schema": {
                        "description": "KameletBinding is the Schema for the kamelets binding API",
                        "properties": {
                            "apiVersion": {
                                "description": "APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources",
                                "type": "string"
                            },
                            "kind": {
                                "description": "Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds",
                                "type": "string"
                            },
                            "metadata": {
                                "type": "object"
                            },
                            "spec": {
                                "description": "KameletBindingSpec --",
                                "properties": {
                                    "integration": {
                                        "description": "Integration is an optional integration used to specify custom parameters",
                                        "properties": {
                                            "configuration": {
                                                "items": {
                                                    "description": "ConfigurationSpec --",
                                                    "properties": {
                                                        "type": {
                                                            "type": "string"
                                                        },
                                                        "value": {
                                                            "type": "string"
                                                        }
                                                    },
                                                    "required": [
                                                        "type",
...........
...........
```

So users didn't have autocompletion or schema validation at all.
This patch adds the logic to download and use all schemas (when available) for all crds and their multiple versions. 
Below a short demonstration of the cool autocompletion we obtain for the camel-k operator

kamelet

@apupier it would be awesome if you could test it and tell me if it works as expected. Keep in mind that when the plugin starts it fetch everything and then feed the YAML plugin with the schemas so if you have your `kamelet` file already opened when the k8s extensions activates you won't see anything. So close it and reopen it.

### DIFF
--- a/src/components/swagger/swagger.ts
+++ b/src/components/swagger/swagger.ts
@@ -3,6 +3,7 @@ import { Errorable, failed } from "../../errorable";
 import { Kubectl } from "../../kubectl";
 import { proxy } from "../kubectl/proxy";
 import { fs } from '../../fs';
+import { ExecResult } from '../../binutilplusplus';
 
 export type Swagger = any;
 
@@ -33,4 +34,42 @@ export async function getClusterSwagger(kubectl: Kubectl): Promise<Errorable<Swa
     } finally {
         proxySession.dispose();
     }
+}
+
+export async function getCrdSchemas(kubectl: Kubectl): Promise<{[key: string]: object | undefined} | undefined> {
+    const er = await kubectl.invokeCommand(`get crd -o jsonpath="{range .items[*].spec}[{.group}, {.names.plural}, {.names.singular}, {range .versions[?(.schema.openAPIV3Schema.properties)]}{.name}{\\" \\"}{end}]{end}"`);
+
+    if (ExecResult.failed(er)) {
+        kubectl.reportFailure(er, { whatFailed: 'Failed to get crds' });
+        return;
+    }
+    try {
+        const crdSchemasMapping: { [key: string]: object } = {};
+        const crdInfos = er.stdout.substring(1).split('[');
+        for (const crdInfo of crdInfos) {
+            const crdInfoAsArray = crdInfo.split(',');
+            const group = crdInfoAsArray[0].trim();
+            const plural = crdInfoAsArray[1].trim();
+            const singular = crdInfoAsArray[2].trim();
+            const versions = crdInfoAsArray[3].trim() !== '' ? crdInfoAsArray[3].trim().split(' ') : [];
+            if (versions.length > 0) {
+                const crdSchemas = await getSchemas(kubectl, `${plural}.${group}`);
+                for (const version of versions) {
+                    const versionObject = crdSchemas.spec.versions.filter((v: { name: string }) => v.name === version);
+                    crdSchemasMapping[`${group}/${version}@${singular}`] = versionObject.length > 0 ? versionObject[0].schema.openAPIV3Schema.properties : undefined;
+                }
+            }
+        }
+        return crdSchemasMapping;
+    } catch (e) { }
+    return;
+}
+
+async function getSchemas(kubectl: Kubectl, crd: string) {
+    const json = await kubectl.asJson<any>(`get crd ${crd} -o json`);
+    if (failed(json)) {
+        return;
+    }
+    return json.result;
+
 }
--- a/src/components/swagger/swagger.ts
+++ b/src/components/swagger/swagger.ts
@@ -3,6 +3,7 @@ import { Errorable, failed } from "../../errorable";
 import { Kubectl } from "../../kubectl";
 import { proxy } from "../kubectl/proxy";
 import { fs } from '../../fs';
+import * as kubectlUtils from '../../kubectlUtils';
 import { ExecResult } from '../../binutilplusplus';
 
 export type Swagger = any;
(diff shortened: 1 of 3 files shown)


itowlson:
That's odd. k8s is meant to include the schemas for CRDs in its Swagger.  I recall doing significant work on active context tracking to make this work when it first came in.  At the time I think it was a config option to include the CRD schemas, but it was planned to  make it the default in what, 1.16? 1.17?  Maybe that didn't happen...?

itowlson:
Gosh, the k8s documentation is a shambles.  I found this:

"publishing of OpenAPI validation schemas and therefore support for kubectl client-side validation, and kubectl explain support (beta in Kubernetes 1.15)" (https://v1-18.docs.kubernetes.io/blog/2019/06/20/crd-structural-schema/)

Still struggling to find anything on the current status though.

itowlson:
"With the OpenAPI v2 Publishing feature enabled, CustomResourceDefinition OpenAPI v3 validation schemas which are structural and enable pruning (opt-in in v1beta1, enabled by default in v1) are published as part of the OpenAPI v2 spec from Kubernetes API server."

https://v1-17.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2

Implies it still needs enabling while in beta, but also says stable as of 1.16:

"FEATURE STATE: Kubernetes 1.16 stable"

"Note: OpenAPI v2 Publishing is available as beta since 1.15, and as alpha since 1.14. The CustomResourcePublishOpenAPI feature must be enabled, which is the case automatically for many clusters for beta features."

So yeah, I'm utterly confused.

itowlson:
According to https://v1-18.docs.kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-graduated-or-deprecated-features the `CustomResourcePublishOpenAPI` defaults to true from 1.16 and was removed as an option from 1.18.

lstocchi:
@itowlson sorry i was PTO yesterday.

> That's odd. k8s is meant to include the schemas for CRDs in its Swagger. 

I was expecting the same but after searching it for days i haven't found anything. This is why i opted for the "manual" way. Not optimal as i also discovered that jsonpath has some limitations. If there is some way to make kubernetes makes the job for us would be great.

bacongobbler:
Yeah, we had to do something simliar for Helm to include CRDs into the OpenAPI validation schema. Here's how it's done in Helm:

https://github.com/helm/helm/blob/43853ea772e779b617b011a8d4e69205636fa4f9/pkg/kube/client.go#L75-L84

bacongobbler (approved):

Issue #146, opened 2018-03-29
Need more strict validation on specific value in YAML file


songkaiape:
For example: 
The ports value should be an integer between 0 ~65536. But plugin only validate if the value is integer. Invalid value like 888888 or -1 should not pass validation.

ImagePullPolicy should only be one of Always, Never, IfNotPresent. Now it only validate if it is a string

songkaiape:
screenshot
![image](https://user-images.githubusercontent.com/4318793/38072397-a5be5256-3358-11e8-954b-a5ffaa27112f.png)

bacongobbler:
Helm 3 introduces JSONSchema support for chart values (see https://github.com/helm/helm/pull/5350). Once Helm 3 lands, the extension could load the `values.schema.json` (if present) to provide validation against what the user has provided in the values.yaml. That should handle use cases such as the few cases provided here.

thernstig:
@bacongobbler I just tried to add a `values.schema.json`. I also added this:

```json
  "yaml.schemas": {
    "charts/server/values.schema.json": [
      "charts/server/values.yaml"
    ]
  }
```
and noticed that when I tried to auto-comlete things in in values.yaml in VS Code, it did not give me IntelliSense for the items I had specified. Since Helm 3 has been out for quite some time now, (and Helm 2 has been deprecated), is there a chance this feature could get some priority? :)

I have no idea how you relate this to the original issue in this thread though, so if you want me to I can make this a separate issue. The original issue by @songkaiape is about a Deployment manifest. That has nothing to do with `values.schema.json` which should only be about validating the `values.yaml` file.

thernstig:
In addition, https://github.com/redhat-developer/vscode-yaml/issues/196 seems to have an issue here as well. What is the reason the `values.yaml` file has this identifier? I need to manually change the language mode to `yaml` for the `yaml.schemas` setting to work with my `values.yaml`.

```sh
"languages": [
            {
                "id": "helm",
                "aliases": [
                    "helm-template",
                    "helm"
                ],
                "filenamePatterns": [
                    "values.yaml",
                    "Chart.yaml",
                    ......
```

PR #927, opened 2021-04-21
Remove values.yaml association with helm-template

ezramorris:
values.yaml is for user-specified values rather than a Helm chart specification. Associating with helm-template causes issues when using JSON schema validation, as noted in #146, redhat-developer/vscode-yaml#196, and others.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1199,7 +1199,6 @@
                     "helm"
                 ],
                 "filenamePatterns": [
-                    "values.yaml",
                     "Chart.yaml",
                     "requirements.yaml",
                     "**/templates/*.yaml",


bacongobbler:
Just to clear something up here... The initial statement is false. values.yaml is part of the Helm chart specification. Users can pass custom values via the --values or --set flags.

The Helm chart file structure is defined here:

https://helm.sh/docs/topics/charts/#the-chart-file-structure

You can see that values.yaml is part of the file structure.

That being said, I can understand the issues related to the values being related with the helm-template syntax highlighter. It should be rendered as plain yaml. Anything in the `templates` directory should be treated with the `helm-template` syntax highlighter.

Because of that, this LGTM.

bacongobbler (approved):

ezramorris:
Sorry, poor phrasing on my part. I meant it's not a Helm template file, so shouldn't be treated as such. I'm aware it's part of the Helm specification.

PR #934, opened 2021-04-29
Fix logs view css + webpack 

lstocchi:
Based on your comment https://github.com/Azure/vscode-kubernetes-tools/pull/933#issuecomment-828946641 I added everything in one PR but i actually just copied your stuff for the package.json. It worked fine.
The missing piece was the broken css. Still not sure why it stops working after packaging it. I'll investigate on it when i have time but this should fix our issue for the moment

### DIFF
--- a/package.json
+++ b/package.json
@@ -1209,9 +1209,12 @@
         "debuggers": []
     },
     "scripts": {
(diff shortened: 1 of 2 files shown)


itowlson (approved):
Looks great - thanks!

PR #937, opened 2021-04-30
Fix Ingress snippet indentation

markszente:
Presently, the Ingress snippet produces invalid YAML. This PR fixes the issue by correcting the indentation.


itowlson (approved):
Thanks for spotting this!

Issue #543, opened 2019-05-12
Settings for running the extension in a remote environment


radu-matei:
With the new [VS Code remote capabilities](https://code.visualstudio.com/docs/remote/remote-overview), users can set certain extensions to run either locally, or in the remote context (WSL, a local container, or a remote SSH machine).

This can be done by adding the following in the user settings:

```
    "remote.extensionKind": {
        "ms-kubernetes-tools.vscode-kubernetes-tools": "workspace/ui",
    }
```

Setting the extension to run in the remote environment (`workspace`) and then  restarting VS Code leads to the extension correctly detects the missing tools, but the automatic dependency install feature doesn't correctly work - offering to install the missing tools in the local environment, rather than in the remote one. 

This can be overridden by manually getting the dependencies and setting their paths in _an additional `settings.json` file_ (one exists for each environment).

The extension should detect when running in a remote environment and adjust all the paths - see [this document for more info about extensions using the remote environment](https://code.visualstudio.com/api/advanced-topics/remote-extensions).

BernierCR:
Hey first thanks for the post. 

However, it's been 8 months. Shouldn't someone have fixed this by now? This bug steals several minutes of productivity every day.

Also, it isn't working. I tried the first part, and the extension became grayed out. There is no Kubernetes icon on the toolbar anymore. I didn't try setting the paths yet, as with no UI, no need for the paths really. 

Most developers are going to use VSCode with WSL. WSL needs to be a top priority. This extension does not work on WSL right now. You are shipping the extension essentially broken for a large customer segment. I do all my kubernetes stuff without the extension because it won't work. 

Any help please? Thanks you very much.

squillace:
Hi @BernierCR, we'll have another look. BTW, most developers using VS Code are not on Windows, but we'd love it to work everywhere.

itowlson:
@BernierCR Sorry you are having a frustrating time with this.  I just opened one of my test folders using the "Reopen Folder in WSL" command and most things seem to work okay (using the marketplace version of the k8s extension).

There is a one-off problem if you have explicit Windows binary paths in your VS Code settings file.  This can be fixed by changing them to apply on Windows only:

```
"vs-kubernetes.kubectl-path.windows": "C:\\Users\\itowlson\\.vs-kubernetes\\tools\\kubectl\\windows-amd64\\kubectl.exe"
                           ^^^^^^^^
```

[...]

PR #944, opened 2021-05-07
Read the install locations from User and default configuration to decide whether the dependencies are installed or not

rakeshvanga:
**Description:**
This PR fixes the issue where the dependencies(kubectl, helm, minikube) are not being downloaded in remote-containers automatically. Currently in #543, manual workaround are specified to unblock the users but this PR would remove the need to have a manual workarounds.
**How it is being fixed:**
Once the install dependencies button is clicked, on a fresh machine the binaries are being downloaded to /home/username/.vs-kubernetes/tools/< toolname >/< toolBinary >. After the download is completed (for example in installKubectl ), the path of the tool is updated in the config values at Global, Workspace & WorkspaceFolder level using `addPathToConfig`.
Previously the values are being retrieved from the custom config values.

### DIFF
--- a/src/components/config/config.ts
+++ b/src/components/config/config.ts
@@ -129,7 +129,11 @@ export function getToolPath(_host: Host, shell: Shell, tool: string): string | u
 
     const baseKey = toolPathNewBaseKey(tool);
     const osKey = osOverrideKey(os, baseKey);
-    const topLevelToolPath = config.get<string>(osKey) || config.get<string>(baseKey);
+    const topLevelToolPath = 
+        userValues[osKey] ||
+        defaultValues[osKey] ||
+        userValues[baseKey] ||
+        defaultValues[baseKey];
 
     return topLevelToolPath || globalBackCompatSetting;
 }


squillace:
Luca, you want to look, or not?

lstocchi:
@squillace sure :+1:

rakeshvanga:
> LGTM, tested and works fine. Thanks!!

@lstocchi Thanks for reviewing the PR. I'm good with the changes. Can you merge these changes?

Tatsinnit:
💡thought share: just adding small idea to double check, but I hope the binary issue and file permission issue will not arise in these 2 new location. I think it will not, but worth to make sure: #922 and #850 , thanks heaps 🙏

rakeshvanga:
> 💡thought share: just adding small idea to double check, but I hope the binary issue and file permission issue will not arise in these 2 new location. I think it will not, but worth to make sure: #922 and #850 , thanks heaps 🙏

@Tatsinnit, Thanks for pointing out the other scenarios. This change will not affect the logic introduced in the PRs you have shared. It will continue to set the right permissions for the executables.

PR #949, opened 2021-05-13
Improve performance log viewer

### LINKED ISSUE #942: Log view filling up very slow for larger logfiles
Hi everyone,

first of all thank you for providing such an improved log exploration experience with the new release. 

I have some issues while digging into "large" logfiles (>2MB). If I try to follow that file from the beginning, it takes ages until the log is loaded and continues to follow the log. 

If I use the tail option for the last 500 lines, scrolling is nearly instant. 
Filtered output is a bit faster.

But if the logfile gets larger the ingest of that logfile seems to be much slower and it takes 10 minutes until the log is read. The same operation takes a few seconds on the normal terminal.

Are there any settings in vscode to improve that behavior?
Thanks for your efforts

lstocchi:
it fixes #942 

This patch fixes some performance issues discovered while reading big logs (>80k rows, size 10MB).
The content loaded is now calculated based on where the user is reading and only a part of the full content is displayed (~500rows). This prevent the UI from freezing while highlighting all the content.

After the user stops scrolling the viewer calculate which rows should be displayed and render them.

All schemas have been merged based on their final highlighting color to reduce their numbers (and so the loops), switched from inline-style to css classes and replaced arrays with objects.

### DIFF
--- a/log-colors-rules.json
+++ b/log-colors-rules.json
@@ -1,5 +1,9 @@
 [
     {
+        "regex": "\"[^\"]*\"|(?<![\\w])'[^']*'|\\[(warning|warn|wrn|wn|w)\\]|(?<=^[\\s\\d\\p]*)\\bW\\b|\\b(WARNING|WARN|Warn|WW)\\b",
+        "color": "ol"
+    },
+    {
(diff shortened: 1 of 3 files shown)


itowlson (approved):
Looks good - thanks!

PR #963, opened 2021-06-30
security fixes npm part

ivolzhevbt:
Hi maintainers! I'm vscode-kubernetes-tools and I work at [Bison Trails](https://bisontrails.co/). We're doing an internal hackathon this week.

Our team loves vscode-kubernetes-tools, but it got flagged during a recent security review. Our team voted to use our hackathon time to patch vulnerabilities in our favorite OSS tools and specifically to try to get vscode-kubernetes-tools approved for usage within our organization. 

## What's here:
- updated package version which are free from vulnerability
- dependencies forced to vulnerability-free versions when intermediate dependencies are not patched

## Impact: 
These changes should improve vscode-kubernetes-tools's security posture. We identified these vulnerabilities using [Salus](https://github.com/coinbase/salus) and the [OSSF scorecard](https://github.com/ossf/scorecard). The specific issues I am hoping to address here are:

| advisory | package | vulnerability | severity |
| --- | --- | --- | --- |
| [565](https://npmjs.com/advisories/565) | ssri | Regex DoS | mod |
| [1179](https://npmjs.com/advisories/1179) | minimist | Prototype Pollution | low | 
| [1500](https://npmjs.com/advisories/1500) | yargs-parser | Prototype Pollution | low |  
| [1589](https://npmjs.com/advisories/1589) | ini | Prototype Pollution | low  |
| [1654](https://npmjs.com/advisories/1654) | y18n | Prototype Pollution | high | 
| [1673](https://npmjs.com/advisories/1673) | lodash | Command Injectition | high | 
| [1674](https://npmjs.com/advisories/1674) | underscore | Arbitrary Code execution | high |
| [1677](https://npmjs.com/advisories/1677) | hosted-git-info | Regex DoS | mod |
| [1678](https://npmjs.com/advisories/1678) | url-parse | Path traversal | high |
| [1748](https://npmjs.com/advisories/1748) | ws | Regex DoS | mod |
| [1751](https://npmjs.com/advisories/1751) | glob-parent | Regex DoS | mod |
| [1754](https://npmjs.com/advisories/1754) | css-what | DoS | high |
| [1755](https://npmjs.com/advisories/1755) | normalize-url | Regex DoS | high | 

Thank you so much for your work maintaining vscode-kubernetes-tools, and thank you for taking the time to look at these proposed changes!

PR #964, opened 2021-06-30
security fixes go part

ivolzhevbt:
continuation of #963 

this part switches to go modules and fixes 

| vulnerability | severity | description |
| --- | --- | --- |
| [G304](https://cwe.mitre.org/data/definitions/22.html) | MEDIUM | While input is not controlled by user input it's an equal effort to sanitize input or suppress warning I went with sanitizing input |
| [G703](https://cwe.mitre.org/data/definitions/703.html) | MEDIUM | The warning is about uncatched error from deferring closing a file. While nothing can be done if this will error, it's clearer to explicitly ignore it |

Thank you for the maintaining vscode-kubernetes-tools and please let me know what do you think about it?


buildsnippets/snippets.go
```diff
@@ -97,11 +97,13 @@ var snippets = map[string]Snippet{
 
 func load(loc string) []string {
 	loc = filepath.Join("rawsnippets", loc)
-	f, err := os.Open(loc)
+	f, err := os.Open(filepath.Clean(loc))
 	if err != nil {
 		panic(err)
 	}
-	defer f.Close()
```
bacongobbler, line 104: While I understand the reasoning behind the change, this is the standard convention, and it performs exactly the same as your proposed change. This convention is described in the [Effective Go guide](https://golang.org/doc/effective_go#defer), which is referred to when styling conventions are called into question.

> This document gives tips for writing clear, idiomatic Go code. It augments the language specification, the Tour of Go, and How to Write Go Code, all of which you should read first.

I would suggest that whichever tool you're using to identify this issue should b updated to follow this convention.

bacongobbler (changes requested):

buildsnippets/snippets.go
ivolzhevbt: @bacongobbler thanks for your feedback
Salus uses gosec under the hood, and I'm quite sure they might be not very cooperative regarding this rule.
at least their recommendation https://github.com/securego/gosec/issues/512 here proposes to explicitly handle errors.

will silencing this particular warning be considered?

itowlson:
Talked to @bacongobbler and he's happy with the revised PR so will merge - thanks!

PR #967, opened 2021-07-04
Provide ability to config path to vsdbg

### LINKED ISSUE #962: Provide ability to config path to vsdbg
PR #674 added the ability to attach and debug containers with .net core inside.

Unfortunately, path to vsdbg is hardcoded in PR (`/vsdbg/vsdbg`) and differs from instructions (`~/vsdbg/vsdbg`): [vscode-omnisharp](https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes#installing-vsdbg-on-the-server), [debug instruction for extension itslef](https://github.com/Azure/vscode-kubernetes-tools/blob/master/debug-on-kubernetes.md#6-dotnet-debugging).

It will be nice to provide the ability to config path to vsdbg and change default path to debugger from `/vsdbg/vsdbg` to `~/vsdbg/vsdbg`.

pantosha:
Fixes #962

* add new option `vs-kubernetes.dotnet-vsdbg-path` to set path to vsdbg inside container
* change default path to `vsdbg` from `/vsdbg/vsdbg` to `~/vsdbg/vsdbg`

### DIFF
--- a/package.json
+++ b/package.json
@@ -234,6 +234,10 @@
                             "type": "number",
                             "description": "Remote debugging port for Python. Usually 5678."
                         },
+                        "vs-kubernetes.dotnet-source-file-map": {
+                            "type": "string",
+                            "description": "The compilation root for the vsdbg debugger in order to have a sourceFileMap in the attach configuration of debug (.NET)."
+                        },
                         "vs-kubernetes.resources-to-watch": {
                             "type": "array",
                             "description": "List of resources to be watched."
--- a/package.json
+++ b/package.json
@@ -234,6 +234,10 @@
                             "type": "number",
                             "description": "Remote debugging port for Python. Usually 5678."
                         },
+                        "vs-kubernetes.dotnet-vsdbg-path": {
+                            "type": "string",
+                            "description": "The path to the vsdbg debugger in the container (.NET)."
+                        },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -234,6 +234,14 @@
                             "type": "number",
                             "description": "Remote debugging port for Python. Usually 5678."
                         },
+                        "vs-kubernetes.dotnet-vsdbg-path": {
+                            "type": "string",
+                            "description": "The path to the vsdbg debugger in the container (.NET)."
+                        },
+                        "vs-kubernetes.dotnet-source-file-map": {
```
itowlson, line 241: Is this also new?  Does it need to be added to the defaults or docs?

Actually... it looks like it's not used at the moment, or am I missing something?

itowlson (changes requested):
Thanks for this!  Just one concern about a config entry that looks like it might be orphaned - if you can clarify that then we are good to go!

package.json
pantosha: No, this option was added in PR #736.

itowlson: Yikes!  Thanks for catching that!

itowlson (approved):

PR #969, opened 2021-07-06
Fix packaging issue

itowlson:
@ivolzhevbt This is the proposal I alluded to in my comment on #963 - it addresses all `npm audit --production` issues but does not try to remediate the `gulp` issues as they cause problems for the `vsce` packager.


ivolzhevbt:
@itowlson I believe it's the best what could be done

thanks for sorting this out.

PR #971, opened 2021-07-09
fix issue when displaying many large secrets on tree (#968)

### LINKED ISSUE #968: issue empty list for secrets child tree'is back
I would reopen https://github.com/Azure/vscode-kubernetes-tools/issues/823, but i don't have the right.
Same symptom since viscode /plugin update
image
image
image

kubectl get secret  -o wide | wc 
280 secrets found

lstocchi:
it resolves #968 

Example of performances below. Quite fast.
secrets

@teamtitangithub could you please verify it solves your issue? Thanks

### DIFF
--- a/src/kubectlUtils.ts
+++ b/src/kubectlUtils.ts
@@ -2,7 +2,7 @@ import * as vscode from "vscode";
 import { Kubectl } from "./kubectl";
 import { kubeChannel } from "./kubeChannel";
 import { sleep } from "./sleep";
-import { ObjectMeta, KubernetesCollection, DataResource, Namespace, Pod, KubernetesResource, CRD } from './kuberesources.objectmodel';
+import { ObjectMeta, KubernetesCollection, DataResource, Namespace, Pod, KubernetesResource, CRD, KeyValuePairs } from './kuberesources.objectmodel';
 import { failed, Errorable } from "./errorable";
 import { ExecResult } from "./binutilplusplus";
 import { shellMessage } from "./shell";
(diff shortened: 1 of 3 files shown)


bacongobbler:
@lstocchi would you mind explaining how the fix resolves the issue? I can't really tell how removing the `-o json` and parsing the output from `--show-labels` would fix the issue. Is it because the response from Kubernetes is significantly reduced?

src/kubectlUtils.ts
```diff
@@ -158,15 +158,54 @@ export async function deleteCluster(kubectl: Kubectl, context: KubectlContext):
...
+        };
+    }
+    return dataResource.result;
+}
+
 export async function getAsDataResources(resource: string, kubectl: Kubectl): Promise<DataResource[]> {
     const currentNS = await currentNamespace(kubectl);
 
-    const resources = await kubectl.asJson<KubernetesCollection<DataResource>>(`get ${resource} -o json --namespace=${currentNS}`);
+    const resources = await kubectl.asLines(`get ${resource} --show-labels --namespace=${currentNS}`);
```
bacongobbler, line 182: Should we explicitly set the `-o` flag here? IIRC kubernetes does not guarantee that the `--output` flag's default value will remain consistent, and could be subject to change.

bacongobbler: Also would the `--output jsonpath` be a more portable/stable option than parsing the output from `--show-labels`?

lstocchi:
> @lstocchi would you mind explaining how the fix resolves the issue? I can't really tell how removing the `-o json` and parsing the output from `--show-labels` would fix the issue. Is it because the response payload from Kubernetes is significantly reduced?

Yes, exactly. This changes the amount of data downloaded from the cluster to create the tree. By using `-o json` it downloads everything, with an amount of data which could be very large (e.g the user has 300 secrets 1 Mb each). And most of those data are useless. In this use case, who's going to expand every single secret from the tree? 
By parsing the output from `--show-labels` we have all the info we need to create all the secret nodes. When a node is selected/expanded then the plugin makes a call to request additional data.

About timing with my minikube cluster, which can be increased based on user network.
```
$ kubectl get secrets -o json

real	0m41.070s
user	0m19.586s
sys	0m1.210s

$ kubectl get secrets --show-labels

real	0m2.451s
user	0m0.205s
sys	0m0.033s
```

src/kubectlUtils.ts
lstocchi: AFAIK jsonpath is not faster than `--show-labels` and it doesn't solve our problem. I didn't dig up how `get resource --show-labels` works but i guess that this command gets the data directly while jsonpath parses all content trying to find a match with what's in the jsonpath expression. And it takes time. But i have to admit that i'm not an expert and maybe there are ways to write jsonpath expressions which perform better than those i write.

lstocchi: For example, this is the result by trying to only take the secrets names with jsonpath.
```
kubectl get secrets -o jsonpath="{range .items[*]}{.metadata.name}{end}"

real	0m21.091s
user	0m16.469s
sys	0m0.538s
```

itowlson (approved):
LGTM - thanks!

PR #984, opened 2021-10-12
start values completion whenever .Values. occurs.

flowftw:
Helm values completion didn't work the way I would have expected it to do. Typing `.Values.` or `.Values.service.` and triggering autocompletion by pressing ctrl + space didn't present any actual values. Personally I use autocompletion a lot to explore available values which the new behaviour helps with. The completion didn't show
Old behaviour:
Bildschirmfoto 2021-10-12 um 20 12 53
New behaviour:
Bildschirmfoto 2021-10-12 um 20 11 49


itowlson (approved):
LGTM - thanks!

PR #992, opened 2021-10-28
Watch for changes in helm's values.yaml

flowftw:
The PR uses the `vscode.FileSystemWatcher` API to refresh the helm values cache whenever `values.yaml` is saved. With these changes in place, it is no longer necessary to reload VSCode to include new values in autocompletion.


itowlson (approved):
Looks good to me - thanks, this is a neat feature to have!

PR #1001, opened 2021-11-11
close opened kubectl processes when extension is deactivated

lstocchi:
This PR should fix the problem that we have with "all" kubectl processes that are left opened if vscode is shutted down before those processes are closed by the user.

The only issue i wasn't able to fix and i don't know why is related to the observable that are opened when following pod logs.

What i tried to do was to use an abortController and put the abortSignal inside the shell execOptions (as described in the documentation https://nodejs.org/api/globals.html#class-abortsignal). Ideally when the signal is sent, the process should be killed. It didn't work.

So i tried another way which is less elegant, https://github.com/Azure/vscode-kubernetes-tools/pull/1001/files#diff-68da841fbea6534a636dc451cf3178e52d0d0c0beab917207420fd7127ed071bR284 , but it doesn't always work. Sometimes it doesn't close all processes. 

@itowlson Can you please try following logs of 3-4 pods at the same time, closing vscode and see the kubectl process that are not closed? Do you have any? Of course if you know the reason or have a fix feel free to edit the PR yourself. After 1 week trying it, i'm giving up for some days :laughing:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
     "name": "vscode-kubernetes-tools",
-    "version": "1.3.4",
+    "version": "1.3.7",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
-            "version": "1.3.4",
+            "version": "1.3.7",
             "license": "MIT",
             "dependencies": {
                 "@kubernetes/client-node": "^0.14.0",
@@ -205,9 +205,6 @@
             },
             "engines": {
                 "node": ">=8"
-            },
-            "funding": {
-                "url": "https://github.com/sponsors/sindresorhus"
             }
         },
         "node_modules/@electron/get/node_modules/cacheable-request/node_modules/lowercase-keys": {
@@ -229,11 +226,6 @@
             },
             "engines": {
                 "node": ">=6.0"
-            },
-            "peerDependenciesMeta": {
-                "supports-color": {
-                    "optional": true
-                }
             }
         },
         "node_modules/@electron/get/node_modules/defer-to-connect": {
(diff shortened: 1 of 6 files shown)


itowlson:
Thanks heaps for looking into this.  From what you're seeing, are we hitting the code to kill the processes?  That is, is the problem:

* We never execute the kill code; or
* The kill code doesn't kill the processes?

I'll have a tinker but if you know which of these is the issue then that would help narrow it down.

Also, do we know if this issue happens on all operating systems or only a particular one?  (Again, if you don't know offhand then I can have a play.)

lstocchi:
> Thanks heaps for looking into this. From what you're seeing, are we hitting the code to kill the processes? That is, is the problem:
> 
> * We never execute the kill code; or
> * The kill code doesn't kill the processes?
> 
> I'll have a tinker but if you know which of these is the issue then that would help narrow it down.
> 
> Also, do we know if this issue happens on all operating systems or only a particular one? (Again, if you don't know offhand then I can have a play.)

We completely missed the code to kill processes. This patch adds it but it doesn't always work. Atleast when we don't have a direct reference to a process. If we are the ones who create that process, it's easy, i kill it directly, i have its id. 
But, there are cases when we only have access to the stdout of a process because it is spawned inside a library. I tried to send an abortSignal within the execute options and, ideally, this should be good enough to tell that process to abort when i want. Sadly, it doesn't work. 

I think it is a problem related to all OSes but not 100% sure.

lstocchi:
@itowlson Removed the abort controller which wasn't working as expected. Now it should work fine. I still keep it on draft as i don't have much time to test it extensively today but you can give it a look.

lstocchi:
@itowlson tested it. It looks like all leaked kubectl processes are killed properly now. Please give it a look. Thanks!!

itowlson (approved):
This looks terrific.  Thanks so much for nutting this out.

Issue #1002, opened 2021-11-13
vsdbg default path not working


tmds:
https://github.com/Azure/vscode-kubernetes-tools/pull/967 changed the default path of vsdbg to `~/vsdbg/vsdbg`.

This doesn't work. Expanding the `~` is the responsibility of the shell, and `kubectl exec` doesn't involve a shell.

For example:
```
$ /usr/bin/kubectl exec -i dotnet60-app-76bb77ff87-gvfpx -- echo '~/vsdbg/vsdbg'
~/vsdbg/vsdbg
```
It can be fixed by using a shell that causes the `~` to be expanded:
```
$ /usr/bin/kubectl exec -i dotnet60-app-76bb77ff87-gvfpx -- /bin/sh -c 'echo ~/vsdbg/vsdbg'
/opt/app-root/vsdbg/vsdbg
```

`vsdbg` should be started using:
```
kubectl exec -i <pod> -- /bin/sh -c '<path> --interpreter=vscode'
```

cc @itowlson @pantosha

itowlson:
@pantosha Would you be able to take a look at this please?  I'm not very familiar with the .NET debugging component.  (If you don't have bandwidth let me know and I'll go with @tmds' fix.)

pantosha:
I'll take a look during current week

itowlson:
Fixed via #1004.

PR #1004, opened 2021-11-20
Use `/bin/sh` to run debugCommand in dotnet (#1002)

pantosha:
Using shell allows expanding `~`.


itowlson (approved):
Looks good, thanks!

PR #1009, opened 2021-11-25
Allow user to exclude YAML file from processing

### LINKED ISSUE #995: The extension falsely reports validation errors in Kustomization patch manifests
## Problem description

Can´t ignore / disable or change schema validation on kustomization patch files. The extension believes it's a true kubernetes manifest with all its expections, and validates with the kubernetes schema. Therefore the extension complains that the .yaml patch file doesn´t match the expectations.

## Steps to Reproduce

1. Create a **patch.yaml** file with the following content:

``` yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app
spec:
  replicas: 1
```
2. Observe the problems in VSCode
image

## Expected Behavior

Different schema to validate against, or disable/ignore certain files (like kustomization patches) altogether. 

## Current Behavior

All files get validated according to the kubernetes schema even if kustomization files are meant to not adhere to the schema.

## Relevant installed and enabled extensions

- vscode-kubernetes-tools
- redhat.vscode-yaml (uses https://github.com/redhat-developer/yaml-language-server)
- vscode-eslint
- prettier-vscode
- vscode-docker

## Proposed solutions

1. Allow the extension to be configured with **glob patterns** to ignore validation on certain directories / files. Something like this: https://github.com/redhat-developer/vscode-yaml
2. Or maybe when a `kustomization.yaml` file is detected in the folder, smartly detect the files that are kustomization patches and ignore them
3. A comment in the beginning of a manifest to let the extension know that this is a Kustomization patch and have other rules, or another schema to use. Example:
``` yaml
# kubernetes-is-kustomization-patch: true
# OR
# vscode-kubernetes-tools-ignore-validation: true
# OR something else..
apiVersion: apps/v1
kind: Deployment
```

itowlson:
Fixes #995.

@lstocchi @rille111 would you be able to give this a bit of a real world test?  I've written unit tests but due to recent shenanigans I'm not in a position to test in a real setup right now.  Thanks and sorry for being so half-baked!

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,6 +253,18 @@ Here are the various linters, you can enable or disable them individually using 
 }
 ```
 
+## YAML processing directives
+
+You can customise how the extension processes a YAML file by putting a comment of the form
+`# vscode-kubernetes-tools: ...directive(s)...` at the top of the file. The comment must appear before
+any YAML elements (this saves us having to scan the whole of YAML files that could easily
+be tens of thousands of lines!). Other comments and blank lines are okay.
+
+The following directives are supported:
+
+* `exclude`: Do not treat this file as a Kubernetes manifest. This is useful when you have
+  files that look a lot like Kubernetes manifests but aren't, such as Kustomize patch files.
+
 ## Known issues
 
   * `Kubernetes: Debug` command currently works only with Go, Node.js, Java, Python and .NET applications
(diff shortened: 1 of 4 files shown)


lstocchi:
Adding the YAML master to verify this @evidolob

lstocchi (approved):
LGTM

evidolob:
This works very good! Thx @itowlson.

lstocchi:
@itowlson we should add a section in the README

itowlson:
Good catch @lstocchi - thanks and done!

PR #1019, opened 2021-12-21
add support for arm64 tools installation for windows/darwin (#950)

### LINKED ISSUE #950: Extension downloads wrong binary on Apple M1
Extension downloads wrong minikube binary. After the extension update, I have to manually download minikube binary from GitHub, rename it, make it executable and move it to the proper folder.

lstocchi:
it resolves #950


itowlson (approved):
This looks plausible to me, so good to go as soon as we can confirm that it works on real life Macs!

PR #1027, opened 2022-01-11
Do not log values refresh message

### LINKED ISSUE #1026: Turn off Helm "refresh values cache" output
Greetings,

I'm using the Kubernetes tools extension, and every time I save a `values.yaml` file, the following prompt always comes up:

helm-values-refresh

It's a bit annoying, and I don't need to know every time this happens.

Is this a simple entry in the `settings.json` file?

Thank you!

flowftw:
Do not write message to VSCode Helm Output every time Helm values have been refreshed. Fixes #1026.


itowlson (approved):
Sweet, thanks!

Issue #1003, opened 2021-11-15
.NET process filtering is too restrictive


tmds:
Currently, any process that is not named `dotnet` is filtered out:

https://github.com/Azure/vscode-kubernetes-tools/blob/268260093d0bd3265d981fb571261d3c4818c436/src/debug/dotNetDebugProvider.ts#L81-L84

Since .NET 3.1 (released 2 years back), .NET applications come with a native executable that launches the app. These apphost are not named `dotnet`.

```
$ dotnet new web -o web
$ cd web
$ dotnet build
$ bin/Debug/net6.0/web 
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5000
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: https://localhost:5001
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Production
info: Microsoft.Hosting.Lifetime[0]
      Content root path: /tmp/web
```

A way of filtering .NET apps is to look at the `/tmp/dotnet-diagnostic-<pid>-*-socket` files.
```
$ pidof web
16873
$ ls /tmp/dotnet-diagnostic-$(pidof web)-*-socket
/tmp/dotnet-diagnostic-16873-1204654-socket
```

Alternatively, the filtering could be left out, since there are not that many processes expected to be running in a container anyhow.

tmds:
cc @itowlson

itowlson:
Hmm, I'm not too familiar with the .NET debugging component.  @tedchamb you were the original author of this feature - are you able to take a look please?

itowlson:
@pantosha Sorry to pester you, but the original author of this code isn't responding right now, and I know you have some knowledge of this area.  Would you be able to take a look at this issue too?  Thanks ever so much!

pantosha:
@itowlson,  I will take a look

tmds:
> A way of filtering .NET apps is to look at the /tmp/dotnet-diagnostic-<pid>-*-socket files.

@pantosha this is a good way to filter .NET apps. Though it has some complexity to it. Alternatives are to do no filtering, or another alternative would be to allow specifying the process name with a setting (which defaults to `dotnet`).

tmds:
@pantosha are you still looking into the issue?

[...]

PR #1029, opened 2022-01-15
Shift dotnet process identification to vscode-omnisharp (#1003)

pantosha:
Previous dotnet debugging implementation looked for processes with
dotnet name only. Since .NET Core 3.1, .NET applications may have a native
executable to launch the app with name that differs from dotnet.

The new implementation doesn't set process id when call vscode-omnisharp
debugger. Now, vscode-omnisharp is used for executing `ps` and showing
vscode UI for choosing process.

### DIFF
--- a/debug-on-kubernetes.md
+++ b/debug-on-kubernetes.md
@@ -128,9 +128,7 @@ https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes
 
    2. Notes:
       * `dotnet` debugging is only supported for the `attach` scenario. The `launch` scenario is not yet supported.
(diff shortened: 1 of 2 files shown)


itowlson (changes requested):
Looks good - thanks!

Unfortunately I have to ask for one change please.  Because we are now part of the CNCF, we have switched from the old CLAs to DCOs, which require commits to be signed.  Could you re-commit and re-push with a signature please?  (See the "failed DCO" check on the main PR page for more info.)

pantosha:
> Looks good - thanks!
> 
> Unfortunately I have to ask for one change please. Because we are now part of the CNCF, we have switched from the old CLAs to DCOs, which require commits to be signed. Could you re-commit and re-push with a signature please? (See the "failed DCO" check on the main PR page for more info.)

Finally, done

itowlson (approved):
Awesome, thanks!

PR #1030, opened 2022-01-17
Allow user to suppress warning if no kubectl binary

### LINKED ISSUE #1023: Add "Do not show again" option to "Kubectl not found. Many features of the Kubernetes extension will not work." notification
I am using the Kubernetes extension on a system that does not have kubectl installed. I am not using it for that functionality, but rather for the helm capabilities of the extension.

Each time I open a yaml file, I get two notifications about missing kubectl:
image

The menu of the notification (gear icon) only has a `Manage extension` button. It would be great if there was a `Do not show again` button so that I can disable them.

itowlson:
Fixes #1023.


lstocchi (approved):
LGTM

PR #1056, opened 2022-04-08
Use custom setting to silence extension recommendation notifications (#1053)

### LINKED ISSUE #1053: Extension asks to search for CRD-based suggestions but ignores "Don't ask again" answer.
I believe the title says it all.
I've been looking for a setting to change this behaviour, alas.
It's really really annoying to have to watch "Error from server" every time one opens a workspace or adds a folder and there's nothing one can do about it.

lstocchi:
it resolves #1053


itowlson (approved):
Looks good - thanks Luca!

Issue #1058, opened 2022-04-14
Logs should 'Follow Logs' and 'Run' by default


juozasg:
For logs display, a small change would significantly improve UX to be able to click Logs on a pod and see the logs immediately, without having first to click 'Follow Logs' and then 'Run'. And it should scroll to the bottom:

![image](https://user-images.githubusercontent.com/4100/163392429-3febb9a1-a51d-4fdc-b395-75d1aac2d591.png)

kingdonb:
It looks like there has actually been progress on this issue upstream at:

* https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1062

kingdonb:
I had a try at https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1066 and found it makes this setting possible, you can change "Follow Logs" to be the default, and autorun has been added as an option that you can enable. We may have to call this out in our docs in order to make it more fully accessible, I'm not sure if they'll make it default behavior or if there are any drawbacks, but it worked for me!

Another open question is whether we could enable it by default, when we install the Kubernetes extension as a dependency? IDK 🤷

lstocchi:
Closing as #1066 has been merged long ago.

PR #1062, opened 2022-04-28
feat: add log viewer options

SethFalco:
Whenever I view the logs of a pod, I always set the same options pretty much. ^-^'
This just creates some global settings which sets the default log viewer options.

For now, it only covers `follow`, `timestamp` and `wrap`.
I'm considering doing others later, but figured for now this would be nice.

It looks to me (though I didn't check manually) that the `reset()` method wasn't resetting the `wrap` checkbox anyway, so this also fixes that.

### Related
* Partially resolves x https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1058#issue-1204480246
  * I can also make an option to run by default on viewing logs too after/if this is merged.

### DIFF
--- a/src/components/logs/logsWebview.ts
+++ b/src/components/logs/logsWebview.ts
@@ -6,6 +6,7 @@ import { fs } from '../../fs';
 import { Container } from '../../kuberesources.objectmodel';
 import { Kubectl } from '../../kubectl';
 import { getLogsForContainer, LogsDisplayMode } from '../kubectl/logs';
+import { isLogViewerFollowEnabled, isLogViewerTimestampEnabled, isLogViewerWrapEnabled } from '../config/config';
 
 export class LogsPanel extends WebPanel {
     public static readonly viewType = 'vscodeKubernetesLogs';
@@ -141,10 +142,9 @@ export class LogsPanel extends WebPanel {
     }
 
     private getLogViewerOptions(): object {
-        const configuration = vscode.workspace.getConfiguration();
-        const follow = configuration.get('vscode-kubernetes.log-viewer.follow', false);
-        const timestamp = configuration.get('vscode-kubernetes.log-viewer.timestamp', false);
-        const wrap = configuration.get('vscode-kubernetes.log-viewer.wrap', false);
+        const follow = isLogViewerFollowEnabled();
+        const timestamp = isLogViewerTimestampEnabled();
+        const wrap = isLogViewerWrapEnabled();
 
         return {
             follow,
--- a/src/components/logs/logsWebview.ts
+++ b/src/components/logs/logsWebview.ts
@@ -6,7 +6,7 @@ import { fs } from '../../fs';
 import { Container } from '../../kuberesources.objectmodel';
 import { Kubectl } from '../../kubectl';
 import { getLogsForContainer, LogsDisplayMode } from '../kubectl/logs';
-import { isLogViewerFollowEnabled, isLogViewerTimestampEnabled, isLogViewerWrapEnabled } from '../config/config';
+import { isLogViewerFollowEnabled, setLogViewerFollowEnabled, isLogViewerTimestampEnabled, setLogViewerTimestampEnabled, isLogViewerWrapEnabled, setLogViewerWrapEnabled } from '../config/config';
 
 export class LogsPanel extends WebPanel {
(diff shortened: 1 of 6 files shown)


SethFalco:
While we're at it, I'm just wondering about this. 🤔 

```js
// eslint-disable-next-line @typescript-eslint/prefer-for-of
for (let i = 0; i < newContent.length; i += 1) {
    if (newContent[i].length > 0) {
        content[counter] = newContent[i];
        fpcCounter++;
        fullPageContent[fpcCounter] = newContent[i];
        counter++;
    }
}
```

Is there a reason you're not using a for-each? I normally would change this without asking, but given in the repo the warning it explicitly disabled too, I wanted to ask about it.

src/components/logs/logsWebview.ts
```diff
@@ -126,4 +140,16 @@ export class LogsPanel extends WebPanel {
             .replace('<!-- meta http-equiv="Content-Security-Policy" -->', meta);
     }
 
+    private getLogViewerOptions(): object {
+        const configuration = vscode.workspace.getConfiguration();
+        const follow = configuration.get('vscode-kubernetes.log-viewer.follow', false);
```
lstocchi, line 145: Better to move these to `config.ts` for simmetry and so that config remains responsible for knowing the key names.

lstocchi (commented):
Thanks for this @SethFalco. Looks good. 
I'm wondering if we should add a button or anything else to save the settings directly from the viewer. It would be faster for users to set them and also it would be easily seen by those that do not read the README/changelog. @itowlson WDYT?

lstocchi:
> While we're at it, I'm just wondering about this. thinking
> 
> ```js
> // eslint-disable-next-line @typescript-eslint/prefer-for-of
> for (let i = 0; i < newContent.length; i += 1) {
>     if (newContent[i].length > 0) {
>         content[counter] = newContent[i];
>         fpcCounter++;
>         fullPageContent[fpcCounter] = newContent[i];
>         counter++;
>     }
> }
> ```
> 
> Is there a reason you're not using a for-each? I normally would change this without asking prior, but given in the repo the warning is explicitly disabled too, I wanted to ask about it.

The log viewer has a long and troubled history as it changed quite a bit lately. I guess this was made during some performance enhancement PRs. First we showed all logs in a single push but it was slow when highlighting all words based on a list of regexes and the logs were > 10MB. So we tried to optimize every thing we could. Using objects instead of arrays, pagination, reduce the number of regexes. I guess I also switched to the old style for which could make me earn some ms of execution. Maybe after pagination and all other changes this is not necessary anymore but i didn't look at it anymore honestly.

SethFalco:
* Moved the config to the centralized config.ts.
* Changed some wording for consistency. (Trying to use the term "Settings" everywhere rather than switch between "Options" or "Settings".
* Add a "Save Settings" button in the webview which saves the 3 settings.

Note, for now this is only allowing `follow`, `timestamp`, and `wrap` to be configured and saved. However, I'll happily do the other settings too, plus the "Run" by default.

Maybe I can do those in a separate PR, though?

I'd like if we can just establish if we're happy with the structure of this first, then I can just follow-through for the rest.

SethFalco:
image

Ignore the other default values like `0` or `seconds`. (That's for a separate PR I was working on. Just subjective usability improvement and refactors, but that's out of scope for this PR.)

SethFalco:
> @SethFalco could you please rebase it? Thanks!!

Rebased and conflicts resolved. 👍🏾

SethFalco:
Hmm, the failed test seems odd. 🤔 
Could we maybe do a retry to see if that was a random timeout first?

itowlson:
It cure looks like a random timeout - I've re-started the actions.

itowlson (approved):
LGTM - thanks!

SethFalco:
Ahh, sorry about that.
Just a few minutes ago I was looking at this and wanted to tweak `setSettings` a bit, so I force-pushed there anyway. ^-^'

Thought for future, it'd be better if it took an options object rather than a bunch of parameters.

That way in future to set particular settings we could just do:

```js
setSettings({
  sinceSelect: 1
});
```

Instead of something like:
```js
setSettings(null, null, null, 1);
```

itowlson (approved):

lstocchi (approved):
Wonderful, thanks!! LGTM

PR #1063, opened 2022-04-29
chores: refactor methods, edit configs, fix formatting, etc

SethFalco:
### Functional
This list is anything that actually changes the end-user-experience with the extension.

* In the log viewer, changes the default option for `since-input` to `-1` and `since-select` to `s`. (This stills results in `getSinceDuration()` returning `0` by default because of `sinceInput <= 0`.)
* Removes the empty string ('') option for `since-select`. (Now users only have to complete one field instead of 2 if they want to get log history i.e. `select-since` of `3600` for past hour. They can still do both if they want to, i.e. `1` and `hour`.)
* Changes the input type of `since-input` and `tail-input` to `number`. (So now users can't type arbitrary text in them.

I won't get to benefit from the above ideally, since my other PR(s) will add settings to configure default values by default anyway which I intend to use.

However, in general I think this will make for a better initial developer experience with the extension.

### Refactors/Chores
This list is just random other stuff, ideally it shouldn't have any impact on functionality of the extension.

* In the Log Viewer HTML, replace all single quotes for HTML attributes with double quotes. (Also trimmed whitespace.)
* In `.vscode/extensions.json` remove `dbaeumer.vscode-eslint` as this is no longer available on the marketplace.
* Trim trailing whitespace on Debugging on Kubernetes.
* NPM automatically modified the `package-lock.json`, I just went with it.
* Various refactors in the JavaScript to be subjectively cleaner.

---

You're welcome to tell me to undo any particular change and I'm happy to do it.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,6 @@
     "recommendations": [
         "redhat.vscode-yaml",
         "dbaeumer.vscode-eslint",
-        "eamodio.tsl-problem-matcher"
+        "amodio.tsl-problem-matcher"
     ]
 }
--- a/debug-on-kubernetes.md
+++ b/debug-on-kubernetes.md
@@ -15,7 +15,7 @@ One of the key features of VS Code Kubernetes Extension is its one-click debuggi
 
       ### Miscellaneous commands
       * `Kubernetes: Select Pod` - It allows to select a pod from a list of pods belonging to the "current" namespace. It can be invoked through command substitution (`${command:extension.vsKubernetesSelectPod}`) inside files such as launch.json or tasks.json.
-   
+
          An example of how to use the command inside a launch.json configuration can be found below.
 
 ```javascript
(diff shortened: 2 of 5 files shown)


.vscode/extensions.json
```diff
@@ -1,7 +1,6 @@
 {
     "recommendations": [
         "redhat.vscode-yaml",
-        "dbaeumer.vscode-eslint",
-        "eamodio.tsl-problem-matcher"
```
lstocchi, line 5: I think it has been updated to `amodio.tsl-problem-matcher`? @itowlson do you want to update it or remove it?

lstocchi (approved):
LGTM, not sure about the recommendation extension.

.vscode/extensions.json
SethFalco: Ahh, you're right!
I only looked checked the repo and didn't see any Marketplace links, so I assumed it was dead.

Now that I think about it, last commit being in April 2022 should've given it away. ^-^'

They should really add a link on the README. ^-^'

For now, I can add it back with the new name. If later it's requested to remove, I can just rebase to remove it again.

PR #1066, opened 2022-05-06
feat: add option to run automatically

SethFalco:
Adds a setting to automatically run the log viewer using the default settings once the viewer has finished initializing.

### Related
* Partially solves x https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1058
  * Still missing a setting to scroll to bottom by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ Minikube tools to be installed and available on your PATH.
    * `vscode-kubernetes.log-viewer.follow` - Set to true to follow logs by default in the log viewer.
    * `vscode-kubernetes.log-viewer.timestamp` - Set to true to show timestamps by default in the log viewer.
    * `vscode-kubernetes.log-viewer.wrap` - Set to true to wrap lines by default in the log viewer.
+   * `vscode-kubernetes.log-viewer.autorun` - Set to true to automatically begin fetching logs once the log viewer is opened using the default settings.
 
 ## Custom tool locations
 
--- a/package.json
+++ b/package.json
@@ -396,6 +396,11 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Set to true to wrap lines by default in the log viewer."
+                },
+                "vscode-kubernetes.log-viewer.autorun": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Set to true to automatically begin fetching logs once the log viewer is opened using the default settings."
                 }
             }
         },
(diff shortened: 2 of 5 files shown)


kingdonb:
This is great, exactly what I was looking for:
Screen Shot 2022-05-06 at 1 18 00 PM

I would make these the defaults:
Screen Shot 2022-05-06 at 1 18 46 PM

...but even merging as-is would be a marked improvement, as these settings are made significantly more accessible by the addition of the autorun setting 👍 

Are there any performance concerns? I haven't quite worked out performance profiling myself in VScode extensions yet, but it doesn't seem to be using any CPU at idle, which is good, that's what I'd expect 😉

SethFalco:
> I would make these the defaults:

For making them default, we'll see what the maintainers think.

I think `follow` would be a good default because (and this is an assumption) that the average user is probably seeking future logs if they're connecting via the extensions log viewer.

Logs from the past are more frequently accessed via platforms like Sentry or Loggly. (To diagnose an error, for example.)

> Are there any performance concerns?

(Technically still replying to the default settings thing, but it's mostly linked with performance so putting it under this.)

Autorun by default is an iffy one in my opinion. (But I don't have authoritative say, so we'll see what the maintainers think.)

It might seem like a sensible default, especially now that all the other settings are configurable. (Well not all now, but soon it will be all.)

Imagine a developer's first experience with the extension, assuming autorun is enabled by default, if a pod with hundreds of useless logs going through on every incoming request is viewed. Awful instability with the code editor because the user didn't have time to apply other filters to limit to a certain time or to contain specific text.

I agree the option should exist, and I intend to have it enabled myself, but I'm wary of the situation above making the initial experience with the extension shitty and proceeding to uninstall it.

kingdonb:
> Imagine a developer's first experience with the extension, assuming autorun is enabled by default, is by viewing a pod with hundreds of useless logs going through on every incoming request

You're doing great work, attention to these concerns will make sure this extension stays useful and performant.

I'm convinced, with the `Save Settings` shortcut to update the preferences, I think it's perfectly accessible now, and I will include a page in the docs for our extension that highlights how you can set these optimally, to follow logs.

lstocchi (approved):
LGTM

I don't have a strong opinion about default values. I think these can be left as they are and then update them later based on community feedback if necessary, but everyone has different needs so it would be hard to guess the best initial settings.

Only thing I would improve would be the visibility of this auto-run setting. If i'll see many people opening issues about the need of this auto-run, i'll make it more prominent with a checkbox in the editor toolbar. For now, let's keep it in the documentation.

Thanks again @SethFalco :pray:

PR #1059, opened 2022-04-15
add setting to disable fetching all crds

lstocchi:
it resolves #1050 

This adds a setting to allow users enabling/disabling the crds fetch when the plugin tries to download all custom schemas. If the general setting is undefined the plugin uses the crds count to decide if to proceed or not and ask the user for confirmation. @itowlson

PR #1069, opened 2022-05-10
Fix unresponsive UI after getting CRD json

peterbom:
This addresses a problem that happens:
- on Windows
- on a cluster with a large number of CRDs (~80 in my case)

**Symptom**

The UI becomes completely unresponsive for well over a minute, in a couple of places:
- immediately *after* the new (#1059) CRD popup appears (making it impossible to make the choice about fetching CRDs until it's recovered).
- after expanding the 'Custom Resources' item.

These are both times when the code invokes `kubectl get crd -o json`.

Interestingly, nothing much seems to be happening at the time it hangs. Both the `kubectl` call and the JSON parsing have already completed. It appears to happen on the first `await` after those things have happened, so my best guess is that parsing the massive json array created a lot of detritus in memory, and the unresponsiveness is down to the garbage collector kicking in.

**Fix**

Whatever the cause, the fix I'm proposing here is to minimize the length of stdout produced by the `kubectl` calls, which seems to make the problem go away.

### DIFF
--- a/src/components/clusterexplorer/node.folder.crdtypes.ts
+++ b/src/components/clusterexplorer/node.folder.crdtypes.ts
@@ -2,7 +2,6 @@ import { Kubectl } from '../../kubectl';
 import * as kubectlUtils from '../../kubectlUtils';
 import { Host } from '../../host';
 import * as kuberesources from '../../kuberesources';
-import { CRD } from '../../kuberesources.objectmodel';
 import { ClusterExplorerNode } from './node';
 import { GroupingFolderNode } from './node.folder.grouping';
 import { ResourceFolderNode } from './node.folder.resource';
@@ -13,13 +12,6 @@ export class CRDTypesFolderNode extends GroupingFolderNode {
     }
     async getChildren(kubectl: Kubectl, _host: Host): Promise<ClusterExplorerNode[]> {
         const objects = await kubectlUtils.getCRDTypes(kubectl);
-        return objects.map((obj) => ResourceFolderNode.create(this.customResourceKind(obj)));
-    }
-    private customResourceKind(crd: CRD): kuberesources.ResourceKind {
-        return new kuberesources.ResourceKind(crd.spec.names.singular, crd.spec.names.plural, crd.spec.names.kind, this.safeAbbreviation(crd), crd.spec.names.plural);
-    }
-    private safeAbbreviation(crd: CRD): string {
-        const shortNames = crd.spec.names.shortNames;
-        return (shortNames && shortNames.length > 0) ? shortNames[0] : crd.metadata.name;
+        return objects.map(ResourceFolderNode.create);
     }
 }
--- a/src/kubectlUtils.ts
+++ b/src/kubectlUtils.ts
@@ -238,7 +238,7 @@ export async function getCRDTypesNumber(kubectl: Kubectl): Promise<Errorable<num
 }
 
 export async function getCRDTypes(kubectl: Kubectl): Promise<ResourceKind[]> {
-    const crds = await kubectl.asLines(`get crd -o jsonpath="{range .items[*]}{.metadata.name}{\\" \\"}{.spec.names.kind}{\\" \\"}{.spec.names.singular}{\\" \\"}{.spec.names.plural}{\\" \\"}{.spec.names.shortNames}{\\"\\n\\"}{end}"`);
+    const crds = await kubectl.asLines(`get crd -o jsonpath="{range .items[*]}{.metadata.name}{\\" \\"}{.spec.names.kind}{\\" \\"}{.spec.names.singular}{\\" \\"}{.spec.names.plural}{\\" \\"}{range .spec.names.shortNames}{@}{end}{\\"\\n\\"}{end}"`);
     if (failed(crds)) {
         vscode.window.showErrorMessage(crds.error[0]);
         return [];
--- a/src/kubectlUtils.ts
+++ b/src/kubectlUtils.ts
@@ -223,42 +223,41 @@ export async function getGlobalResources(kubectl: Kubectl, resource: string): Pr
 }
 
 export async function getCRDTypesNumber(kubectl: Kubectl): Promise<Errorable<number>> {
-    const crdTypes = await kubectl.asLines(`get crd -o name`);
-    if (failed(crdTypes)) {
-        return {
-            succeeded: false,
-            error: crdTypes.error
-        };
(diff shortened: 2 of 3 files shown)


lstocchi:
There are 2 issues i think. The first is that asLines expects a query which contains a header with newline which is removed after the command is executed https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/4183b84a9996b2f6a8158b7c97038ab5bd94a6bc/src/kubectl.ts#L461 
The second is that the newline is not added in the output of the command at the end of each row so the split fails. I guess the `@` in the jsonpath expression gives some trouble.

peterbom:
> There are 2 issues i think. The first is that asLines expects a query which contains a header with newline which is removed after the command is executed
> 
> https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/4183b84a9996b2f6a8158b7c97038ab5bd94a6bc/src/kubectl.ts#L461
> 
> 
> The second is that the newline is not added in the output of the command at the end of each row so the split fails. I guess the `@` in the jsonpath expression gives some trouble.

Thank you for taking the time to dig into this, @lstocchi!

Oh dear, I did have a bit of trouble with those newlines; I experienced the same issue as you on WSL before my second commit, which I *thought* fixed it, but it sounds as if the jsonpath behaviour for nullable array properties varies between `kubectl` versions.

Where is `abbreviation` in the `ResourceKind` type used? Could I rewrite the command to skip the `shortNames` retrieval altogether and construct it using `metadata.name` for the abbreviation (which it falls back to in the majority of cases anyway)?

Great catch on `asLines` dropping the first line! Two possible ideas:
- Add an `options` parameter to `asLines` containing something like `skipFirstLine` (defaulting to `true`)
- use `invokeCommand` directly and split the lines at the calling location

Any preference, or other suggestions? (I notice there's a `TODO: analyse uses for these and convert these over` above `asLines`, implying maybe it's considered for deprecation?)

lstocchi:
My suggestion would be to just take the first element in the list of shortnames. In any case we are going to do it here https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1069/files#diff-2fce73bf2e0e24b5ece741097faca60c7e4d0e3822b81c993ea0e7ae8a6fbf1cR260
so we could rewrite the jsonpath to do it by default :thinking: 

Something like 
```
get crd -o jsonpath="........{\\" \\"}{.spec.names.plural}{\\" \\"}{.spec.names.shortNames[0]}...
```

So we skip the issue with the evil `@` which unfortunately idk how to make the `asLines` method digest it.
At that point your code should be good, just some small changes like the `getSafeAbbreviation` -> `shortname ? shortname : metadataName` and the skipFirstLine in the asLines method.

The `abbreviation` is used in several places and i'm worried we could cause some issue by changing it to metadataName by default. It could also be a pain to verify that everything keeps working fine. 

You can also use `invokeCommand` instead of asLines, up to you.

`TODO: analyse uses for these and convert these over` this was made by the boss @itowlson . I don't remember what was the plan honestly. Maybe you're right and he wanted to deprecate it sooner or later.

Hope i didn't miss anything :smile: 
Thanks again for your help :pray:

itowlson:
That TODO, if it’s what I think it is (sorry away from computer), refers to these variants being old ones which can be noisier and more side-effectful than the new ones, but may be used by existing code which would need updates to error handling and recovery if changed to the new variants.  In particular, the things I’m thinking of are used by the API, and we can’t change the behaviour of that without creating a new major version.

New code should definitely use the new variants.  I apologise for the mess!

peterbom:
> My suggestion would be to just take the first element in the list of shortnames. In any case we are going to do it here https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1069/files#diff-2fce73bf2e0e24b5ece741097faca60c7e4d0e3822b81c993ea0e7ae8a6fbf1cR260 so we could rewrite the jsonpath to do it by default 🤔
> 
> Something like
> 
> ```
> get crd -o jsonpath="........{\\" \\"}{.spec.names.plural}{\\" \\"}{.spec.names.shortNames[0]}...
> ```
> 
> So we skip the issue with the evil `@` which unfortunately idk how to make the `asLines` method digest it. At that point your code should be good, just some small changes like the `getSafeAbbreviation` -> `shortname ? shortname : metadataName` and the skipFirstLine in the asLines method.
> 
> The `abbreviation` is used in several places and i'm worried we could cause some issue by changing it to metadataName by default. It could also be a pain to verify that everything keeps working fine.
> 
> You can also use `invokeCommand` instead of asLines, up to you.
> 
> `TODO: analyse uses for these and convert these over` this was made by the boss @itowlson . I don't remember what was the plan honestly. Maybe you're right and he wanted to deprecate it sooner or later.
> 
> Hope i didn't miss anything 😄 Thanks again for your help 🙏

That `jsonpath` expression is similar to (but better than) my original attempt, which I "fixed" with the '@' because it wasn't working in my WSL environment. The reason was my kubectl version (1.22.1). Unfortunately, using `{.spec.names.shortNames[0]}` also suffers from the same problem. To illustrate:

```
docker run -it --rm -v $KUBECONFIG:/.kube/config bitnami/kubectl:1.22.1 get crd -o jsonpath="{range .items[*]}{.metadata.name}{\" \"}{.spec.names.kind}{\" \"}{.spec.names.singular}{\" \"}{.spec.names.plural}{\" \"}{.spec.names.shortNames[0]}{\"\n\"}{end}"
```

Comparing these two approaches on a few different versions (turns out I'm in the very narrow range where your suggestion will fail):
- `range` + `@`: works on all versions >= 1.19 (1.18 suffers from missing newlines)
- `shortNames[0]`: works on everything except 1.22.0 and 1.22.1 (again with the missing newlines)

I'll keep praying for inspiration for an approach that works across kubectl versions. Do you have a minimum supported version? Is there any precedent in the repo for conditional behaviour based on kubectl version?

In the meantime I'll also make the change to use `invokeCommand` so that it's not stripping the first line.

peterbom:
@lstocchi - pushed an update which:
- uses `invokeCommand` instead of `asLines` (so we don't lose the first line of the output)
- uses your suggestion of `shortNames[0]`, but with the newline moved to the start of the `range` expression to work around the newline issues on 1.22.[0-1].

lstocchi (approved):
LGTM!! Thanks @peterbom for fixing it and testing out with different kubectl versions. 
I'll keep it open until Monday so also @itowlson can give it a look.
Really appreciate your help!! :pray:

itowlson (approved):

PR #1074, opened 2022-05-15
feat: add more log viewer options to global settings

SethFalco:
Adds the remaining three options to the global settings so users can override their default values. Only saved when the user clicks Save Settings.

This omits the drop-down boxes, I might do those later but just wasn't sure 100% what the best approach to serializing those were yet. Just storing an ordinal wouldn't be the most user-friendly in my opinion. 🤔


itowlson:
Agree on not wanting to store an ordinal.  You could do it as a string and enforce specific values in the config schema and TS type system?  (the `type Suit = 'heart' | 'diamond' | ...` idiom)  You would need safe converters of course so it's still a faff!

package.json
```diff
@@ -392,6 +392,21 @@
                     "default": false,
                     "description": "Set to true to show timestamps by default in the log viewer."
                 },
+                "vscode-kubernetes.log-viewer.since": {
+                    "type": "integer",
+                    "default": -1,
+                    "description": "The default value for the since field in the log viewer. Set to -1 for all logs."
```
itowlson, line 398: Documentation says 0?

README.md
```diff
@@ -207,6 +207,9 @@ Minikube tools to be installed and available on your PATH.
    * `disable-linters` - Disable specific linters by name
    * `vscode-kubernetes.log-viewer.follow` - Set to true to follow logs by default in the log viewer.
    * `vscode-kubernetes.log-viewer.timestamp` - Set to true to show timestamps by default in the log viewer.
+   * `vscode-kubernetes.log-viewer.since` - The default value for the since field in the log viewer. Set to 0 for all logs.
```
itowlson, line 210: Might be worth mentioning examples of the format so people don't need to go look at it to find out?

package.json
```diff
@@ -392,6 +392,21 @@
                     "default": false,
                     "description": "Set to true to show timestamps by default in the log viewer."
                 },
+                "vscode-kubernetes.log-viewer.since": {
+                    "type": "integer",
```
itowlson, line 396: I thought the `since` field took a string so it could have units, e.g. `5m` or `2h`?  If not, we need to say what the units are for the config setting.

package.json
```diff
@@ -392,6 +392,21 @@
...
+                    "default": -1,
+                    "description": "The default value for the since field in the log viewer. Set to -1 for all logs."
+                },
+                "vscode-kubernetes.log-viewer.tail": {
+                    "type": "integer",
+                    "default": -1,
+                    "description": "The number of recent logs to display by default in the log viewer. Set to 0 or -1 for all log lines."
+                },
+                "vscode-kubernetes.log-viewer.terminal": {
+                    "type": "boolean",
```
itowlson, line 406: Consider making this a "destination" with permitted values of `webview` and `terminal` or something like that?  (Just in case we one day come up with _yet another_ way to display logs!)

(You can still convert it to a boolean for code purposes if it's more convenient of course.)

package.json
```diff
@@ -392,6 +392,21 @@
...
                 },
+                "vscode-kubernetes.log-viewer.since": {
+                    "type": "integer",
+                    "default": -1,
+                    "description": "The default value for the since field in the log viewer. Set to -1 for all logs."
+                },
+                "vscode-kubernetes.log-viewer.tail": {
+                    "type": "integer",
+                    "default": -1,
+                    "description": "The number of recent logs to display by default in the log viewer. Set to 0 or -1 for all log lines."
```
itowlson, line 403: Is there a reason for having two ways to express this?  Maybe just have one so that users don't have to worry about which one to choose?

itowlson (changes requested):
Mostly nitpicks but I do think the `since` config needs clarifying (unless am I completely misunderstanding something).  Otherwise looks great, and will certainly make people's lives easier - thanks!

package.json
SethFalco: I can change this to 0 if you'd prefer.

I used `-1` because `0` and `-1` do the same thing in this extension, and the default value of the setting was `-1`. ([Reference](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1074/files#diff-79cdfcae916cbf064210c646791439ef36aabbbb50c25e5bcd7164f4b4bc810bR343))

The default value was changed in a previous PR I made here:
> In the log viewer, changes the default option for since-input to -1 and since-select to s. (This stills results in getSinceDuration() returning 0 by default because of sinceInput <= 0.)
> * https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1063

I thought `-1` would be more generally understood as unspecified unlike `0` which could be mistaken for 0 seconds for those less familiar with Kubernetes.

Perfectly happy to use the number 0 instead, but probably best to also change the default value back if so.

README.md
SethFalco: This is only for the number, there are 2 input fields that make up the duration.
`since-input` and `since-select`, from my understanding, the documentation suggests that the user could type it but it's inaccurate.

```js
function getSinceDuration() {
    const sinceType = document.getElementById('since-select').value;
    const sinceInput = document.getElementById('since-input').value;
    if (isNaN(sinceInput) || sinceInput <= 0 || sinceType.trim() === '') {
        return 0;
    }
    return `${sinceInput}${sinceType}`;
}
```

The string is actually built by using the number from `since-input` and appending the value from `since-select` (`h`, `m`, or `s`.)

Ofc it could be that these 2 input fields should be scrapped in favor of a single input field which is just a duration string like `5m`.

Please just let me know if you feel I've misunderstood anything.

package.json
itowlson: I don't mind whether it's 0 or -1 it was just that the changes to README.md said the default was 0 and I found it confusing that it said different things in the two places.

package.json
SethFalco: Please check this:
https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1074#discussion_r874228172

Yeah I wasn't 100% sure how to handle this part tbh.
I didn't specify units since technically, it's accompanied by another field which is the units, so it's a bit confusing. 🤔 

I'll think about this.

image

package.json
SethFalco: No particular reason, just that both did the same thing, so I thought I'd make that explicit.

Perfectly happy to only describe one. I'll probably use `-1` since that's the default value I think.

README.md
itowlson: Gotcha - thanks for explaining.  In that case I think we do need to, as you discussed, save the value of the unit drop-down as well.  (Or, as you say, change the whole thing to free entry.)  E.g. if the user sets `since` to an hour and we save `1` but then the webview comes up with the drop-down being in minutes then that would be surprising...?

package.json
SethFalco: Ahh, my bad! Yeah I should probably update the README too then! ^-^'

itowlson:
My suggestion would be to keep the existing 'since' UI, at least for now, because I am awfully conservative that way.  I don't have a strong opinion, though, except that if we change it to free entry, then we need to provide some sort of help at the point of use.

Agree on keeping the anchoring fix for a separate PR.  Thanks heaps for all your effort on this!

itowlson:
This looks great!  I've held off hitting the 'approve' button because you're still tweaking the 'since' stuff, but I'm happy for it to merge as soon as that's good to go.  Thank you!

SethFalco:
Hmm, so I was implementing the last 2 things and wasn't into storing the units in the configuration separately. Thoughts on decoupling the view from the config for `since`?

I'm thinking regardless of how it's displayed in the view, the configuration doesn't actually have to match it at all.

In other words, maybe the configuration should just store the number of seconds and that's it? Then the view can be responsible for interpreting it.

Pros:
* If we ever change how the view displays these options, the setting will continue to work as it's decoupled from the view.
* We automatically set the `since-input` and `since-select` in the simplest way. If the user set 3600 seconds out of laziness, it'll be 1 hour next time they reopen the log viewer.
* Only 1 setting instead of 2.

Cons:
* What the user selects isn't 100% what we store. (Since they can set it to 3600 seconds, but later when they come back it's 1 hour.)
* The view actually has to have code to handle this.

I've pushed this as a new commit, so it's easy to review. If you'd prefer there just be a new configuration instead, feel free to say. I personally think this is nicer.

itowlson:
> Thoughts on decoupling the view from the config for since?  I'm thinking regardless of how it's displayed in the view, the configuration doesn't actually have to match it at all.

Yep, I'm fine with this - thanks!

itowlson:
(I probably don't need to say this, but in this case the configuration docs need to say what the units are.)

itowlson:
Which you have already done.  I'll shut up and read the code.

SethFalco:
> (I probably don't need to say this, but in this case the configuration docs need to say what the units are.)

Already covered!

In `README.md` and `package.json` it specifies seconds.

* https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1074/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R210
* https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1074/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R399

SethFalco:
Just did a quick last minute change. Realized `setSinceDuration` isn't a very accurate name. (Was originally going to actually set the input fields there as well.)

Renamed it to `splitSinceDuration`.

itowlson (approved):
Looks good to me - thanks!  @lstocchi do you want to look over?

lstocchi (approved):
LGTM. Thanks!!

PR #1075, opened 2022-05-15
rework: make "that match" default option

SethFalco:
### Keywords

* "key words" → "keywords"
  * https://www.wordnik.com/words/keyword
  * https://www.dictionary.com/browse/keywords
  * https://en.wikipedia.org/wiki/Keyword

### Filter Select

I removed the "all" option, as I felt this was unnecessary.

A more intuitive way to handle this is to not apply a filter if the text boxes value is empty rather than a dedicated `all` value. An empty text box, similar to the value `-1` in a numerical text box, is generally implied as unset. (Even if a user were to interpret it otherwise, an empty string or regex matches all logs.)

Now when a user wants to filter logs, they only have to set 1 field instead of 2. (Unless they want to use a different filter mode.)

I firmly believe this improves the user-experience with the extension, but am happy to revert that if maintainers disagree.

### DIFF
--- a/src/components/logs/app/index.html
+++ b/src/components/logs/app/index.html
@@ -143,14 +143,13 @@
         <div class="mt-5">
             <span class="mr-2">Filter:</span>
             <vscode-select id="filter-select" class="w-130 mr-2">
-                <vscode-option value="all" selected>all</vscode-option>
-                <vscode-option value="include">that match</vscode-option>
+                <vscode-option value="include" selected>that match</vscode-option>
                 <vscode-option value="exclude">that don't match</vscode-option>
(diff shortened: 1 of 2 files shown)


itowlson:
I've got mixed feelings about "empty text box means unset" but if @lstocchi is okay with it then I will conclude that it's just me! **grin**

lstocchi (approved):
For me it can work. 
I remember that we tried something similar but there was some issues during the filtering in the old implementation. Or maybe i'm confusing it with another plugin. BTW by trying it it works nicely. 
Sorry @itowlson :laughing:

Issue #1040, opened 2022-02-11
Logs window - auto scroll to bottom feature isn't working


MurzNN:
On Logs configuration form we have a button "Scroll To Bottom" - on manual click it scrolls well, but it doesn't sticks, so with "Follow Logs" mode new log lines appears at bottom, and we need to click "Scroll To Bottom" button again and again manually.

As I see in https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/593 - this feature must work as separate "auto-scroll checkbox" but I don't see it in the interface:
![image](https://user-images.githubusercontent.com/336662/153571664-95f430bd-23c6-4e63-9036-cc7558debef9.png)

Is it removed in recent versions? If yes - how can I enable "auto scroll to bottom" feature now?

lstocchi:
Hi @MurzNN, no it should work as you described by default. If you are at the bottom of the page and new logs appear, the page should stick there and you should see new logs.

But i am not able to replicate your issue. To test it i created a small script which generate random strings and it works fine for me.
At first it is stick in the middle of the page, then when i click scroll to bottom it sticks at the bottom even when new logs appear (gif below).

Can you give us more infos about your env? Vscode version? Os? Any error related to this in the console (Command palette -> Open Webview Developer Tools -> Console errors)? Thank you and sorry for the trouble

![scroll-auto](https://user-images.githubusercontent.com/49404737/153588635-3537afa5-abe9-4cf5-a5d5-9c8ed57dc357.gif)

MurzNN:
**Kubernetes extension version:** v1.3.6

**VS Code versions:**
Version: 1.64.1
Commit: d6ee99e4c045a6716e5c653d7da8e9ae6f5a8b03
Date: 2022-02-07T17:17:47.733Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Linux x64 5.13.0-28-generic

Screencast with opened console:
https://user-images.githubusercontent.com/336662/153600346-4d969ceb-2425-4f30-a9fd-d0d6f798c53e.mp4

lstocchi:
Thanks @MurzNN. I'll try to update my test script to replicate it. l'll let you know if i have some news asap

SethFalco:
The issue is word wrap.
Scroll to bottom words as documented when word wrap isn't enabled.

[...]

PR #1076, opened 2022-05-15
fix: scroll to bottom and word wrap

SethFalco:
This fixes 1 bug, and reduces the severity of another.

### Word Wrap

Word wrap only _partially_ wrapped lines before.
It would split lines by whitespace only, so for example long words or minified JSON would still extend over and require a horizontal scrollbar to view the full log.

Untitled

This adds `overflow-wrap: anywhere` which tells the view that it's ok to split in the middle of words. Now users won't experience a horizontal scroll bar at all when word-wrap is enabled!

More info:
* https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

### Scroll to Bottom

We had 2 problems before when there were too many logs.

#### div Height Cap… or something?

The `a#bottom` element would no longer anchor to the bottom. This was because the parent divs height would cap out it seems and the rest of the logs were only visible because it overflowed underneath.

Why did this happen? I have no idea. ^-^'

Because of this, the `a#element` ends up above the last log, so a better approach would be to simply select and scroll to the last log.

#### Chromium or VSC Doesn't Render all Elements

When the document is a bit too long, either Chromium or VSC has an optimization which prevents all elements from being rendered.

This breaks the scroll to bottom button for example because if a user starts at the top and has thousands of logs (especially if they're very long logs with word wrap enabled) then the `a#bottom` could be pushed so far down that it doesn't render and therefore can't be selected in the DOM.

In my case, my log viewer had a height of over 100,000 pixels, so the optimization does make sense.

I removed the `a#bottom` element and instead in the `scrollToBottom` function just select the last log and scroll to that instead.

More info:
* https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector
* https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type

However, this solution isn't perfect either.

Before when there were many logs, scroll to bottom was basically broken.

Now, if there are _way too many_ logs, it'll only scroll the user part of the way down if they're start all the way at the top.

Now we select the last rendered log, so even if the bottom isn't visible in VSC, we can at least scroll down to the bottom visible one. (And when the user is at or near the bottom, it'll work normally.)

* Mostly fixes x https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1040
  * It's not a 100% fix, but it's a way better situation than before.

---

Sorry, I'd like to include more screenshots/GIFs but I'm testing with things for work so I can't show the actual logs. ^-^'


itowlson (dismissed):
Great stuff - thanks for diving so deeply into the rendering side!

lstocchi (commented):
I don't like it too much honestly. 
With this patch we are converting the `scroll to bottom` to `page down`.
I don't think it would solve the issue users have.

The problem here is that the back-end pre-calculate the height of the div container (the one which will be used to render the scroll bar) by using the original number of rows to render without considering if the `wrap lines` option is checked. So if we need to display 1000 long rows, the backend assumes that the bottom of the page will be the 1000th row (~N pixel), which may not be true if the `wrap line` is checked. In that case we could ends up with 1000/1500/2000+  rows (~N+m pixel).

Now, if we keep in mind this behavior what we need to do to always make the scroll to bottom works (both with wrap enable and disable) is to go to the latest page (which will always be the same with both `wrap lines` enabled and disabled as the backend doesn't give a s... about it :laughing: ) and then scroll to the bottom.

Example:
By taking the 1000 rows example above. Let's assume we split them to have 10 pages so 100 rows displayed per time.
At the beginning the backend pre-calculate the height of 1000 rows to render the scrollbar => height: 100000px
With `wrap lines` disabled (100000px to render): -> we scroll to 100000 and the `top` of the content div would be 100000 - something (which should be the height of the page or lines rendered). Everything works fine.
With `wrap lines` enabled (100000px + M pixel to render) -> we scroll to 100000 and the `top` of the content div would be 100000 - something (which should be the height of the page or lines rendered). The problem is that the end of the div is not really 100000 but a bit more (M pixel more exactly - it goes outside of the precalculated height as we wrap lines and they get more spaces). But we are at the latest page so everything is already rendered on view. So we perform another scroll down to go to the actual bottom.

By having something like 

```
    const innerLogPanel = document.getElementById("innerLogPanel");
    const logPanel = document.getElementById("logPanel");
    setTimeout(() => {
        logPanel.scrollBy({top: innerLogPanel.offsetHeight, left: 0});
        setTimeout(() => {
            document.getElementById('bottom').scrollIntoView();
        }, 100);
    }, 200);
```
should work? :thinking: 
I don't know why but without the setTimeout it doesn't always works. Or maybe it works by luck the other times. :laughing: 

Did i say some nonsense? The only issue i foresee, if what i wrote above works, is if logs are written very fast and the scroll to bottom is slower. But maybe the setTimeout can be removed :man_shrugging:

SethFalco:
>  With this patch we are converting the scroll to bottom to page down.

I actually disagree with this way of putting it. It's very different from page-down. 🤔 

Plus a before and after of where the "Scroll to Bottom" button takes the user with the same amount of logs.

### Before
image
> Less then half way.

### After
image
> All the way to the bottom in this case.

With your proposed implementation, all it does for me is jitter the view to the same place that the old implementation would take the user.

Peek 2022-05-18 16-17

SethFalco:
Even using the dev tools to "Scroll Into View" on `#bottom` doesn't take the user to the bottom while it's rendered.

Meanwhile, doing "Scroll Into View" with selecting the final element (via the GUI) does as intended and scrolls to the bottom.

### Before
image

### After
image

SethFalco:
Ahh… I was investigating this further… -_-
I'm a dummy.

I didn't notice _we_ (as in the extension itself) was setting the height of the `#innerLogPanel` element. XD

I'm not sure why we are yet, but I can at least concede that perhaps my proposal isn't the optimal solution in that case now that I've seen that.

tejhan:
Thank you for this.

Closing as this has been fixed with PR #1516

PR #1079, opened 2022-05-20
Publish 1.3.9

itowlson:
@lstocchi this is primarily to get the Intellisense fix out (I saw you fielded another issue about it, thanks!) - if there are other PRs you want to get in then no worries


lstocchi (approved):
Yeah i think we can go with this. LGTM

PR #1081, opened 2022-05-29
Use XDG Base Directory variables when creating the base installation directory

hyperupcall:
The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) specifies various user directories for applications to store data. There are a lot of [good reasons](https://xdgbasedirectoryspecification.com) to support this, and this PR adds support in a backwards-compatible way

- This PR adheres to the spec by creating the `vs-kubernetes` directory in `~/.local/state`, but _only if_ a directory has not been created at `~/.vs-kubernetes` _and_ the current platform is Linux
  - In the case where we can follow the XDG Base Directory Specification, users can customize where the `vs-kubernetes` directory is created with the `XDG_STATE_HOME` variable. If the env var is empty or if it's a relative path (as per the spec), the default of `~/.local/state` is used

### DIFF
--- a/src/components/installer/installationlayout.ts
+++ b/src/components/installer/installationlayout.ts
@@ -1,9 +1,25 @@
+import * as os from 'os';
 import * as path from 'path';
 
+import { fs } from '../../fs';
 import { Platform, Shell } from "../../shell";
 
+export function vsKubernetesFolder(shell: Shell): string {
+    const originalDir = path.join(shell.home(), `.vs-kubernetes`);
+
+    if (fs.existsSync(originalDir) || os.platform() !== `linux`) {
+        return originalDir;
+    }
+
+    let xdgStateHome = process.env.XDG_STATE_HOME || ``;
+    if (xdgStateHome[0] !== `/`) {
+        xdgStateHome = path.join(os.homedir(), `.local/state`);
+    }
+    return path.join(xdgStateHome, `vs-kubernetes`);
+}
+
 export function baseInstallFolder(shell: Shell): string {
-    return path.join(shell.home(), `.vs-kubernetes/tools`);
+    return path.join(vsKubernetesFolder(shell), `tools`);
 }
 
 export function getInstallFolder(shell: Shell, tool: string): string {
--- a/README.md
+++ b/README.md
@@ -275,6 +275,10 @@ The following directives are supported:
 * `exclude`: Do not treat this file as a Kubernetes manifest. This is useful when you have
   files that look a lot like Kubernetes manifests but aren't, such as Kustomize patch files.
 
+## Extension data
+
+This extension stores its associated files and data in `$XDG_STATE_HOME/vs-kubernetes` (at `~/.local/state/vs-kubernetes` by default). However, if the legacy directory of `~/.vs-kubernetes` still exists, then that will be used instead. 
+
 ## Known issues
 
   * `Kubernetes: Debug` command currently works only with Go, Node.js, Java, Python and .NET applications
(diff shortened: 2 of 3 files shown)


src/components/installer/installationlayout.ts
```diff
@@ -1,9 +1,25 @@
...
+export function vsKubernetesFolder(shell: Shell): string {
+    const originalDir = path.join(shell.home(), `.vs-kubernetes`);
+
+    if (fs.existsSync(originalDir) || os.platform() !== `linux`) {
+        return originalDir;
+    }
+
+    let xdgStateHome = process.env.XDG_STATE_HOME || ``;
+    if (xdgStateHome[0] !== `/`) {
+        xdgStateHome = path.join(os.homedir(), `.local/state`);
```
lstocchi, line 16: I think it would be better to keep the default path as it is now `path.join(shell.home(), '.vs-kubernetes')`. If one deletes the existing folder for some reason and then it is recreated somewhere else the user doesn't get what's going on.
Then by doing it we would only have two different cases 1) user who doesn't use XDG specification 2) users who use it... and we don't need to remember that old linux/windows/mac users uses `.vs-kubernetes` folder, new linux users use `.local/state` and people who use xdg specification may have their custom folder.

lstocchi (commented):
I didn't know the XDG base directory specification so thanks for this. Very interesting.

The only concern I have is to use a different default path when the user is not adhering to this specification. See comment below.

The other aspect is we need to update the README otherwise nobody would know that the extension leverages it.

BTW I'm also open at accepting all changes, I would wait to hear from @itowlson

src/components/installer/installationlayout.ts
hyperupcall: Ah, so if I understand correctly, your point is that it would be less confusing to users if there is an explicit "opt-in" to the "XDG Base Dir Specification". So for example, we would only follow the spec if `XDG_STATE_HOME` is set explicitly?

I feel like doing it this way would be more counterintuitive for the user and we would realize less gains from attempting to use the spec in the first place. Mostly because most users expect things to go in `~/.local/state`, `~/.config`, etc. automatically. The spirit of the spec is that this should be automatic behavior, and if we have to check if `XDG_STATE_HOME` is set (which _very_ few people actually do), then most people wouldn't be affected by this benefit (and in this case i would rather the PR not be merged since its an incomplete solution). There is also the issue of potentially having to do more of these checks for every new XDG variable used (assuming consistency is valued here) - whereas currently we only ever need to do `existsSync(~/.vs-kubernetes)` once

I think your statement about the end-user confusion when deleting `~/.vs-kubernetes` and it being recreated somewhere else is definitely valid, and I believe your mention of updating the README to address this PR in general can address that. Maybe we can say somewhere in the README along the lines of "the extension will follow the XDG Base Directory Specification appropriately (defaulting to writing config and other files in `~/.local/state/vs-kubernetes`), but only if the legacy directory of `~/.vs-kubernetes` does not exist". That way, if the user searches for "config" or ".vs-kubernetes" in the docs, they know what's up right away

itowlson: I like "use the legacy directory if it exists, otherwise use the XDG directory."  Seems like a nice pragmatic solution.

itowlson (approved):
Looks good to me - thanks!  @lstocchi you okay with these changes?

itowlson:
@hyperupcall Could you sign off for the DCO please?  Thanks!

lstocchi (approved):

hyperupcall:
DCO done! :+1:  Also as mentioned before, I also updated the readme to explicitly document this to prevent any user confusion :)

PR #1086, opened 2022-06-07
add justMyCode debug option

yehiyam:
Add the `justMyCode` option to the debug configuration when debugging [.NET](https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md#just-my-code) and [python](https://code.visualstudio.com/docs/python/debugging#_justmycode) with `debug (attach)`
The `justMyCode` property instructs the debugger to *not* skip loading external code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,1375 +1,1385 @@
 {
-    "name": "vscode-kubernetes-tools",
-    "displayName": "Kubernetes",
-    "description": "Develop, deploy and debug Kubernetes applications",
-    "version": "1.3.9",
-    "publisher": "ms-kubernetes-tools",
-    "engines": {
-        "vscode": "^1.52.0"
-    },
-    "license": "MIT",
-    "categories": [
-        "Snippets",
-        "Linters",
-        "Debuggers",
-        "Azure",
-        "Other"
-    ],
-    "keywords": [
-        "kubernetes",
-        "helm",
-        "aks",
-        "gke",
-        "aws"
-    ],
-    "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
-    "icon": "images/k8s-logo.png",
-    "activationEvents": [
-        "onCommand:extension.vsKubernetesCreate",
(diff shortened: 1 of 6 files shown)


yehiyam:
@itowlson, @lstocchi, @bnookala 
Could someone take a look at the PR?

package.json
```diff
@@ -1,1375 +1,1385 @@
 {
-    "name": "vscode-kubernetes-tools",
```
itowlson, line 2: Is the only change to this the whitespace?  GitHub doesn't make it clear what's happened but I wonder if it accidentally got changed from spaces to tabs?  It should be spaces with an indent of 4.

package.json
```diff
@@ -1,1375 +1,1385 @@
...
+		"typescript": "^3.5.2",
+		"vscode-test": "^1.4.1",
+		"webpack": "^4.35.2",
+		"webpack-cli": "^3.3.5"
+	},
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools"
+	},
+	"__metadata": {
```
itowlson, line 1379: Not sure where this came from - is it intentional?

itowlson (changes requested):
Thanks for this!  It looks good to me except for the broad changes to `package.json` (where I think you meant only to add the new config setting) and `package-lock-json` (which I think doesn't need to change, though I could of course be wrong!).  Happy to merge once those are tidied up.

package.json
yehiyam: Not sure either. It wasn't intentional

package.json
yehiyam: The indent was changed to tabs. Reverted back to spaces

itowlson (approved):
Looks good - thanks!

lstocchi (approved):

itowlson:
@lstocchi thanks - I think this knocks off a reasonable round of PRs - shall we do a release or is there anything else you'd like to get in?

lstocchi:
@itowlson +1 for a new release

PR #1090, opened 2022-06-12
Fix snippets placeholders

heitorlisboa:
Fixed some snippets placeholders that were missing the `$` (dollar sign) symbol


itowlson (approved):
Good eye, thanks for the catch and the fix!

PR #1099, opened 2022-07-28
Implement Copy All action

### LINKED ISSUE #1025: Unable to copy from Logs
I am unable to copy text from the logs for a pod.  I'm not sure if this is a bug or a feature.  But it is often nice to copy the information for put in either DevOps tasks or bugs.

orgads:
- Fixes #1025

### DIFF
--- a/src/components/logs/app/index.html
+++ b/src/components/logs/app/index.html
@@ -161,6 +161,7 @@
             </div>
             <vscode-checkbox id="wrap-chk"></vscode-checkbox>
 
+            <vscode-button id="copyAllBtn" class="float-right mr-50">Copy All</vscode-button>
             <vscode-button id="bottomBtn" class="float-right mr-50">Scroll To Bottom</vscode-button>
         </div>
 
--- a/src/components/logs/app/main.js
+++ b/src/components/logs/app/main.js
@@ -168,6 +168,12 @@ function init() {
         });
     });
 
+    const copyAllBtn = document.getElementById('copyAllBtn');
+    copyAllBtn.addEventListener('click', (_event) => {
+        const content = isFiltering() ? filteredContent : fullPageContent;
+        navigator.clipboard.writeText(Object.values(content).join("\n"));
+    });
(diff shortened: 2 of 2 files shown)


orgads:
This only copies 282 lines :(

How can I improve it?

lstocchi (commented):
Sorry for the late response. By a quick glance the code looks good, maybe we need just to better position the new button.

My concern is about the amount of stuff you are copying and writing to the clipboard. Could it cause some performance issues with huge logs? Could we just copy logs that are actually visible to the user or ask her to provide a range?

Then, is this supported by all browsers?

I'll try to test it more wisely tomorrow. Thank you!!

lstocchi:
Sorry for the super-late reply. Just tried to create a job which pushes 10mb of random text (to simulate a lot of dirty logs) and the `copy action` doesn't work. No errors in the console. But it doesn't paste anything. If i reduce the amount of text it works fine. Maybe some physical limitation of the copy action?? 
I agree this is an edge case but we should at least find the reason and the limit of such a feature so we can tell users when they could expect it to work or not. 

BUT what's the need for this action? Why would someone want to copy everything? If the reason is to have a way to extract all logs at once we could work on a specific action. Let me know if i'm missing something.
 
@itowlson WDYT?

orgads:
I need it for pasting everything in a new tab, then I can search in it, highlight, replace etc.

lstocchi:
@orgads ok so if we create an export logs action this would be solve your issue, right? 

I've also found out that, starting from vscode 1.70, one can customize the context menu in a webview. We could add our custom copy action there or creating some sub-copy actions (like copy all visible logs, copy selected text, copy first/last 5000 rows ) so to avoid issues with very long texts. Atleast until there is a solution that works. @itowlson ??

orgads:
I prefer to have the text in a new tab rather than a file, then I can choose to save it if I want (on most cases I don't).

orgads:
reopen

PR #1110, opened 2022-09-15
add helm upgrade support

cloorc:
Add support for helm upgrading will increse our efficiency during development on kubernetes cluster. Thanks a lot!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,13 @@
 {
     "name": "vscode-kubernetes-tools",
-    "version": "1.3.9",
+    "version": "1.3.10",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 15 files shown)


cloorc:
Add convenient menu on Pod/Job for 'Logs last 300 rows & Follow' to follow logs on shared terminal

lstocchi:
@bnookala @cloorc i see a lot of stuff in this PR. As i am not an expert in helm and i don't work actively with it, can you please tell me what we are trying to solve here?

cloorc:
> @bnookala @cloorc i see a lot of stuff in this PR. As i am not an expert in helm and i don't work actively with it, can you please tell me what we are trying to solve here? A bit of context would help

It's a little complecated. There are a lot commits on this PR. Let me close it first and split it into several pieces on different parts.

PR #1121, opened 2022-10-27
Set debug session name to pod name on attach

WBrettDavis:
This change sets the target kubernetes pod's name as the default name for the debug session when using the "Debug (Attach)".

Currently, by passing in 'undefined' it's falling back to a numeric datetime value as the debug session name and when attaching debuggers to multiple containers this becomes very confusing for the user.

### DIFF
--- a/src/debug/debugSession.ts
+++ b/src/debug/debugSession.ts
@@ -91,7 +91,7 @@ export class DebugSession implements IDebugSession {
         }
 
         vscode.window.withProgress({ location: vscode.ProgressLocation.Window }, async (p) => {
-            let appName;
+            let appName: string | undefined;
             try {
                 // Build/push docker image.
                 p.report({ message: "Building Docker image..."});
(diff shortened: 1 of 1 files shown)


lstocchi (changes requested):
Thanks @WBrettDavis  and sorry for the late response.
The idea looks good to me but this PR should also include a check to prevent the useless clean up that happens when the debug session is finished. 
In launch mode a new deployment is created and then removed at the end. In attach mode this does not happen.
Could you please fix this? Thank you!!

src/debug/debugSession.ts
```diff
@@ -361,10 +363,7 @@ export class DebugSession implements IDebugSession {
             if (proxyResult && proxyResult.proxyProcess) {
```
WBrettDavis, line 363: I left this part in the default callback since it'll be needed at the end of every debug session to close port-forward. Once complete, I call the caller-specified callback for running and context-specific cleanup

lstocchi (approved):
LGTM!! Thanks @WBrettDavis :rocket:  
I'll merge this tomorrow so others have the time to give it a look.

PR #1129, opened 2022-11-16
Fix logs webview jumps back when scrolling up (#1113)

### LINKED ISSUE #1113: logs webview jumps back when scrolling up
The webview is a great feature to go through logs, but I encounter an annoying behavior when scrolling up in the logs. It doesn't happen when scrolling down.

Reproduce:
Open a pods logs and scroll down and up a bit. At some point the view starts to jump back repeatedly to the same position of the logs.
It also happens when using the PgUp key.

It's independent from following the logs or having that running/stopped.

lstocchi:
it resolves #1113


Tatsinnit (approved):
❤️🎉 thanks for this, looks good to me.

PR #1130, opened 2022-11-20
Add inline buttons for describe, logs, and terminal

jamesorlakin:
In our small development team I noticed our interactions with K8S commonly boils down to three tasks: describing a pod to see its status, viewing its logs (and not needing to type a lengthy `kubectl` command!) and occasionally jumping into a terminal to play with args for probes manually.

This PR adds those commands to the inline group to make them easier to reach than the context menu:
image

(If this isn't desirable feel free to close! 🙂)

### DIFF
--- a/package.json
+++ b/package.json
@@ -627,6 +627,11 @@
                 },
                 {
                     "command": "extension.vsKubernetesDescribe",
+                    "group": "inline",
+                    "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource(?!\\.namespace).*/i"
+                },
+                {
+                    "command": "extension.vsKubernetesDescribe",
                     "group": "3",
                     "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource(?!\\.namespace).*/i"
                 },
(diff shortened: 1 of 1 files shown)


squillace (approved):

lstocchi (commented):
I like the idea but I wonder if we can make it customizable somehow. Maybe someone else would like to add other actions as inline buttons. I'll try to give this a look in the next days but if all of you are ok with it, it's good to be merged for me.

lstocchi:
So, an idea could be to add a setting (an array) where users can add actions they'd like to see in their k8s tree as inline buttons. We should make the insert easy with code completion 
An example could be 
```
"vscode-kubernetes.inline-actions": [ "logs", "describe", "portforward" .... ]
```

When building the tree, we take the inline actions value and by merging all args we create a string to be added in the node context
```
treeItem.contextValue =  `vsKubernetes.resource.pods.inner-log-describe-portforward`;
```

Finally, in the package.json , we add a specific rule for inline actions with a custom `when` clause
Something like
```
{
                    "command": "extension.vsKubernetesLogs",
                    "group": "inline@3",
                    "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource\\.(pod|job)\\.(inner.*-log)/i"
                },
```

The cons of this is that we need to use the rule of the original action. In the case of logs, the inline command will be visible in both pods and jobs resources.
The pro is that it is customizable and if someone wants to add or remove inline actions they can do that.

Have i missed something? Can you think at something better? 
WDYT? @squillace @itowlson @Tatsinnit

Tatsinnit:
❤️ Thank you so much for adding to this discussion @lstocchi , this looks like a nice idea. I tested this via sues branch as well and here is the outcome. Thank you and apologies for late reply.

Screenshot 2022-11-27 at 9 44 13 AM

Tatsinnit (approved):

Issue #1132, opened 2022-12-01
Extension failed with `kubectl not found` while it is in the path and setting


tigerinus:
![image](https://user-images.githubusercontent.com/7172560/205116603-fc40398c-ffab-422f-94a9-6239832b5b06.png)

dserodio:
I'm also having this issue.

VS Code 1.73.1
Kubernetes Extension v1.3.11
macOS 12.6.1 

I just opened my settings.json file and saw that this path is set in more than one key, and I never edited the JSON manually, only thru the extension's config panel:

```
  "vscode-kubernetes.kubectl-path.mac": "/Users/danielserodio/.asdf/shims/kubectl",
  "vscode-kubernetes.kubectl-path": "/Users/danielserodio/.asdf/shims/kubectl",
  "vs-kubernetes": {
    "vs-kubernetes.namespace": "",
    "vs-kubernetes.kubectl-path": "",
    "vs-kubernetes.helm-path": "",
    "vs-kubernetes.minikube-path": "",
    "vs-kubernetes.kubectlVersioning": "user-provided",
    "vs-kubernetes.outputFormat": "yaml",
    "vs-kubernetes.kubeconfig": "",
    "vs-kubernetes.knownKubeconfigs": [],
    "vs-kubernetes.autoCleanupOnDebugTerminate": false,
    "vs-kubernetes.nodejs-autodetect-remote-root": true,
    "vs-kubernetes.nodejs-remote-root": "",
    "vs-kubernetes.nodejs-debug-port": 9229,
    "vs-kubernetes.dotnet-vsdbg-path": "~/vsdbg/vsdbg",
    "vs-kubernetes.local-tunnel-debug-provider": "",
    "checkForMinikubeUpgrade": true,
    "imageBuildTool": "Docker"
  }
```

Tatsinnit:
❤️☕️ Thank you so much for this issue @tigerinus and @dserodio , here is what I have noticed and try to replicate and to me this seems like incorrect setting path, I was able to replicates this only if `vscode-kubernetes.kubectl-path` and `vscode-kubernetes.kubectl-path.mac` are outside the scope of `"vs-kubernetes": { ... }` 

2 enable here are the 2 options:

* Option 1:  If you move those 2 setting under `"vs-kubernetes": {..}` it should help this extension sees correct path of `Kubectl`
* Option 2: If you let `install kubectl` button perform its underlying operation it will set you up with the setting

[...]

PR #1133, opened 2022-12-04
Update readme to stable path for kubectl/helm/minikube.

Tatsinnit:
Thank you so much in advance for taking a look at this PR, this PR updates readme to to the non-deprecated settings path for `kubectl, helm and Minikube` 

I found this while trying to replicate an issue which is noon-issue if path is set correctly, or extension defaults are set from the extension installs. https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1132  **deprecated information** here: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/master/package.json#L114

Thanks heaps ❤️☕️🙏🎊 , a gentle ping and cc: @lstocchi , @squillace , @itowlson or @gambtho

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,10 +186,10 @@ Minikube tools to be installed and available on your PATH.
 
    * `vs-kubernetes` - Parent for Kubernetes-related extension settings
        * `vs-kubernetes.namespace` - The namespace to use for all commands
-       * `vs-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
+       * `vscode-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
        * `vs-kubernetes.helm-path` - File path to the helm binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
        * `vs-kubernetes.minikube-path` - File path to the minikube binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -186,10 +186,10 @@ Minikube tools to be installed and available on your PATH.
 
    * `vs-kubernetes` - Parent for Kubernetes-related extension settings
        * `vs-kubernetes.namespace` - The namespace to use for all commands
-       * `vs-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
+       * `vscode-kubernetes.kubectl-path` - File path to the kubectl binary. Note this is the binary file itself, not just the directory containing the file. On Windows, this must contain the `.exe` extension.
```
Tatsinnit, line 189: 💡Wondering and please add your thoughts: I wonder if this is deprecated: and the new setting exist above `vs-kubernetes: {...}` scope. So this should be removed all together.

itowlson, line 189: As you allude to in your comment, the new paths are not nested.  The non-nesting is why they exist - the trust model only allows us to identify _top-level_ config settings as machine-scope (and therefore not overridable by workspaces).  With the nested settings, we could only have set a policy on the _whole_ `vs-kubernetes` wrapper object.  Hence the top-level settings.  These are now the preferred ones; the old nested ones exist for backward compatibility.

In addition, as you have spotted, the names are slightly different.  The old keys use `vs-kubernetes` for compatibility all the way back to Brendan's original extension.  The new ones use `vscode-kubernetes` because that's closer to what the extension is now called.

So the story should be:

* Document the top-level settings
* Remove the documentation for the nested settings
* Use the top-level settings if present and fall back to the nested ones if not (this should already be handled by `components/config/config.ts/getToolPath`)

You can see the structure in `package.json` but you may have to do a lot of collapsing because there are a lot of nested settings between the big object and the top-level settings!

itowlson (changes requested):
Thanks for documenting this and sorry for missing it when I did the workspace trust stuff.  I'm afraid the multiplicity of names and levels is confusing, and the structure described in this PR is not quite right - I _think_ all you need to do is pull your revised entries up to the top level, but I've written a comment that explains the changes and motivations in more detail so hopefully that will provide a signpost for you to check against the code.

README.md
Tatsinnit: Aha, thank you so much for the background ❤️☕️🙏, yes I was and only after searching found out the `deprecated` bits, sound good, this seems like careful removal work, I will try and continue the path mentioned above and make the changes. Thanks heaps.

Issue #1127, opened 2022-11-10
Feature request - different colrs for different pod state


satanshiro:
currently vs code kubernetes extension shows only the running and terminating state of pods(noted by green and red respectively).
I would love to be able to differentiate and see the other status types such as completed, waiting and possibly capture container status also such as not ready when health checks are failing.
I don't have a detailed request but I thought if this gets enough attention I'd give a proposal. 
what are your thoughts?
I really enjoy the simplicity and integration of the plugin, thanks!

carljmosca:
I made some minor changes to this and did a pull request here: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1142

- pods with not ready container(s) have an orange dot instead of green
- pods which are completed have a black dot instead of green

It's a pretty small start to what I think you're asking (and I needed).

PR #1146, opened 2023-01-17
added notReady and completed pod status

carljmosca:
This is in reference to feature request #1127.  Not ready pods are yellow and completed pods are black.

### DIFF
--- a/src/components/clusterexplorer/resourcekinds/resourcekind.pod.ts
+++ b/src/components/clusterexplorer/resourcekinds/resourcekind.pod.ts
@@ -13,14 +13,21 @@ export const podUICustomiser = {
     customiseTreeItem(resource: ResourceNode, treeItem: vscode.TreeItem): void {
         const podInfo = resource.extraInfo!.podInfo;  // TODO: unbang
         if (podInfo && podInfo.status) {
-            treeItem.iconPath = getIconForPodStatus(podInfo.status.toLowerCase());
+            treeItem.iconPath = getIconForPodStatus(podInfo.status.toLowerCase(), podInfo.ready);
         }
     }
 };
 
-function getIconForPodStatus(status: string): vscode.Uri {
-    if (status === "running" || status === "completed") {
-        return assetUri("images/runningPod.svg");
+function getIconForPodStatus(status: string, ready: string): vscode.Uri {
+    if (status === "running") {
+        const readyFields = ready.split("/", 2);
+        if (readyFields[0] === readyFields[1]) {
+            return assetUri("images/runningPod.svg");
+        } else {
+            return assetUri("images/notReadyPod.svg");
+        }
+    } else if (status === "completed") {
+        return assetUri("images/completedPod.svg");
     } else {
         return assetUri("images/errorPod.svg");
     }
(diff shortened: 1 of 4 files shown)


carljmosca:
I don't have node 14 for my (M1) mac.  Is there something obvious that needs attention with this failing test?

lstocchi:
@carljmosca it just needed a restart :smile: BTW thanks for your help and sorry for not responding. I'm in the middle of a reorganization and I'm quite busy, i'll try to look at this tomorrow :+1:

carljmosca:
Excellent, and thank you.  I was just wanted to ensure there was not some process waiting on me, LOL.  And thank you (and others) for a great extension.

itowlson (approved):
LGTM, nice one!

carljmosca:
@itowlson thank you!!

src/components/clusterexplorer/resourcekinds/resourcekind.pod.ts
```diff
@@ -13,14 +13,21 @@ export const podUICustomiser = {
...
         }
     }
 };
 
-function getIconForPodStatus(status: string): vscode.Uri {
-    if (status === "running" || status === "completed") {
-        return assetUri("images/runningPod.svg");
+function getIconForPodStatus(status: string, ready: string): vscode.Uri {
+    if (status === "running") {
+        const readyFields = ready.split("/", 2);
```
Tatsinnit, line 23: 💡 Thank you so much for adding to this review ❤️🙏 Idea / Just some thought to share: Might be if we add comment with regards to what is going on here: aka - return sample `ready: string` expected value and why only limit of 2, also, will this separator be different base on O/S. Thanks.

Tatsinnit (approved):

src/components/clusterexplorer/resourcekinds/resourcekind.pod.ts
carljmosca: Thank you @Tatsinnit for the review.  I will add the needed comment.  I am not sure I follow the question about the limit of two.  I am expecting a string that includes the number of ready and total containers separated by a / character (which I see on macOS, Windows, and Linux).  I am guessing you know something that I have missed about the Ready column values or perhaps the string split method.  Can you please elaborate on the limit of 2 question?

Tatsinnit: Thank you so much @carljmosca, sounds good yeah, further clarification why I shared the observation is below.

- Regarding path - yeah all good, thought I will mention - hence I was sharing/wondering if use of `os.sep` or `path.sep` which could take care of any discrepancy in cases where the path do not have `/`. https://stackoverflow.com/questions/125813/how-to-determine-the-os-path-separator-in-javascript 
- Re: **Limit** - what I was meaning to share the observation of use `split([separator][, limit]);` so the limit of `2` will return first `2` and I think new images are in location `images/newimages.svg` so should be good -  : https://jsfiddle.net/75k1pt8z/  

Thanks heaps and I thought I will share this just as fyi in case. 🙏❤️

lstocchi:
Thank you guys!! :rocket:

PR #1179, opened 2023-06-01
feat: add setting to suppress helm not found warning (#1175)

### LINKED ISSUE #1175: Option to switch off Helm
I don't use Helm, so it isn't installed in my environment. This VSCode extension moans at me constantly "Cannot find Helm binary" and there seems to be no way to switch it off. Am I missing something?

lstocchi:
it resolves #1175


Tatsinnit (approved):
❤️🙏🎊 Looks good, thank you so much.

Issue #940, opened 2021-05-05
Support relative path in tool location settings


sidewinder12s:
Using the following to point VS Code at the Helm install doesn't appear to work:

```json
        "vs-kubernetes": {
            "vs-kubernetes.helm-path.linux": "${workspaceFolder}/tools/bin/helm"
        }
```

Is this expected? I've noticed some extensions support this while others do not.

itowlson:
Hmm, I don't know if VS Code is meant to expand that automatically or if it is extension-specific functionality.  Could you let us know one that does support it and we'll ~see how they do it and copy that~ investigate?  Thanks!

By the way, we recently removed support for configuring binaries at the workspace level, because of the security risk of a malicious workspace [say, a public repo pretending to be useful samples] pointing at a nefarious binary.  So that could be a factor too - are you writing this within a workspace's `settings.json` or at the global level?

sidewinder12s:
The Terraform extension appears to support it, at the workspace level.

I'll have to try without workspace, I was trying to configure it as part of the workspace (for a company wide workspace).

kaiba42:
bump. This is a blocker to creating developer environments with tools and dependencies that are local to a workspace. For example, https://devenv.sh

Issue #1141, opened 2022-12-23
Support `~` symbol for `${HOME}` in `vs-kubernetes.kubeconfig`


felipecrs:
Which would be very helpful for devcontainer scenarios, like:

```jsonc
// devcontainer.json
{
  "settings": {
    "vs-kubernetes": {
      "vs-kubernetes.kubeconfig": "~/.kube/config"
    }
  }
}
```

Tatsinnit:
💡❤️🙏 Thank you so much for this. Since this change exist in `settings.json` I think more permanent solution could be if VsCode Support variables when resolving values in settings from `settings.json`. (More context below) 

* It will be good if we could achieve the [`predefined variable`](https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables) enablement via `VsCode` which will lead to the auto resolution of the these variable before extension can see it. I found an old open WI around this for vscode https://github.com/microsoft/vscode/issues/2809 , please correct me in case I misunderstood anything. Thank you so much in advance.

For more thoughts, ideas share, fyi to @lstocchi , @squillace and @itowlson ☕️❤️ thanks heaps.

felipecrs:
I agree. Perhaps this could help for now: https://www.npmjs.com/package/vscode-variables

lstocchi:
Closing as https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1180 was merged. 
Now you can write the path as `"vs-kubernetes.kubeconfig": "${userHome}/.kube/config"`

PR #1180, opened 2023-06-01
feat: handle predefined vscode variables in extension setting (#940)

lstocchi:
This PR allows to handle a set of vscode predefined variables when used as values in the extension settings  -> https://code.visualstudio.com/docs/editor/variables-reference

This should resolve https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/940 and https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1141

Now the user could write a setting like 
```
"vscode-kubernetes.kubectl-path": "${userHome}/my_custom_kubectl_folder/kubectl"
```

The interpolation function has been taken from https://github.com/DominicVonk/vscode-variables and updated. The license allowed it.
I couldn't import the library as it was very old and not maintained apparently. It relies on the old vscode dependency which break this one when imported (the k8s-tools depends on the new `@types/vscode` and `vscode-test`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ Minikube tools to be installed and available on your PATH.
 
 ## Extension Settings
 
-  The Kubernetes extension supports variable substitution for the set of settings that requires a path as value. Variable substitution is supported using `${variableName}` syntax. A list of the supported variables can be found at [variables-reference](https://code.visualstudio.com/docs/editor/variables-reference).
+  The Kubernetes extension enables variable substitution for settings that require a path as a value. Variable substitution can be done using the `${variableName}` syntax. You can find a list of the variables supported in the [variables-reference](https://code.visualstudio.com/docs/editor/variables-reference) documentation.
 
    * `vs-kubernetes` - Parent for Kubernetes-related extension settings
        * `vs-kubernetes.namespace` - The namespace to use for all commands
--- a/README.md
+++ b/README.md
@@ -184,6 +184,8 @@ Minikube tools to be installed and available on your PATH.
 
 ## Extension Settings
 
+  The Kubernetes extension supports variable substitution for the set of settings that requires a path as value. Variable substitution is supported using `${variableName}` syntax. A list of the supported variables can be found at [variables-reference](https://code.visualstudio.com/docs/editor/variables-reference).
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -184,6 +184,8 @@ Minikube tools to be installed and available on your PATH.
 
 ## Extension Settings
 
+  The Kubernetes extension supports variable substitution for the set of settings that requires a path as value. Variable substitution is supported using `${variableName}` syntax. A list of the supported variables can be found at [variables-reference](https://code.visualstudio.com/docs/editor/variables-reference).
```
Tatsinnit, line 187: ☕️ 💡 Just recommendation: used ai to rephrase this: 


> The Kubernetes extension enables variable substitution for settings that require a path as a value. Variable substitution can be done using the ${variableName} syntax. You can find a list of the variables supported in the [variables-reference](https://code.visualstudio.com/docs/editor/variables-reference) documentation.

src/utils/interpolation.ts
```diff
@@ -0,0 +1,85 @@
...
+*  This file is based on the work made by DominicVonk at https://github.com/DominicVonk/vscode-variables
+*  This has been imported and updated to allow using the latest @types/vscode library
+*/
+import * as vscode from 'vscode';
+import * as process from 'process';
+import * as path from 'path';
+import { homedir } from 'os';
+
+export function interpolateVariables(value: string | undefined, recursive = false): string | undefined {
+    if (!value) {
```
Tatsinnit, line 11: 💡 Purely as suggestions, shall we comment for this function for its usage et. al. although if anyone look at the description and other open issues they could understand the borrowed use of this function. Thank you so much for this.

Tatsinnit (approved):
❤️🙏☕️ Looks good to me and given there are tests, I will defer any specific comments to @itowlson or @squillace but since we borrowed this from the https://www.npmjs.com/package/vscode-variables ; I believe it will be a good enabler for this specific case.

src/utils/interpolation.ts
lstocchi: Updated, let me know if you want something else to be added

README.md
lstocchi: Updated. Thanks for the suggestion!! 🚀

Issue #1170, opened 2023-04-23
Improper YAML Causes Memory Leaks in VSCode


h3xcat:
An improperly formatted YAML file results in memory leaks when attempting to open it in Visual Studio Code with the Kubernetes extension enabled.

### Test YAML file: (include dashes and dots)
```yaml
---
test: "aaa
...
```

### Error Message:
![](https://files.h3x.cat/o4edvlm19)

Tatsinnit:
💡Thank you so much for this. For wider idea I think this is the issue could be relating to this behaviour (mentioned in link below) and not directly associated with this vscode extension: https://stackoverflow.com/questions/71614897/vscode-crashed-reason-oom-code-536870904

please do add replicating steps so that we can replicate if there is some indication of this extension is the cause please 🙏 

Also, kind cc: @lstocchi + @squillace and @itowlson - for more thoughts or if I missed anything. Thank you so much.

h3xcat:
@Tatsinnit There's not many steps required to replicate, just having the provided yaml code open in vscode with Kubernetes extension enabled is sufficient for memory leak to occur. It happens quite fast and crashes within 10 seconds. 

The memory leak only occurs when the Kubernetes extension is specifically enabled in unrestricted mode.

lstocchi:
Hi @h3xcat, which version of the yaml extension you have? 
are you logged in in a cluster when this happens? What if you log off and try again? 
if you add the directive `# vscode-kubernetes-tools: exclude` at the top of your yaml file what happens?

Basically, to me, this does not seem to be an issue with the yaml file directly. Maybe it triggers something else. 
The directive tells the extension not to provide any schema. It should not happen in any case as the yaml does not contain any kind/apiversion but just to do a test.
Then if you can try it with a cluster and without it to see if it triggers something else, maybe some call. Just throwing ideas on the table as i don't know what it could be atm.

h3xcat:
> which version of the yaml extension you have? 

  redhat.vscode-yaml: v1.12.2
  ms-kubernetes-tools.vscode-kubernetes-tools: v1.3.11

> are you logged in in a cluster when this happens?

  I'm logged in to docker-desktop local kubernetes cluster

> What if you log off and try again?

  Still causes a crash

> if you add the directive `# vscode-kubernetes-tools: exclude` at the top of your yaml file what happens?

  The editor no longer crashes. Removing a single character from directive instantly cause a crash again.

[...]

PR #1182, opened 2023-06-02
fix: detect errors in yaml file before actually parse it with node-yaml-parser lib (#1170)

lstocchi:
After digging a bit i found out that the node-yaml-parser is responsible of the memory leaks detected in https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1170 . If the yaml file is not formatted correctly it start looping in a while and never ends.

It would be great to have this fixed in the lib itself but it seems an abandoned project. I opened an issue 3 years ago about another bug and nobody replied. So i guess this is a fastest solution. Maybe we should find another lib to replace this one.
https://github.com/andxu/yaml-parser

So as a temporary solution, i added another parsing using the yamljs lib. If the yaml is not formatted correctly it fails and skip the node-yaml-parser, otherwise it is called and it should not fail.

The fix can be validated by using the user example at https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1170

### DIFF
--- a/src/yaml-support/yaml-locator.ts
+++ b/src/yaml-support/yaml-locator.ts
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode';
 
 import { parse, findNodeAtPosition } from 'node-yaml-parser';
+import YAML = require('yamljs');
 
 export function isMapping(node: YamlNode): node is YamlMap {
     return node.kind === 'MAPPING';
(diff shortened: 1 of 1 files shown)


Tatsinnit (approved):
☕️🎉❤️ looks nice, thank you.

Issue #1114, opened 2022-10-08
can't debug pod


GKilm:
when I try the `debug(attach) pod` function ,I have a problem to success.It is shown below.

![16652159176569](https://user-images.githubusercontent.com/28996322/194696976-1c8823a3-6243-4af7-95f1-d5a5c73c809e.png)

By the way,the extension is installed in remote-linux-machine.

lstocchi:
Hi @GKilm ,
sorry for the late response, could you please give me some more info to replicate your issue? When do you face this error? After clicking the action? If not, after which step? Do you have your project opened on vscode?
Which vscode version are you using?
I tried but i was not able to replicate it, so any help would be great. Thanks!

Tatsinnit:
💡🙏☕️Saw this and thought I will add just as fyi in case this fix #1121  takes care of the issue mentioned.

Although, as @lstocchi kindly pointed out and I am not 100 percent sure because no reproducible steps are there.

Thanks

GKilm:
Thanks for your response.
I'm sorry for not providing enough information without knowing what is important.

I'm going to show these steps with a gif.
![debug-by-k8s-tool](https://user-images.githubusercontent.com/28996322/200261580-587060cc-84f3-4b14-a843-835e7b103fd8.gif)

Here is the example project [demo](https://github.com/GKilm/test-debug-for-k8s)

The debug work is leaded by [debug-on-kubernetes.md](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/master/debug-on-kubernetes.md)

version information:
```
vscode version : 1.72.2
kubernetes version : v1.3.10
Remote-SSH version : v0.90.1
```
![vscode-verson](https://user-images.githubusercontent.com/28996322/200261508-93ad9b8c-2a30-4f9f-85a8-5254b186aa3f.png)
![k8s-tool-version](https://user-images.githubusercontent.com/28996322/200261520-446e374b-fa27-433f-a3d1-46a3a67f7602.png)

Thank you again for your contribution.If more infomation needed,notice me,please.

lstocchi:
Closing as https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1185 was merged. Please feel free to reopen it if you still face the same issue in the next release. Thanks!!

PR #1185, opened 2023-06-05
fix: debug attach not working with go app (#1184)

### LINKED ISSUE #1184: Can't debug a Go program using "Debug - Attach" menu
I tried using "Debug - Attach" menu with different Go programs and I got same error message with all of them.
Specifically, I tried it with bacongobbler/helloworld-go:1.0.0 image.
VSCode seems to connect to Dlv successfully and then I get the same error message again.

```
[Set up port forwarding 2023-06-02 17:22:53.061]
Setting up port forwarding on pod helloworld-596f8bd749-7t5kw...
Created port-forward 10000:2345 
Waiting for port forwarding to be ready...
Port forwarding is ready.
[Start debug session 2023-06-02 17:22:53.746]
Starting debug session...
```
error-message

lstocchi:
It looks like the inputs field doesn't actually work. By looking at the examples on the vscode documentation I think it should be one level up the debug configuration but it's not possible if we create it by code.

Something like
```
{ 
 type: "go",
 request: "attach",
},
inputs: {
...
}
```

but we have 
```
{ 
 type: "go",
 request: "attach",
 inputs: {
    ...
  }
},
```

So this patch adds a new step to ask the user for the remote root path.

Another thing is that it was missing the current workspace variable.
it resolves #1184 and #1114 

To test it i used the `bacongobbler/helloworld-go` project. 
So cloned the repo, init/build the app. Created a Dockerfile having the dlv to debug it. Kubernetes run to push it on cluster, then attach debug.

```
# Create build stage based on buster image
FROM golang:1.20-buster AS builder
# Create working directory under /app
WORKDIR /app

RUN go install github.com/go-delve/delve/cmd/dlv@latest
# Copy over all go config (go.mod, go.sum etc.)
COPY go.* ./
# Install any required modules
RUN go mod download
# Copy over Go source code
COPY *.go ./
# Run the Go build and output binary under hello_go_http
RUN go build -o /hello_go_http
# Make sure to expose the port the HTTP server is using
EXPOSE 8080 40000
# Run the app binary when we run the container
CMD ["dlv", "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "exec", "/hello_go_http"]
```

### DIFF
--- a/src/debug/goDebugProvider.ts
+++ b/src/debug/goDebugProvider.ts
@@ -32,22 +32,24 @@ export class GoDebugProvider implements IDebugProvider {
     }
 
     public async startDebugging(workspaceFolder: string, sessionName: string, port: number | undefined, _pod: string, _pidToDebug: number | undefined): Promise<boolean> {
+        const remotePath = await vscode.window.showInputBox({ 
+            prompt: "Which is the project root path of your remote app?", 
+            placeHolder: "/go/src/",
+            ignoreFocusOut: true,
+        });
+        if (remotePath === undefined) {
+            return false;
+        }
+        
         const debugConfiguration = {
             type: "go",
             request: "attach",
             mode: "remote",
             name: sessionName,
             hostName: "localhost",
-            remotePath: "${inputs.remotePath}",
+            remotePath,
             port,
-            inputs: [
-                {
-                    id: "remotePath",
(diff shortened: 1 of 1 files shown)


Tatsinnit (approved):
🎊❤️☕️

PR #1188, opened 2023-06-08
fix: trim container name and filter empty ones (#1187)

### LINKED ISSUE #1187: Unable to open terminal on pod
image

Noticing the issue after the recent latest update. running on windows 11 with remote SSH onto ubuntu 18.04 vm
Terminal closing with error: 
`Error from server (BadRequest): container postgres is not valid for pod postgres-89b45d47-h4gvh`

lstocchi:
it resolves #1187 

There was also an issue in the log viewer. @Tatsinnit Please give a look at everything that comes to your mind that works with containers. https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1150 added some misbehavior apparently. Hopefully I tracked all issues. If you can do some other testing it would be better.

I guess we should make a new release when this is merged.

### DIFF
--- a/src/components/logs/app/main.js
+++ b/src/components/logs/app/main.js
@@ -26,7 +26,7 @@ window.addEventListener('message', (event) => {
             const { containers, colors } = message;
             schemaColors = JSON.parse(colors);
             if (containers.length === 1) {
-                defaultContainer = containers[0];
+                defaultContainer = containers[0].name;
                 return;
             }
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1392,9 +1392,10 @@ async function getContainerQuery(resource: ContainerContainer, containerType: st
     }
     const containersEx = c.result.map((s) => {
         const bits = s.split('\t');
-        return { name: bits[0], image: bits[1], initContainer: containerType === 'initContainers'};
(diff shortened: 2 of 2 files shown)


src/components/logs/app/main.js
```diff
@@ -26,7 +26,7 @@ window.addEventListener('message', (event) => {
             const { containers, colors } = message;
             schemaColors = JSON.parse(colors);
             if (containers.length === 1) {
-                defaultContainer = containers[0];
+                defaultContainer = containers[0].name;
```
Tatsinnit, line 29: 💡 observation only, I think the `containers[0].name` is safe operation rather then checking: `if value-exist case` hence in that spirit this looks safe, please let me know if that is not the case. Looks good to me and I can cut release for this once this is merged.

Thank you. 🙏

lstocchi: I didn't get what you're trying to say 😅 you're asking if we should check if the name property exists? As i filter empty names from https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1188/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R1398 i guess that check would be superfluous.

Tatsinnit: Thank you! 🙂☕️ Sounds good, basically I was trying to archive the intention more obvious for anyone else see this in future. Sounds good. Feel free to merge it and I can cut a release again, if that sounds good.

PR #1192, opened 2023-06-19
Fix tool path configuration handling

a544jh:
Configuring custom tool paths was not working correctly.

#1181 introduced a bug which causes the `baseKey` to not be handled at all.

Also, a configuration property in vscode cannot be a string and a object (with nested properties) at the same time.

eg. if you have both keys `vscode-kubernetes.kubectl-path` and `vscode-kubernetes.kubectl-path.linux` set, vscode will not return the nested `.linux` property at all. Therefore, it's better to use properties at the same level so that the overrides can work correctly (by changing `.` to `-`).

Also corrected the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ Minikube tools to be installed and available on your PATH.
 
 ## Custom tool locations
 
-For `kubectl` and `helm`, the binaries do not need to be on the system PATH. You can configure the extension by specifying the locations using the appropriate `vs-kubernetes -> vs-kubernetes.${tool}-path` configuration setting.  See [Extension Settings](#extension-settings) below.
+For `kubectl` and `helm`, the binaries do not need to be on the system PATH. You can configure the extension by specifying the locations using the appropriate `vscode-kubernetes -> vscode-kubernetes.${tool}-path` configuration setting.  See [Extension Settings](#extension-settings) below.
 
 The extension can install `kubectl` and `helm` for you if they are missing - choose **Install dependencies** when you see an error notification for the missing tool.  This will set `kubectl-path` and `helm-path` entries in your configuration for the current OS (see "Portable extension configuration" below) - the programs will *not* be installed on the system PATH, but this will be sufficient for them to work with the extension.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -232,17 +232,17 @@ If you are working with Azure Container Services or Azure Kubernetes Services, t
 
 If you move your configuration file between machines with different OSes (and therefore different paths to binaries) you can override the following settings on a per-OS basis by appending `.windows`, `.mac` or `.linux` to the setting name:
 
-  * `vs-kubernetes.kubectl-path`
-  * `vs-kubernetes.helm-path`
-  * `vs-kubernetes.minikube-path`
+  * `vscode-kubernetes.kubectl-path`
+  * `vscode-kubernetes.helm-path`
+  * `vscode-kubernetes.minikube-path`
```
lstocchi, line 237: They should not be changed as we are talking about the `vs-kubernetes` section here. This is the old way to set 
 paths.

lstocchi (changes requested):
It would be great to see your settings. if i remember correctly before that patch the configurations outside the `vs-kubernetes` section were not taken in consideration so the new setting were not working at all.

It looks like that the WorkspaceConfiguration doesn't handle dotted names as it should.
If you use 
```
"vscode-kubernetes.kubectl-path":"<path>",
"vscode-kubernetes.kubectl-path.windows": "<path>",
```
the `windows` field is removed and only the "parent" is returned.
While if you use 
```
"vscode-kubernetes.kubectl-path.windows": "<path>",
"vscode-kubernetes.kubectl-path.linux": "<path>",
"vscode-kubernetes.kubectl-path.mac": "<path>",
```
all three are correctly detected.

lstocchi:
> So another option could be to remove the non-os specific fields completely, as the might only confuse users, and the current implementation is broken.

I would keep all of them. The same settings file could be exchanged between users/machines and they could need to setup custom paths.
Updating the name as you did solve the vscode issue so we could go that way

README.md
lstocchi: Yes, but if you want to update the readme you need to describe how it works correctly.
In the beginning it mentions that to configure the setting you have to use ` vs-kubernetes -> vs-kubernetes.${tool}-path configuration setting.`. Then you mention `vscode-kubernetes.kubectl-path`, and finally the section `vscode-kubernetes`.
So either you remove the readme change and it will be covered in another PR or you update the entire section.

lstocchi (approved):
LGTM, thanks!!

PR #1198, opened 2023-07-05
Include missing bracket pairs to helm language configuration

### LINKED ISSUE #990: Helm bracket pair colorization for helm template files
https://github.com/microsoft/vscode/issues/132240#issuecomment-953674123 has a fix to properly do bracket pair colorization via the setting `editor.language.colorizedBracketPairs`.

It would be cool if this extension could add the proper colorization as per the above comment, without users having to manually change their settings for this to work.

sergiogarciadev:
Fixes #990 

This commit will fix the cases where the open and close brackets where not of the same kind.

In Go templates, we can use `{{-` and `-}}`  to ignore the generated newlines, but those do not need to match each others, we can use `{{- ... }}` to ignore just the newline before the block and `{{ ... -}}` to ignore just the newline after the block.

`helm create` in its templates use a lot of  `{{- ... }}`, for example:

to:

A example of a range block with totally broken syntax:

to:


Tatsinnit (approved):
☕️ thank you for this PR, looks good to me, gentle ping for other eyes before we merge this. @lstocchi , @peterbom , @squillace and @gambtho

lstocchi (approved):

PR #1201, opened 2023-07-16
Added snippet for CronJob

aceat64:


lstocchi (approved):
Thanks @aceat64 . LGTM

PR #1204, opened 2023-08-23
Kill the entire process tree instead of just the process

### LINKED ISSUE #997: Extension is "leaking" kubectl processes after vscode closes
I'm using this extension as it's a dependency for _Bridge to Kubernetes_. It my current environment, whenever I close vscode, a kubeclt process is left running pointing to the workspace folder. From what I could gather, the command it's running is "proxy".

The expected behavior is to close any child processes when vscode is closed.

----
Environment:

Version:          Code - Insiders 1.63.0-insider (a7104f13ff61bcd9f4c827d8fe3c0d27bb479fd8, 2021-11-08T05:14:49.440Z)
OS Version:       Windows_NT x64 10.0.19044
Extension version: v1.3.3

mikeseese:
This PR is a follow on from #1001 to fix #997 once and for all. While #1001 closes lingering processes for traditional/standard scenarios, anyone using https://github.com/flavio/kuberlr (and by extension, Rancher Desktop) will have `kubectl` spawn a child process to `~/.kuberlr/<platform>/kubectl<version>`. NodeJS's `ChildProcess::kill()` only kills the process at hand, and does not kill child processes. [tree-kill](https://www.npmjs.com/package/tree-kill) is a simple NPM module that will call the OS-specific call to kill an entire process tree given a PID; you can see the implementation of that at https://github.com/pkrumins/node-tree-kill/blob/master/index.js?rgh-link-date=2023-08-23T06%3A00%3A23Z#L28-L49

This PR installs `tree-kill` and replaces current `ChildProcess::kill()` calls with `treeKill(process.pid, "SIGKILL")`. I opted for `SIGKILL` instead of `SIGINT` as to really just kill the process and be rid #997. If there are any concerns of not letting `kubectl` shutdown gracefully, we can switch to `SIGINT`.

CC @lstocchi


lstocchi (approved):
LGTM

Tatsinnit (approved):

Issue #1216, opened 2023-09-27
extension is crashing vscode


michaellzc:
## The problem

It appears whenever `kubernetes` extension is enabled, it will freeze vscode and potentially take down all other extensions along with it (to a point I can't even create new file or search for things in VSCode). Once disabled, there are no more errors from `Extension Host` output, all extensions are running as usual.

In debug output

```
[error] Error: spawn EBADF
    at ChildProcess.spawn (node:internal/child_process:413:11)
    at spawn (node:child_process:817:9)
    at /Users/michael/.vscode/extensions/ms-kubernetes-tools.vscode-kubernetes-tools-1.3.14/dist/extension.js:165:2130921
    at c (/Users/michael/.vscode/extensions/ms-kubernetes-tools.vscode-kubernetes-tools-1.3.14/dist/extension.js:165:2130331)
    at /Users/michael/.vscode/extensions/ms-kubernetes-tools.vscode-kubernetes-tools-1.3.14/dist/extension.js:165:2130528
    at Array.forEach (<anonymous>)
    at ChildProcess.<anonymous> (/Users/michael/.vscode/extensions/ms-kubernetes-tools.vscode-kubernetes-tools-1.3.14/dist/extension.js:165:2130462)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1091:16)
    at Socket.<anonymous> (node:internal/child_process:449:11)
    at Socket.emit (node:events:513:28)
    at Socket.emit (node:domain:489:12)
    at Pipe.<anonymous> (node:net:322:12)
2023-09-27 13:14:38.847 [info] Extension host with pid 77205 exiting with code 0
2023-09-27 13:14:39.888 [error] ReferenceError: Cannot access 'T' before initialization
    at d (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:122:56312)
    at /Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:122:54858
    at Set.forEach (<anonymous>)
    at process.<anonymous> (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:122:54847)
    at process.emit (node:events:525:35)
[...]

PR #1218, opened 2023-09-29
Revert "Kill the entire process tree instead of just the process 

Tatsinnit:
This PR is to cater and iteratively re-think #1216 the more in-depth discussion is in the thread and collective effort to come to this idea, hence opening this PR to revert 1 commit and re-release this extension.  (This reverts commit 45decd5592246fcf6313c942f0084836ca690b58.)

Thanks all so much all, ❤️🙏 cc: @mikeseese , @peterbom , @lstocchi fyi @gambtho and @squillace


mikeseese (approved):
LGTM

Issue #928, opened 2021-04-25
Terminal for Windows containers shall be launching powershell and not sh


artisticcheese:
When using Kubernetes extension with Windows containers if you choose to launch terminal against the pod, it tries to launch `sh` instead of `powershell` and hence failure below
![image](https://user-images.githubusercontent.com/13176884/116005838-53a37780-a5ce-11eb-8620-15ac0df48e74.png)

itowlson:
Our current logic is to check if `bash` is present and run that if it is, otherwise `sh`.  Clearly this isn't right for Windows containers.  I'll have a look and see how we can detect that case.  Thanks for flagging this!

PR #1273, opened 2024-05-23
Added windows container shell support

chemi0213:
Per issue https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/928 we don't handle a case for windows containers. Here is a quick fix to handle it.


Moazzem-Hossain-pixel (approved):
Thanks

Moazzem-Hossain-pixel (approved):
Thanks

Issue #1276, opened 2024-07-01
Maintenance: Enable Dependabot, stocktake the required packages.


Tatsinnit:
Thank you all who contribute, I open this issue to track and remove the unwanted, unnecessary aspect of this code base which might not be required for the package update and general health of this repo. ❤️🙏

Few things like:

* Enabling Dependabot.
* Enable security scanning. 
* Remove any half baked code/code gardening et. al. 

Update: I am currently doing this against my fork and will open the PR once I have decent drop, I see there are over 41 security alerts which we should cater. A specific area which is never developed was the `site` so I might remove it moving forwards and tag folks if anyone had any idea for the reasons et. al. Thank you so much.

Opening this as FYI. @squillace , @gambtho , @qpetraroia, @hsubramanianaks

Thanks all.

qpetraroia:
Ack on this. Thank you @Tatsinnit for maintaining this CNCF project!

Tatsinnit:
❤️ Closing this now, this is all taken care of now, please refer to the PRs above which are linked. Thank you so much.

PR #1277, opened 2024-07-02
Remove old site scaffolding project and Update/package dependencies

Tatsinnit:
Hiya All, this pull request is open with intent to cater the #1276  ❤️🙏☕️

Currently this repo do not have Dependabot and other scanning enabled, the plan is that long term it will be beneficial if we update these things and take the auto updates of packages and stay ahead of any vulnerability updates in the underlying packages.

### What does this PR caters

* Removes the old site folder which I think is never worked on and its half done.  _I do have Ivan's old knowledge recollection here_ (**Please do suggest if anyone differs and thinks they can contribute to finish the intent to maintain separate doc website via same repo**) 
   * I believe the intent was that back then where there was full team working on this that folks will create separate sites for extension and apis. The current status of this site is half done and very old packages in yarn. screenshot attached below or navigate to the link here:  https://vscode-kubernetes-tools.github.io/vscode-kubernetes-tools 
* I enabled the dependants locally at my fork and there were over 41 Security issues, with these changes inside this PR the security alerts were down to 4.
* Add Dependabot workflow.

Thank you co much all, kind cc and fyi to @squillace ,  @gambtho , @qpetraroia, @hsubramanianaks 🙏

Screenshot 2024-07-02 at 6 31 21 PM

Screenshot 2024-07-01 at 11 52 58 AM


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
```
hsubramanianaks, line 8: Can we also add grouping of PR's based ? more details here - https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups. I can do a PR quickly to update, if you are okay with it.

hsubramanianaks (approved):
Just a small suggestion to either group the PR's or limit the number of PR's opened by dependent bot so that there isn't overwhelming requests of PR's. Thank you @Tatsinnit

.github/dependabot.yml
Tatsinnit: Hiya, thank you so much for this. Absolutely, will be great to see how would you group it, usually default is single PR which kind of makes it very self contained. Because sometimes API changes are breaking and they are hard to maintain and isolate when its a 1 PR grouped for many version update. ❤️☕️

What I meant by single PR and maintenance is kind of documented in the link shared above. But yes, please add the PR and will be good to see how that will group will benefit and the advantages. 

> By default, Dependabot raises a single pull request for each dependency that needs to be updated to a newer version. You can use groups to create sets of dependencies (per package manager), so that Dependabot opens a single pull request to update multiple dependencies at the same time.

hsubramanianaks: sounds good to me, let's keep it simple without grouping the PR for now and can tackle it in future if needed. Thank you.

hsubramanianaks (approved):

qpetraroia:
@Tatsinnit, thanks for taking initiative here and maintaining this CNCF project. As a Kubernetes extension user these are all welcomed changes.

qpetraroia (approved):

PR #1279, opened 2024-07-08
Fix the ubuntu failure.

Tatsinnit:
This PR fixes the broken Ubuntu build. Might make it a part of #1276 under same umbrella.

Sample Run of this Fix against my fork is here: https://github.com/Tatsinnit/vscode-kubernetes-tools/actions/runs/9833761662/job/27144491500?pr=12

Thanks heaps. ❤️🙏 fyi : @squillace , @hsubramanianaks , @qpetraroia  fyi @gambtho

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -42,6 +42,7 @@ jobs:
     - name: Package extension
       if: github.event_name == 'push' && matrix.os == 'ubuntu-latest'
       run: |
+        npm run package
         vsce package
         mkdir vsix
         mv *.vsix vsix


.github/workflows/build-dev.yml
```diff
@@ -42,6 +42,7 @@ jobs:
     - name: Package extension
       if: github.event_name == 'push' && matrix.os == 'ubuntu-latest'
```
hsubramanianaks, line 43: @Tatsinnit I think npm run package does the same vsce package command which seems like a duplicate at line 46 and also the sample run from your fork didn't run the Package Extension step because it has to be github.event_name == 'push' - probably merging the PR will trigger this step to run. can you check on this please ?

hsubramanianaks (commented):
Hi Tats - left a comment, can you please check if your run in the forked branch did run the package extension step, as far as I checked it didn't run.

.github/workflows/build-dev.yml
Tatsinnit: This fix is focused to fix the build at the moment. The Idea was to not touch too much of workflow for the fix, and revisit all this properly. aka whatever was the behaviour it will stay as it is.

Kind of want to avoid fix vs what's the pipeline doing, please check history / commit history and the relevance to properly revisit this workflow separately. cc: @hsubramanianaks

hsubramanianaks: @Tatsinnit thanks for the clarification (Sorry I am missing the history behind it), I thought you are fixing the vsce package which got removed in previous PR's. But if this is just to fix the build, then we are good.

hsubramanianaks (approved):
lgtm

PR #1265, opened 2024-04-02
Prune and upgrade npm package dependencies

peterbom:
Updates dependencies to the latest versions, prior to enabling Dependabot to do this for us.

Change summary:
- Unused dependencies have been removed
- `@types` packages have been removed where packages publish their own types
- Node version is updated (to prevent issues where more recent packages require later node versions)
- TypeScript (and associated build/lint tools) are updated to latest versions
- Some missing dependencies have been added (`glob` and `request`)
- Additions to `tsconfig.json` to prevent TypeScript analysing the `node_modules` directory and enable the recommended `esModuleInterop` setting)
- Dependency updates and fixes for breaking changes

This [builds, runs and passes automated tests](https://github.com/peterbom/vscode-kubernetes-tools/actions/runs/8515519844), but it's probably worth doing some further manual testing.

PR #1294, opened 2024-07-19
Use `vscode-single-select` instead of `vscode-select`

### LINKED ISSUE #1253: The option to select a container when viewing logs is not available
When viewing logs, only the logs of the default container are opened. If there is more than one container running in the pod, it is not possible to view its logs.  It would be logical for there to be an option to select the container, as is done when selecting "terminal"

mikeseese:
This PR fixes #1253 by replacing all instances of `vscode-select` with `vscode-single-select`.  #1265 introduced the #1253 bug by upgrading `@bendera/vscode-webview-elements` which claims in the release notes that `0.17.2` shouldn't have Single/Multi Select until v1.0.0 in the new NPM package (https://www.npmjs.com/package/@vscode-elements/elements), but nevertheless, this PR does fix the issue.

A separate issue/effort should likely be considered to upgrade `@bendera/vscode-webview-elements@0.17.2` to `@vscode-elements/elements@latest` (where latest is `1.3.0` at the time of writing).

### DIFF
--- a/src/components/logs/app/index.html
+++ b/src/components/logs/app/index.html
@@ -5,10 +5,10 @@
     <title>Kubernetes logs</title>
     <!-- meta http-equiv="Content-Security-Policy" -->
     <style>
-        vscode-select.w-130 {
+        vscode-single-select.w-130 {
             width: 130px;
         }
-        vscode-select.w-80 {
+        vscode-single-select.w-80 {
             width: 80px;
         }
         .float-right {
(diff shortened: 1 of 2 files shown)


Tatsinnit (approved):
Thanks heaps for this. Looks good to me, ❤️🙏☕️ I am sure others will comment in case I missed anything.

Just minor thing, if we can fix the `sign-off` please.: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1294/checks?check_run_id=27651851874

mikeseese:
@Tatsinnit fixed the sign-off, sorry about that

PR #1318, opened 2024-08-22
Remove linter for not having a cpu resource limit

SonnyX:
Kubernetes documentation describe that each pod is guaranteed their CPU request amount, therefor not specifying a limit will never starve other resources. If anything specifying a limit could lead to under-utilization of available resources.

https://www.datadoghq.com/blog/kubernetes-cpu-requests-limits/

And from the kubernetes website:
image


brendandburns:
If you don't specify a limit, the Pod is free to use more CPU than it requests. If all pods use more CPU at the same time, then your system is over scheduled and all pods will slow down. Additionally anything on the system which isn't protected by the node system reserve. 

Additionally, given that it isn't part of the request, your application can't rely on being able to burst above it's request, so your system is inherently less stable since it is relying on probability for correct operation vs. absolute guarantees.

The safest thing to do is to set request and limit to the same. This ensures that a pod can only use as much CPU as it has requested. This is described in one of your links above:

https://www.datadoghq.com/blog/kubernetes-cpu-requests-limits/#predictability

Obviously, this has two downsides:
* if a Pod needs to burst it can not burst (this is the analogy one of the links above)
* you will generally under-utilize resources.

Removing the limit, and setting request, causes better utilization, but riskier operation. The linter favors stability over optimization as the default. If you know what you are doing and want to favor utilization efficiency over stability that's a more advanced option.

SonnyX:
I understand your arguments, however this should be an active choice by the user in my opinion.

In what I assume is the most common scenarios (linux machines dedicated for kubernetes) bursting would be the preferred action to ensure resource utilization. Setting any limit at all will negatively affect bursting, if a value such as 0 can be used to actively tell kubernetes & the linter that there should be no limit, then maybe we can adapt the message to give better direction whether to go for performant or safe and that 0 (or some other value) is equivalent to not setting a limit.

nathanperkins:
>The linter favors stability over optimization as the default. If you know what you are doing and want to favor utilization efficiency over stability that's a more advanced option.

Is there an option to turn this finding off? The Kubernetes maintainers themselves recommend not setting CPU limits. I don't want yellow squigglies all over my YAML when I'm following best practices and I don't want to have to add nolint comments everywhere.

Ironically, this linter doesn't insist on having CPU or memory requests, which _is_ a recommended best practice.

LTBD-Admin (approved):
Thanks

edwardgalligan:
> The linter favors stability over optimization as the default

The linter should favour established community & maintainer norms over either stability or optimization, except in extreme cases. This doesn't seem like such a case.

Issue #1357, opened 2024-10-09
Deprecate clipboardy and find replacement


hsubramanianaks:
With Recent release version 1.3.17 , we had an issue with extension not loading properly
read more about it #1355 @Tatsinnit FYIP
screenshot:
![image](https://github.com/user-attachments/assets/5e64458f-7a92-4ba7-b3dc-37163b8cfbdd)

It was due to following error from one of the dependency - clipboardy not converting the url path's based on the platform/os.
TypeError: File URL path must be absolute
	at getPathFromURLWin32 (node:internal/url:1433:11)
	at fileURLToPath (node:internal/url:1466:35)

so we have to find a replacement for this dependency which is maintained.

Tatsinnit:
💡 Totally sounds like a good idea. I think it has minimalist usage given last stable release was 4 years ago we should move to something in-house. Thanks heaps. ❤️

Moazzem-Hossain-pixel:
Help me to get

PR #1361, opened 2024-10-13
Simple clipboard implementation.

Tatsinnit:
This simple implementation is to implement the copy Kubernetes cluster name funcitonliaty.

This also caters the need mentioned in #1357 . I've test this agains my Mac, but sharing the six for folks to give some good check as well please. Thank you so much. ❤️ 

* In order to use this file, don't un-cmpress this, just remove the `.zip` from file extension and that will get you the `vsix` to test, reason being vsix is not allowed to be directly uploaded into GH hence this way to share.

* Here is the six to try it out please: [vscode-kubernetes-tools-1.3.1-testclipboard.vsix.zip](https://github.com/user-attachments/files/17352827/vscode-kubernetes-tools-1.3.1-testclipboard.vsix.zip)

Thanks heaps.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
                 "@kubernetes/client-node": "^0.22.0",
                 "@vscode/test-electron": "^2.4.1",
                 "ansi-to-html": "^0.7.2",
-                "clipboardy": "^2.3.0",
                 "compare-versions": "^6.1.1",
                 "docker-file-parser": "^1.0.7",
                 "dockerfile-parse": "^0.2.0",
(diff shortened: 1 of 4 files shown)


Moazzem-Hossain-pixel (approved):
Pls do as well

lstocchi (commented):
An enhancement could be to give some more info based on the value/action performed in the notification. In this case "cluster name copied to clipboard".

hsubramanianaks (approved):
looks good to me, works in ubuntu. Thank you @Tatsinnit 

image

PR #1369, opened 2024-10-20
Fix update with engine udpate.

Tatsinnit:
This PR is raised ignorer to properly fix dependant update: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1362 

with the vscode version update it is required to update the `engine` otherwise latter build will fail.

thank you so much.


ReinierCC (approved):
Looks good

PR #1384, opened 2024-11-07
Fix the vscode update.

Tatsinnit:
This will fix and take care of #1382 , this PR takes care of engine update with it.

thanks heaps.


ReinierCC (approved):

moazzem-pixel (commented):
Need support

Tatsinnit:fix/vscode-engine-1.95

moazzem-pixel (commented):
Thanks for helping

Issue #1373, opened 2024-10-25
problem on showing log of init container


lcasavola:
Hello,
with v.1.3.18 when i try to get log from a init container i got the following error ( more evident by using the Termina destination)
 kubectl logs pod/omnis11v2-684f969958-45wp6 --namespace=omnis-bas-001 --container=logshipper (init) -f --since=0
bash: syntax error near unexpected token `('
I think the problem is due to the fact that the container list box present the suffix "(init)" for the init container and that suffix is not excluded when you generate thecontainer name on  kubectl log command.
 IN the previous version the log for init container worked correctly.
Rgds
Luca

hime:
Having the same issue.

bcmyguest:
@Tatsinnit feel free to change my PR, I can be slow to reply on this account

hime:
Do we know what release contains this fix?

bcmyguest:
@hime No release yet :(

PR #1391, opened 2024-11-14
Fix:Init container logs not working

bcmyguest:
Continuation of changes made here: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1294

Fixes: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1373

### DIFF
--- a/src/components/logs/app/main.js
+++ b/src/components/logs/app/main.js
@@ -40,15 +40,12 @@ window.addEventListener('message', (event) => {
             for (let i = 0; i < containers.length; i += 1) {
                 const option = createElement('vscode-option', containers[i].name,
                     (containers[i].initContainer ? containers[i].name + ' (init)' : containers[i].name));
-                if (i === 0) {
-                    option.setAttribute('selected', '');
-                }
                 select.appendChild(option);
             }
             containersPanel.appendChild(select);
         }
         case 'content': {
-            const text = message.text.replace(/\n$/, '');
+            const text = message.text?.replace(/\n$/, '');
             if (!text) {
                 return;
             }
@@ -126,8 +123,9 @@ function debounce(func, wait, immediate) {
 
 function createElement(type, value, content) {
     const element = document.createElement(type);
-    if (value) { element.value = value; }
+    if (value) { element.setAttribute('value', value); }
     if (content) { element.textContent = content; }
+
     return element;
 }
 


src/components/logs/app/main.js
```diff
@@ -40,15 +40,12 @@ window.addEventListener('message', (event) => {
             for (let i = 0; i < containers.length; i += 1) {
                 const option = createElement('vscode-option', containers[i].name,
                     (containers[i].initContainer ? containers[i].name + ' (init)' : containers[i].name));
-                if (i === 0) {
```
bcmyguest, line 43: This wasn't doing anything.

src/components/logs/app/main.js
```diff
@@ -40,15 +40,12 @@ window.addEventListener('message', (event) => {
...
-                    option.setAttribute('selected', '');
-                }
                 select.appendChild(option);
             }
             containersPanel.appendChild(select);
         }
         case 'content': {
-            const text = message.text.replace(/\n$/, '');
+            const text = message.text?.replace(/\n$/, '');
```
bcmyguest, line 48: Console errors when there are no logs present.

src/components/logs/app/main.js
```diff
@@ -124,10 +121,11 @@ function debounce(func, wait, immediate) {
     };
 }
 
-function createElement(type, value, content) {
+function createElement(type, value, label) {
     const element = document.createElement(type);
-    if (value) { element.value = value; }
-    if (content) { element.textContent = content; }
+    if (value) { element.setAttribute('value', value); }
+    if (label) { element.textContent = label; }
```
bcmyguest, line 127: basically the problem here was that the 'value' attribute doesn't get set on the element. So when `vscode-single-select` goes to find the option that's selected, it can only use the content value

Moazzem-Hossain-pixel (approved):
Thanks

Moazzem-Hossain-pixel (approved):

Moazzem-Hossain-pixel (approved):

Moazzem-Hossain-pixel (approved):

bcmyguest:
Um bump?

Tatsinnit:
Thank you, @bcmyguest, for this valuable contribution!

If possible, could you kindly provide some additional context about how this addresses the tagged issues? Additionally, it would be great if you could share a VSIX file to help with replication. This would be incredibly helpful for archiving the fix in the repo and allowing community members to test it as well. ❤️ 

Thanks so much in advance for support!

bcmyguest:
Will do tomorrow on the VSIX file. 

# Explanation 
The basic trouble is that the `vscode-single-select` needs an html attribute `value` not the html property `value` to be used, otherwise it uses the content of the option. Because we are appending some additional context " (init)" to the end of the content, the value and content don't match for init containers, and the appended " (init)" is added to the log command when you run it. 

Note: this may be a bug in the library we're using because this seems like some strange behaviour to not have thought about. 

I think based on the comments the issue originated from upgrading/side grading to a new `@bendera/vscode-webview-elements` package https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1265 . While I agree with the author of https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1294 , this isn't the library you likely want to use, I just wanted to fix this small bug for now but can probably find some time to do the implementation on changing the libraries later this month or next.

Pardon the typos, on my phone.

Tatsinnit (approved):

Moazzem-Hossain-pixel (commented):
Need support



bcmyguest:fix/init-container-logs

PR #1448, opened 2025-02-03
remove acs and it's options

hsubramanianaks:
This PR removes the Azure Container Services options which no longer exists and replaced with Azure Kubernetes Services. 

Only AKS and minikube options are provided now. Please reach out if you have any questions. Thank you all. 

Tested in Ubuntu, works well cc: @qpetraroia @Tatsinnit 

image

### DIFF
--- a/src/components/clusterprovider/azure/azure.ts
+++ b/src/components/clusterprovider/azure/azure.ts
@@ -1,13 +1,13 @@
 'use strict';
 
-import { compareVersions } from 'compare-versions';
-import { Errorable, failed } from '../../../errorable';
-import { FS } from '../../../fs';
 import { Shell } from '../../../shell';
+import { FS } from '../../../fs';
+import { ActionResult, fromShellJson, fromShellExitCodeAndStandardError, fromShellExitCodeOnly, Diagnostic } from '../../../wizard';
+import { Errorable, failed } from '../../../errorable';
+import { compareVersions } from 'compare-versions';
 import { sleep } from '../../../sleep';
-import { Dictionary } from '../../../utils/dictionary';
-import { ActionResult, Diagnostic, fromShellExitCodeAndStandardError, fromShellExitCodeOnly, fromShellJson } from '../../../wizard';
 import { getKubeconfigPath } from '../../kubectl/kubeconfig';
+import { Dictionary } from '../../../utils/dictionary';
 
 export interface Context {
     readonly fs: FS;
--- a/src/components/clusterprovider/azure/azure.ts
+++ b/src/components/clusterprovider/azure/azure.ts
@@ -105,6 +105,7 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
     return fromShellExitCodeAndStandardError(sr, "Unable to set Azure CLI subscription");
 }
 
+// removed clusterType since it is NOT required after removing Azure Container Service option
 export async function getClusterList(context: Context, subscription: string): Promise<ActionResult<ClusterInfo[]>> {
     // log in
     const login = await setSubscriptionAsync(context, subscription);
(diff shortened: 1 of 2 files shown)


tejhan (approved):
LGTM 👍, tested on WSL ubuntu.

src/components/clusterprovider/azure/azure.ts
```diff
@@ -1,13 +1,13 @@
 'use strict';
 
-import { Shell } from '../../../shell';
-import { FS } from '../../../fs';
-import { ActionResult, fromShellJson, fromShellExitCodeAndStandardError, fromShellExitCodeOnly, Diagnostic } from '../../../wizard';
-import { Errorable, failed } from '../../../errorable';
 import { compareVersions } from 'compare-versions';
+import { Errorable, failed } from '../../../errorable';
```
Tatsinnit, line 4: 💡 Please note: if this PR is purely to remove `ACS` it might be worth to avoid `prettier` or indentation formationg changes so that we isolate the removal part clearly.

src/components/clusterprovider/azure/azure.ts
```diff
@@ -105,7 +105,7 @@ export async function setSubscriptionAsync(context: Context, subscription: strin
     return fromShellExitCodeAndStandardError(sr, "Unable to set Azure CLI subscription");
 }
 
-export async function getClusterList(context: Context, subscription: string, clusterType: string): Promise<ActionResult<ClusterInfo[]>> {
+export async function getClusterList(context: Context, subscription: string): Promise<ActionResult<ClusterInfo[]>> {
```
Tatsinnit, line 108: 💡 Very important note: We are changing signature of a function which is of type `export` keyword, which is used to make function available for use in other files or modules.  

So recommendation will be either leave a note here or add in description that you have tested and nowhere else this signature is called from? please. WDYT?

Although, just two answer my own question I do see that is it only used for the `az command below` but worth a quick scope check. but what I can see is it is used in `azure.ts` and `azurecloudprovider.ts`

Thanks heaps.

src/components/clusterprovider/azure/azure.ts
```diff
@@ -288,8 +264,8 @@ export async function createCluster(context: Context, options: any): Promise<Act
     };
 }
 
-export async function waitForCluster(context: Context, clusterType: string, clusterName: string, clusterResourceGroup: string): Promise<Errorable<WaitResult>> {
-    const clusterCmd = getClusterCommand(clusterType);
+export async function waitForCluster(context: Context, clusterName: string, clusterResourceGroup: string): Promise<Errorable<WaitResult>> {
```
Tatsinnit, line 267: 💡 Important note please: Same goes here: This function is consumed in 3 places jsut fyi: `miniclusterprovider.ts` is missing but if be carefully remove this or if any function with `export` needs to be kept same. signature with heavy commenting is the right approach? given timeline of codebase here? = ) I am open for quick discussion please. Thanks.

Tatsinnit (commented):
💡 Thank you so much for the kind ping and fyi for this PR, few key things to keep in mind area as follow: (Again, thank you & lets talk we are in awesome direction for this)

* Let's have a quick discussion, also, I think `Readme.md` file also need the removal of this command in that case as part of this PR.
* How we deal with method signature changes as safe changes.
* Might be avoid `formatting` changes please in major removal PR reason being it will keep that `ACS` focused changes and things will be easily more identifiable.

Thank you!

src/components/clusterprovider/azure/azure.ts
hsubramanianaks: I cross verified and added a comment, it is defined in ```azure.ts``` and being used for ```promptForCluster``` in ```azurecloudprovider.ts```. I tested both create and configure flow both looks fine to me.

Moazzem-Hossain-pixel (approved):
Thanks

src/components/clusterprovider/azure/azure.ts
hsubramanianaks: this ```waitForCluster``` is used only for step == ```wait``` and it doesn't seem like used in minikube implementation. I will test this flow to make sure it is working.

src/components/clusterprovider/azure/azure.ts
hsubramanianaks: updated.

Tatsinnit (approved):
❤️ Thanks you so much for this,  I did checked `getCredentials` and seems like these are the only places it is use. Looks good.

PR #1455, opened 2025-02-11
Feat/aks cluster

### LINKED ISSUE #1449: AKS Cluster Fix and Proposition
#### **Description**  

The **AKS Create wizard** is currently broken, preventing users from successfully provisioning AKS clusters through the wizard. This issue impacts workflow efficiency and usability.  

There is another mention here in one of the older issue: - https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1197#issuecomment-1684345898  

#### **Proposed Solution**  

Considering the work already done for **AKS Create** in the **AKS extension**, it may be beneficial to adopt and integrate that implementation into the wizard. This could help ensure consistency and reduce duplication of effort while improving the overall user experience.  

Given there is a work on those lines happening here: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1448

#### **Action Items**  

- Investigate the current issues with the **AKS Create wizard**.  
- Evaluate the feasibility of leveraging the **AKS extension’s AKS Create functionality**.  
- Explore ways to hook the two implementations together for a more streamlined experience.  

#### **Expected Benefits**  

- Improved reliability of the AKS creation process.  
- Better alignment with existing AKS extension workflows.  
- Reduced maintenance effort by reusing established implementation.  

I Would love to hear thoughts from the team on this approach! ❤ 

Thoughts and opinions? @hsubramanianaks , @qpetraroia cc: @squillace , @tejhan + @ReinierCC

hsubramanianaks:
This PR introduces changes to redirect create cluster option for Azure Kubernetes Services via AKS extension and prompt if AKS extension is not installed. 

Also introduces auth changes to get the list of subscriptions using dependency [@microsoft/vscode-azext-azureauth](https://www.npmjs.com/package/@microsoft/vscode-azext-azureauth)

Fixes #1449 

Select Cluster Type: 

image

Select Subscription:

image

AKS create Cluster Page
image

If AKS extension is not installed:
image

will install and redirect to create cluster page.


src/components/clusterprovider/azure/azure.ts
```diff
@@ -47,9 +47,14 @@ export interface WaitResult {
     readonly stillWaiting?: boolean;
 }
 
+export interface Subscription {
```
hsubramanianaks, line 50: This interface was needed since AKS extension expects subscription ID for create cluster page.

src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -34,15 +34,9 @@ function next(context: azure.Context, wizard: Wizard, action: clusterproviderreg
 
 function getHandleCreateHtml(step: string | undefined, context: azure.Context, requestData: any): Sequence<string> {
     if (!step) {
-        return promptForSubscription(requestData, context, "create", "metadata");
-    } else if (step === "metadata") {
```
hsubramanianaks, line 38: These were legacy steps not required.

src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -68,13 +62,16 @@ async function promptForSubscription(previousData: any, context: azure.Context,
...
-    const subscriptions: string[] = subscriptionList.result.result;
+    const subscriptions: azure.Subscription[] = subscriptionList.result.result;
 
     if (!subscriptions || !subscriptions.length) {
         return renderNoOptions('No Azure subscriptions', 'You have no Azure subscriptions.');
     }
 
-    const options = subscriptions.map((s) => `<option value="${s}">${s}</option>`).join('\n');
+    // sort by name
+    subscriptions.sort((a, b) => a.name.localeCompare(b.name));
```
hsubramanianaks, line 72: This is to show them sorted in the dropdown.

tejhan (approved):
Tested out & works in both cases, LGTM👍

One thing I did notice was that when creating a cluster without the extension, it opens the extension installation page in a new tab but stays on the create cluster tab which becomes blank once the user clicks 'next'. - I think it could be a slight benefit to the user if we either leave a redirect message on the original cluster-create panel, send a vscode information message, or maybe switch to the installation tab if possible.

Moazzem-Hossain-pixel (approved):
Thanks

src/components/clusterprovider/azure/azure.ts
```diff
@@ -47,9 +47,14 @@ export interface WaitResult {
...
+    name: string;
+    id: string;
+}
+
 const MIN_AZ_CLI_VERSION = '2.0.23';
 
-export async function getSubscriptionList(context: Context): Promise<ActionResult<string[]>> {
+export async function getSubscriptionList(context: Context): Promise<ActionResult<Subscription[]>> {
```
Tatsinnit, line 57: 💡 Purely for idea sake, this extension utlizes this part of code across all kind of cluster type? if so, we need to be careful which might mean we can create **`overload`** of the function and use it that way, but leaving this for thoughts and idea sake.

src/components/clusterprovider/azure/azure.ts
```diff
@@ -93,10 +98,10 @@ async function azureCliVersion(context: Context): Promise<string | null> {
     }
 }
 
-async function listSubscriptionsAsync(context: Context): Promise<Errorable<string[]>> {
-    const sr = await context.shell.exec("az account list --all --query [*].name -ojson");
+async function listSubscriptionsAsync(context: Context): Promise<Errorable<Subscription[]>> {
```
Tatsinnit, line 101: 💡 Please note: this small point here: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1455/files#r1952167946

src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -134,147 +131,17 @@ async function configureKubernetes(previousData: any, context: azure.Context): P
...
-}
-
-async function promptForAgentSettings(previousData: any, context: azure.Context): Promise<string> {
-    const vmSizes = await azure.listVMSizes(context, previousData.location);
-    if (!vmSizes.succeeded) {
-        return renderCliError('PromptForAgentSettings', {
-            actionDescription: 'listing available node sizes',
-            result: vmSizes
-        });
+async function createClusterViaAKSExtension(previousData: any) {
```
Tatsinnit, line 134: 💡 Edge case to think of: What if a user only have k8s extension and do not have vscode extension? Might be worth to please check that behaviour or do we need to explicitly invoke the dialog box to download that `aks` extension.

src/components/clusterprovider/azure/azureclusterprovider.ts
```diff
@@ -307,32 +174,6 @@ function renderConfigurationResult(configureResult: ActionResult<azure.Configure
             ${getCredsOutput}`;
 }
 
-// Error rendering helpers
-
```
Tatsinnit, line 311: 💡 Thought share: I believe this is all existing web view code for AKS only specific case and it seems obvious to remove this. (Not sure if this is used or shared across any other thing in this codebase.)

Tatsinnit (commented):
💡 Thank you so much for kind ping and this, prima facie looks chill to me, just few things I noticed and worth to test, once use case I will try and test as well is the one when I don't have aks extension and see what happens.

Thank you!

src/components/clusterprovider/azure/azureclusterprovider.ts
hsubramanianaks: removed the old files

src/components/clusterprovider/azure/azureclusterprovider.ts
hsubramanianaks: removed the old files

ReinierCC:
This addition looks great! Really like the experience.

Tested on Windows, with and without the AKS Extension.

The first time I clicked "Create Cluster" I got the following error: Screenshot 2025-02-14 at 3 25 24 PM

But I was unable to replicate the error after restarting/reinstalling many times.

ReinierCC (approved):

src/components/clusterprovider/azure/azure.ts
```diff
@@ -382,4 +382,4 @@ function getClusterCommand(): string {
 
 function getClusterCommandAndSubcommand(): string {
     return 'aks';
-}
+}
```
Tatsinnit, line 385: 💡Thoughts/Suggestions: is this all bunch of formatting changes, if so, since its very specific work, it will be worth to revert all this, especially the larger chunk which appears in the `extension.ts` ❤️

src/extension.ts
```diff
@@ -11,84 +11,84 @@ import { fs } from './fs';
...
 import * as yaml from 'js-yaml';
-import dockerfileParse = require('dockerfile-parse');
+import { pullAll } from 'lodash';
 import * as tmp from 'tmp';
 import * as clipboard from './components/platform/clipboard';
-import { pullAll } from 'lodash';
+import dockerfileParse = require('dockerfile-parse');
 
 // Internal dependencies
-import { host } from './host';
```
Tatsinnit, line 20: 💡 ❤️ Are these vscode formatting setting doing this? Reason I mention specifically is to revert these small which are not really directly associated with this specific work, so that all this doesn't appear as change under this work. Also p[lease do correct me if that is not the case? = )

src/extension.ts
```diff
@@ -2000,7 +2000,45 @@ async function configureFromClusterKubernetes() {
 
 async function createClusterKubernetes() {
     await debounceActivation();
-    runClusterWizard('Create Kubernetes Cluster', 'create');
+    // runClusterWizard('Create Kubernetes Cluster', 'create');
```
Tatsinnit, line 2003: 💡 Nit: we can safely remove this one now, as we are moving away.

src/components/cloudprovider/aksprovider.ts
```diff
@@ -0,0 +1,83 @@
+import { VSCodeAzureSubscriptionProvider } from "@microsoft/vscode-azext-azureauth";
```
Tatsinnit, line 1: 💡 ❤️ = ) Archival reasons only : https://www.npmjs.com/package/@microsoft/vscode-azext-azureauth

Tatsinnit (approved):
💡 ❤️   Looks Nice and thank you so much for chats. Thank yo use much for kind ping as well, +1 and I like the other nom inclusion for auth. Lets keep the for now. 

Regarding testing, thank you @ReinierCC & @tejhan  if you both have given green light, then I am all good for this PR to be merged.

There are few recommendation especially the reason of this work to keep it as that individual unit. 

Thanks heaps.

src/extension.ts
hsubramanianaks: Yes these are all formatting changes

src/components/cloudprovider/aksprovider.ts
hsubramanianaks: Thanks.

src/extension.ts
hsubramanianaks: updated.

src/components/clusterprovider/azure/azure.ts
hsubramanianaks: updated, please check.

PR #1456, opened 2025-02-17
Fix for broken dependabot client-node update.

Tatsinnit:
❤️ This PR is to fix #1421 dependabot fix which breaks the build - I need to upload VSIX for quick test, since it's a major version change it comes with few types change within the package. fyi @hsubramanianaks , @ReinierCC + @tejhan cc: @squillace 

Opening this for reference for now, will add vsix soon.

Thanks.


Tatsinnit:
❤️ for fellow readers only , for the version `1.0` the issue is ESM only support and here are few threads worth to understand and I guess work is already in PR to support both now.

* https://github.com/kubernetes-client/javascript/issues/2014
*  https://github.com/kubernetes-client/javascript/pull/2240 

thanks!

ReinierCC (approved):

PR #1463, opened 2025-02-24
Fix the type/nodes update.

Tatsinnit:
Hiya, This PR will fix #1458 , things need to be reviewed and tested, might be I will also add vsix. 

Thank you so much ❤️ Opening this as a reference for now, cc: @ReinierCC, @tejhan , @hsubramanianaks ,  also cc:  @squillace


ReinierCC (approved):
LGTM

PR #1471, opened 2025-03-04
Fix some scanning alerts.

Tatsinnit:
This PR is to fix few scanning alerts agains this repo, so mainly the changes are to :

1. Ensures correct domain structure: 
   - Matches `"azmk8s.io"` or `"azure.com"` **only when it appears as a full domain or subdomain** (e.g., `sub.azmk8s.io`, `xyz.azure.com`).  
   - Prevents arbitrary words like `"maliciousazmk8s.io.evil.com"` from passing.  

2. Avoids partial matches inside strings:
   - `"exampleazmk8s.io"` (will fail)
   - `"abc.azmk8s.io"` (will succeed)

Thanks heaps, fyi @ReinierCC , @tejhan, @bosesuneha, @qpetraroia , @squillace, @davidgamero

### DIFF
--- a/src/kubectlUtils.ts
+++ b/src/kubectlUtils.ts
@@ -116,7 +116,9 @@ export async function getContexts(kubectl: Kubectl, options: ConfigReadOptions):
     const cluster = clusters.find((cl) => cl.name === c.context.cluster);
 
     // Extract provider type from the cluster's server property if it includes "azk8ms"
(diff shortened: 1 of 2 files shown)


ReinierCC:
Are we able to also set a check for what comes before ".azmk8s"?

to prevent something like "evil.azmk8s.io/"

tejhan (approved):

PR #1475, opened 2025-03-12
Fix for client-node and ESM support.

Tatsinnit:
This PR addresses the issue referenced here: [#1456](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1456). Please refer to the attached discussions in the PR and the direction taken by `client-node`.  

The update transitions the extension from CommonJS to ES Modules (ESM), aligning with modern JavaScript standards for better performance, async support, and tree-shaking. Since `client-node` will no longer support CommonJS, it makes sense to move forward with ESM—unless there's a specific need to retain CommonJS for this extension.  

Opening this PR to gather broader input and thoughts. ❤️ 

Thank you and fyi and cc:  @bosesuneha , @ReinierCC , @tejhan , @squillace , @davidgamero, @qpetraroia and @gambtho


src/components/kubectl/watch.ts
```diff
@@ -1,13 +1,12 @@
...
-import { Request } from 'request';
 import { loadKubeconfig } from './kubeconfig';
 
 interface InactiveWatch {
     readonly active: false;
 }
 interface ActiveWatch {
     readonly active: true;
-    readonly request: Request;
+    readonly request: AbortController;
```
ReinierCC, line 9: is AbortController properly imported?

webpack.config.js
```diff
@@ -2,15 +2,20 @@
 
 'use strict';
 
-const path = require('path');
-const webpack = require('webpack');
+import { resolve as _resolve } from 'path';
```
ReinierCC, line 5: '_resolve' is imported but not seeing usage in file

src/components/kubectl/watch.ts
Tatsinnit: 💡 From the new `client-node` the type `AbortController` is available via `interface` type:

```*.ts
// #region borrowed
    // from https://github.com/microsoft/TypeScript/blob/38da7c600c83e7b31193a62495239a0fe478cb67/lib/lib.webworker.d.ts#L633 until moved to separate lib
    /** A controller object that allows you to abort one or more DOM requests as and when desired. */
    interface AbortController {....
```

webpack.config.js
Tatsinnit: 💡 Indeed, temp kept for line:18 sake for now. I am trying to recall or understand existence of 2 webconfig files at the moment. :)

Tatsinnit:
💡 Side note only: essentially this will not work with ESM from sample either: https://code.visualstudio.com/api/working-with-extensions/testing-extension#advanced-setup-your-own-runner

Tatsinnit:
💡 It feels like the real issue seems to be that `extensionHostProcess.js` is running in CommonJS mode and is trying to `require()` in ESM module (`index.js`).

So almost the order is:

* Test file (`index.js`) is an ESM module (`it uses import/export`).
* The VS Code extension host (`extensionHostProcess.js`) is `CommonJS` and is trying to `require()` in ESM module.
    * Node.js does not allow require() to load ESM files.

Leaving here as thought. (Because compile works fine.)

PR #1500, opened 2025-04-13
Fix: broken telemetry due to old logic in place.

Tatsinnit:
This fixes a broken telemetry for this extension to correctly enable the right data.

The term master node was deprecated and there for this change is valid, for further here is more reference to the docs please:

* https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#urgent-upgrade-notes 
* https://kubernetes.io/docs/concepts/overview/components/

Thanks heaps & a gentle fyi to cc: @qpetraroia,  @squillace , @tejhan , @ReinierCC , @davidgamero,  @bosesuneha and @gambtho thank you!

VSIX: 
[vscode-kubernetes-tools-1.3.20-test.vsix.zip](https://github.com/user-attachments/files/19727539/vscode-kubernetes-tools-1.3.20-test.vsix.zip)


davidgamero (approved):
lgtm

PR #1510, opened 2025-04-18
Add existing cluster [Fix]

tejhan:
This PR serves to fix the implementation for adding an existing cluster, now allowing users to properly add AKS clusters to their kubeconfig & interactable cluster tree in the extension.

The functionality for adding an existing cluster was not working as expected, likely due to older code / deprecated packages. Included below is documentation of observed broken behavior & new implementation for easy reference.

**Observed Behavior**

https://github.com/user-attachments/assets/cb984cf4-2e2d-477d-a91c-cb5253f4e18a

**New Behavior** 

https://github.com/user-attachments/assets/e4b97f72-3a6a-4c4f-9df3-a715630b2e92

**.vsix for testing:** [vscode-kubernetes-tools-1.3.22-addexistingcluster.vsix.zip](https://github.com/user-attachments/files/19817289/vscode-kubernetes-tools-1.3.22-addexistingcluster.vsix.zip)

**Testing instructions:** Click the 3 dots in the "Clusters" Sidebar, then select "Add Existing Cluster" -> select the cluster type (Currently AKS), then select the cluster subscription. If no clusters are found, user should be notified. If some are found, select one from the provided list & then observe if kubeconfig is properly populated with new cluster.

(Note: When developing this fix, several issues arose due to outdated / deprecated packages in the extension, so we should start phasing out all deprecated packages & update (where necessary) ones we are actively utilizing.

### DIFF
--- a/src/components/cloudprovider/aksprovider.ts
+++ b/src/components/cloudprovider/aksprovider.ts
@@ -34,7 +34,7 @@ export class AKSProvider implements CloudProvider {
         );
 
         if (!selectedSubscription) {
-            vscode.window.showErrorMessage(`Failed to get subscriptions for ${this.getName()}`);
+            vscode.window.showErrorMessage(`Failed to get subscription for ${this.getName()}`);
             return;
         }
         return subscriptions.find((sub) => sub.name === selectedSubscription)?.subscriptionId;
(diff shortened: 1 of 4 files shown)


moazzem-pixel (approved):
Thanks

src/extension.ts
```diff
@@ -68,7 +68,7 @@ import { setActiveKubeconfig, getKnownKubeconfigs, addKnownKubeconfig } from './
 import { HelmDocumentSymbolProvider } from './helm.symbolProvider';
 import { findParentYaml } from './yaml-support/yaml-navigation';
 import { linters } from './components/lint/linters';
-import { runClusterWizard } from './components/clusterprovider/clusterproviderserver';
+// import { runClusterWizard } from './components/clusterprovider/clusterproviderserver';
```
Tatsinnit, line 71: 💡 I think it might be safe to remove this if you think that it's not needed in this place.

src/components/cloudprovider/aksprovider.ts
```diff
@@ -63,6 +65,116 @@ export class AKSProvider implements CloudProvider {
...
+        }
+
+        const token = session.accessToken;
+
+        // The logic below does the same thing as the ResourceGraphClient, but we are using 
+        // the REST API directly to avoid bundling issues we currently have.
+        // After we undergo major updates 
+        // to address our outdated/broken dependencies, we can revisit this.
+
+        const url = "https://management.azure.com/providers/Microsoft.ResourceGraph/resources?api-version=2021-03-01";
```
Tatsinnit, line 89: 💡 Nice, I think your plan is to later change this REST call to the API calls.

Tatsinnit (approved):

src/extension.ts
```diff
@@ -1994,12 +1994,8 @@ async function debounceActivation(): Promise<void> {
     }
 }
 
-async function configureFromClusterKubernetes() {
-    await debounceActivation();
-    runClusterWizard('Add Existing Cluster', 'configure');
```
Tatsinnit, line 1999: 💡 Idea: Thanks Tejjan,  I wonder if we can run the new function some part into this function for refactor purpose, but jsut sharing this as idea. thanks.

tejhan: Newest commit has refactors. Thanks

Tatsinnit (approved):

PR #1515, opened 2025-04-21
Removing deprecated download package

tejhan:
This PR serves to remove & replace the `download` package and several other related sub dependencies

The `download` package we use in the extension has been deprecated, and some of our logic still relies on it. This package carries with it a large amount of additional deprecated & outdated versions of dependencies, some of them having high severity vulnerabilities, like our included versions of `got` and `http-cache-semantics`.

We utilize this function for installing dependencies such as helm & minikube, so this PR safely replaces download functionality using our already included `node-fetch library` & `fs`. 

.vsix for testing: [vscode-kubernetes-tools-1.3.22-downloadtest.vsix.zip](https://github.com/user-attachments/files/19838469/vscode-kubernetes-tools-1.3.22-downloadtest.vsix.zip)

This would effect the functionality of downloading dependencies to the vscodekubernetes path, so a good test for this would be deleting Kubectl/Helm/Minikube, then running a command such as Helm:Version which requires helm to operate, then you'll be prompted to download dependencies, which will trigger new logic & a successful installation message should be displayed once complete.


Tatsinnit (approved):

Tatsinnit:
Thank you so much for this, I've tested this and here is the working from my local. ❤️ 

Screenshot 2025-05-21 at 11 02 33 AM

src/components/installer/installdependencies.ts
```diff
@@ -44,6 +46,7 @@ async function installDependency(name: string, alreadyGot: boolean, installFunc:
         const result = await installFunc(shell);
         if (failed(result)) {
             kubeChannel.showOutput(`Unable to install ${name}: ${result.error[0]}`);
+            console.log(result.error);
```
Tatsinnit, line 49: 💡 lets remove this :) I think it was for debugging.

Tatsinnit (approved):

PR #1552, opened 2025-05-21
Work to remove deprecated request package.

Tatsinnit:
This work is for removing decprecated `request` package and replace it with `node-fetch`, intern this will help in fixing: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/security/dependabot/18 and https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/security/dependabot/17 

Thanks heaps and gentle fyi to @tejhan and @squillace ❤️ 

I have tested this change in my local and it had no adverse effect into existing aspect, but in longer run we should think of re-accessing the need of explainer.ts and might be our yaml extension dependency is enough to handle the help hover text.

Thank you so much and here is vsix for playing around.

[vscode-kubernetes-tools-1.3.23-test-pkg-deprecation.vsix.zip](https://github.com/user-attachments/files/20377654/vscode-kubernetes-tools-1.3.23-test-pkg-deprecation.vsix.zip)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
                 "lodash": "^4.17.21",
                 "mkdirp": "^3.0.1",
                 "moment": "^2.30.1",
+                "node-fetch": "^3.3.2",
                 "node-yaml-parser": "^0.0.9",
                 "opn": "^6.0.0",
                 "pluralize": "^8.0.0",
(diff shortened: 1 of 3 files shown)


squillace:
[heart]         Ralph Squillace (Naive User) reacted to your message:
________________________________

Issue #1360, opened 2024-10-10
Error about using minikube amd64 binary on arm64 system, but I don't have/use minikube


judge-red:
So everytime I start VS Code and open a file, I get the error below. The thing is, I have no minikube in my `$PATH` so I have no idea how the extension manages to find an amd64 binary to run unless it's included in the extension. I don't use minikube and I don't have any minikube path specified in the extension settings. But the error is annoying.
```
Failed to determine minikube version: Error: Unexpected output for minikube version check: ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────╮,│ │,│ You are trying to run the amd64 binary on an M1 system. │,│ Please consider running the darwin/arm64 binary instead. │,│ Download at https://github.com/kubernetes/minikube/releases/download/v1.33.1/minikube-darwin-arm64 │,│ │,╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯,CurrentVersion: v1.33.1,LatestVersion: v1.34.0
```

tejhan:
Hi @judge-red, thank you for opening this issue.

Dependencies on Mac machines are installed in the home directory, at `~/.vs-kubernetes`, so if you would like to remove unused ones such as minikube, you should be able to safely delete the .vs-kubernetes folder or the specific tools within the directory as well (`.vs-kubernetes/tools/minikube` for example. ) Once deleted, you should no longer be met with these notifications unless you install dependencies, which should only be prompted when using a helm/minikube related feature. 

It does appear as you pointed out, that the wrong binary was installed, and I've noticed an oversight in the installation logic that doesn't account for Mac machines that run arm64 under the hood but would be interpreted as amd64 from the node process logic. I've created a PR to remedy for this ensuring that future users when installing should get the correct binary on M1 or M2 machines like yourself.

Thank you for bringing this up!

judge-red:
Thanks for the tips and the PR! But just in case it matters: I'm on an M3 mac, not M1/M2 :) But `uname -m` definitely brings up `arm64` so I guess that's alright.

PR #1594, opened 2025-06-30
Detect real CPU arch on macOS to install correct binary

tejhan:
This PR fixes a bug where some Mac devices running under Rosetta (like M3 for example) were being seen as x64 via the `process.arch` function, causing the amd64 binary to be installed and fail at runtime. We now use `uname -m` on macOS to get the true architecture and ensure arm64 builds go to arm64 devices. It seems that some users had the incorrect binary silently installed, which causes a couple related errors. 

Addresses issue: 
- https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/1360

.vsix: [vscode-kubernetes-tools-1.3.24-arch1.vsix.zip](https://github.com/user-attachments/files/21022357/vscode-kubernetes-tools-1.3.24-arch1.vsix.zip)

### DIFF
--- a/src/components/installer/installationlayout.ts
+++ b/src/components/installer/installationlayout.ts
@@ -3,6 +3,7 @@ import * as path from 'path';
 
 import { fs } from '../../fs';
 import { Platform, Shell } from "../../shell";
+import * as shelljs from 'shelljs';
 
 export function vsKubernetesFolder(shell: Shell): string {
     const originalDir = path.join(shell.home(), `.vs-kubernetes`);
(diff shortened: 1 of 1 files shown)


src/components/installer/installationlayout.ts
```diff
@@ -52,6 +53,15 @@ export function formatBin(tool: string, platform: Platform): string | null {
 }
 
 export function platformArch(os: string) {
+    // on macOS ask the kernel directly
+    if (os === "darwin") {
```
Tatsinnit, line 57: 💡 Thoughts/idea: I wonder what default arch for `process.arch` condition trigger will be as scenario?

```*.ts
if (os === "darwin") {
  let arch: string;

  try {
    arch = shelljs.exec("uname -m", { silent: true }).stdout.trim();
  } catch (err) {
    // Fallback to Node's process.arch if shell command fails
    arch = process.arch;
  }

  return arch === "arm64" ? "arm64" : "amd64";
}

```

PR #1597, opened 2025-07-02
Fix the issue relating missing content in release.

Tatsinnit:
This PR is written to fix for something as the content was missing in the release.

* This PR will fix that at the time of release, so like in screenshot below, I think it will next time have a body of the release note from the `changelog` file.

* I used this opportunity to pin some of the action SHA.

Thanks all.

Screenshot 2025-07-03 at 10 14 27 AM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,10 +5,6 @@ jobs:
   build:
     name: release
     runs-on: ubuntu-latest
-    # Expose step outputs as job outputs
-    outputs:
-        currentversion: ${{ steps.package_version.outputs.current-version }}
-        changelog_reader_changes: ${{ steps.changelog_reader.outputs.changes }}
     permissions:
         actions: read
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -5,16 +5,20 @@ jobs:
   build:
     name: release
     runs-on: ubuntu-latest
+    # Expose step outputs as job outputs
+    outputs:
```
bosesuneha, line 9: I think we don't need to add outputs here unless we plan to use them in another workflow

Tatsinnit: Indeed, so idea is I plan to break this into 2 separate Hobs, build and release. (So, tis part of that kind of thinking.) = ) I would avoid massive changes under minor fix and do that separately as feature upgrade PR.

bosesuneha (approved):
lgtm

PR #1644, opened 2025-08-31
Fix security vulnerability by escaping shell arguments.

Tatsinnit:
This PR is there to fix few leftover medium security scan alerts.

1. Created a new shared utility module `shell-escape.ts` that contains the `escapeShellArg` function with proper documentation.
2. Updated `shell.ts`:
    * Added import for the shared `escapeShellArg` function
    * Removed the local duplicate function
    * All existing usage remains the same
3. Updated `wsl-fs.ts`:
    * Added import for the shared `escapeShellArg` function
    * Removed the local duplicate function
    * All existing usage remains the same

I have moved a common shared escape function into new module, I am totally fine in duplicating into separate fils as well, only one re4ason I convinced to make this function shared via new file is because of thinking in the lines of DRY Principle and for consistency as both files now use the exact same escaping logic.

Thanks heaps, gentle fyi @tejhan, @ReinierCC, @bosesuneha , @davidgamero & @qpetraroia ❤️ ❤️ cc: @squillace , @gambtho

### DIFF
--- a/src/shell.ts
+++ b/src/shell.ts
@@ -8,6 +8,7 @@ import { host } from './host';
 import { ChildProcess } from 'child_process';
 import { getKubeconfigPath } from './components/kubectl/kubeconfig';
 import { ExecResult } from './binutilplusplus';
+import { escapeShellArg } from './utils/shell-escape';
 
 export enum Platform {
     Windows,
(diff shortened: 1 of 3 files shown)


tejhan (approved):
lgmt

PR #1729, opened 2025-12-04
Adding progress indicator for helm install & auto-show output

tejhan:
Currently, `Helm: Install` does not show any indication of progress, and the output channel is also suppressed by default, leaving the user 0 feedback during installation- which can take up to 5 minutes before timing out. 

To address this, I've added both 
- `longRunning` function to indicate progress to the user while installation occurs 
- `logAndShow` command to correctly show output channel for helm install & dry-run commands both. 

These commands (`Helm Install` & `Helm: Dry-Run`) are only ever invoked explicitly by the user, so the output shouldn't be suppressed, especially for the dry-run.

### DIFF
--- a/src/helm.exec.ts
+++ b/src/helm.exec.ts
@@ -313,9 +313,11 @@ function helmInspect(arg: any, s: InspectionStrategy) {
 export function helmDryRun() {
     pickChart(async (path) => {
         const syntaxVersion = await helmSyntaxVersion();
-        const generateNameArg = (syntaxVersion === HelmSyntaxVersion.V3) ? '--generate-name' : '';
+        const generateNameArg = (syntaxVersion === HelmSyntaxVersion.V3 || syntaxVersion === HelmSyntaxVersion.V4) ? '--generate-name' : '';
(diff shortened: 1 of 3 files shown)


src/helm.exec.ts
```diff
@@ -515,15 +515,18 @@ async function helmInstallCore(kubectl: Kubectl, chartId: string, version: strin
     const nsArg = ns ? `--namespace ${ns}` : '';
     const versionArg = version ? `--version ${version}` : '';
     const generateNameArg = (syntaxVersion === HelmSyntaxVersion.V3) ? '--generate-name' : '';
-    const sr = await helmExecAsync(`install ${chartId} ${versionArg} ${nsArg} ${generateNameArg}`);
+    logger.log(`⎈⎈⎈ Installing ${chartId}...`);
```
Tatsinnit: 💡 Just a sanity ask only :) nothing user specific critical information here which we avoid to log?

tejhan: Yeah, the chartId would be user input, either typed or selected from their helm repos so we are logging information that's already generally visible.

PR #1770, opened 2025-12-28
Remove spawn-rx dependency and replace with native Node.js implementation

Tatsinnit:
This PR removes the problematic `spawn-rx` dependency and replaces it with a native Node.js `child_process.spawn` implementation that provides the same functionality. Also updates TypeScript module resolution to use the modern syntax. helps resolving issue with #1768 👮 

CC and gentle fyi to ❤️ @tejhan , @ashu8912 , @davidgamero , @bosesuneha , @ReinierCC fyi sideline cc: @squillace and @gambtho. 

### Problem

- I think the `spawn-rx` package has incorrect type definitions in its `package.json` (points to non-existent index.d.ts instead of index.d.ts)
   - This causes TypeScript compilation errors: "Cannot find module 'spawn-rx' or its corresponding type declarations"
   - The package appears to be incorrectly built/packaged

### Solution

- Replaced `spawn-rx` usage with native `child_process.spawn` in the `invokeTracking` function
- Maintained identical functionality for streaming process output
- Improved error handling and process cleanup
- Removed external dependency entirely
- Updated TypeScript config to use modern module resolution syntax - relevance - https://www.typescriptlang.org/tsconfig/#moduleResolution

### Changes

- **Removed**: `import * as spawnrx from 'spawn-rx';` 
- **Modified**: `invokeTracking` function to use `child_process.spawn` directly
- **Updated**: tsconfig.json to use `"moduleResolution": "node"` instead of deprecated `"node10"`
- **Dependency**: removed `"spawn-rx": "^6.0.0"` from package.json

Thanks all!

### DIFF
--- a/src/binutilplusplus.ts
+++ b/src/binutilplusplus.ts
@@ -270,8 +270,9 @@ export async function invokeTracking(context: Context, args: string[]): Promise<
 
         let pending = '';
         const childProcess = spawnChildProcess(bin, args, execOpts);
-        childProcess.stdout.on('data', (chunk: string) => {
-            const todo = pending + chunk;
+        childProcess.stdout.on('data', (chunk: Buffer) => {
+            const chunkText = chunk.toString('utf8');
+            const todo = pending + chunkText;
             const lines = todo.split('\n').map((l) => l.trim());
             const lastIsWholeLine = todo.endsWith('\n');
             pending = lastIsWholeLine ? '' : lines.pop()!;


package.json
```diff
@@ -1288,11 +1289,9 @@
         "rxjs": "^7.8.2",
         "semver": "^7.7.3",
         "shelljs": "^0.10.0",
-        "spawn-rx": "^5.1.2",
         "tar": "^7.5.2",
         "tmp": "^0.2.5",
         "unzipper": "^0.10.14",
-        "@vscode/extension-telemetry": "^1.2.0",
```
ashu8912: I don't understand this change why is @vscode/extension-telemetry moved??

Tatsinnit:
> Hii @Tatsinnit can you add some steps to test this change?

hey hey, sure, `observeCommand` in log.ts line `139` the observe command gets invoked, `observeCommand` ~~> `invokeTracking` (that's where impacting changes are) so whenever the `getLogsForContainer` gets called.  (LogWebview)

PR #1901, opened 2026-05-10
Ability to Restart Resources

AzhanL:
## Summary 
Currently, there is no way to quickly restart resources. The following patch adds a restart option for Deployments, StatefulSets, and DaemonSets

## Screenshots
image

image

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -634,6 +634,11 @@
                     "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource\\.(deployment|job|rc|rs|statefulset)/i"
                 },
                 {
+                    "command": "extension.vsKubernetesRestart",
+                    "group": "2@7",
+                    "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource\\.(deployment|statefulset|daemonset)/i"
+                },
+                {
                     "command": "extension.vsKubernetesLogs",
                     "group": "inline",
                     "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource\\.(pod|job)/i"
@@ -915,6 +920,11 @@
             {
                 "command": "extension.vsKubernetesScale",
                 "title": "Scale",
+                "category": "Kubernetes"
+            },
+            {
+                "command": "extension.vsKubernetesRestart",
+                "title": "Restart",
                 "category": "Kubernetes"
             },
             {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1000,8 +1000,15 @@ async function restartKubernetes(target?: any) {
 }
 
 async function invokeRestartKubernetes(kindName: string, namespace: string) {
+    const trimmedNamespace = (namespace || '').trim();
+    if (trimmedNamespace.toLowerCase() === 'all') {
+        vscode.window.showErrorMessage('Restart is not supported across all namespaces. Please select a specific namespace.');
+        return;
+    }
+
+    const namespaceArg = trimmedNamespace ? ` --namespace=${trimmedNamespace}` : '';
     const er = await host.longRunning(`Restarting ${kindName}...`, () =>
-        kubectl.invokeCommand(`rollout restart ${kindName} --namespace=${namespace}`)
+        kubectl.invokeCommand(`rollout restart ${kindName}${namespaceArg}`)
     );
     await kubectl.reportResult(er, {});
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -199,6 +199,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<APIBro
         registerCommand('extension.vsKubernetesTerminal', terminalKubernetes),
         registerCommand('extension.vsKubernetesDiff', diffKubernetes),
         registerCommand('extension.vsKubernetesScale', scaleKubernetes),
+        registerCommand('extension.vsKubernetesRestart', restartKubernetes),
         registerCommand('extension.vsKubernetesDebug', debugKubernetes),
         registerCommand('extension.vsKubernetesRemoveDebug', removeDebugKubernetes),
         registerCommand('extension.vsKubernetesDebugAttach', debugAttachKubernetes),
@@ -979,6 +980,28 @@ function promptScaleKubernetes(kindName: string) {
 async function invokeScaleKubernetes(kindName: string, replicas: number) {
     const er = await host.longRunning(`Scaling ${kindName} to ${replicas} replicas...`, () =>
         kubectl.invokeCommand(`scale --replicas=${replicas} ${kindName}`)
+    );
+    await kubectl.reportResult(er, {});
+}
+
+async function restartKubernetes(target?: any) {
+    if (target && explorer.isKubernetesExplorerResourceNode(target)) {
+        const kindName = target.kindName;
+        const namespace = target.namespace || await kubectlUtils.currentNamespace(kubectl);
+        await invokeRestartKubernetes(kindName, namespace);
+    } else {
+        const isMinimalWorkflow = config.isMinimalWorkflow();
+        const kindName = await findKindNameOrPrompt(kuberesources.restartableKinds, 'restart', { skipFreeTextPrompt: isMinimalWorkflow });
+        if (kindName) {
+            const namespace = await kubectlUtils.currentNamespace(kubectl);
+            await invokeRestartKubernetes(kindName, namespace);
+        }
+    }
+}
+
+async function invokeRestartKubernetes(kindName: string, namespace: string) {
+    const er = await host.longRunning(`Restarting ${kindName}...`, () =>
+        kubectl.invokeCommand(`rollout restart ${kindName} --namespace=${namespace}`)
     );
     await kubectl.reportResult(er, {});
 }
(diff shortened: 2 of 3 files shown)


tejhan (commented):
Thanks for submitting this.

There are a few open comments from Copilot, can you please address them?

Thank You.

src/extension.ts
```diff
@@ -983,6 +984,34 @@ async function invokeScaleKubernetes(kindName: string, replicas: number) {
...
+    const trimmedNamespace = (namespace || '').trim();
+    if (trimmedNamespace.toLowerCase() === 'all') {
+        vscode.window.showErrorMessage('Restart is not supported across all namespaces. Please select a specific namespace.');
+        return;
+    }
+
+    const namespaceArg = trimmedNamespace ? ` --namespace=${trimmedNamespace}` : '';
+    const er = await host.longRunning(`Restarting ${kindName}...`, () =>
+        kubectl.invokeCommand(`rollout restart ${kindName}${namespaceArg}`)
+    );
```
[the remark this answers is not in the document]
AzhanL: This suggestion is invalid because the `kubectl.invokeCommand` does not support passing arguments as an array. Other usages also use the command by string concatenation. Changing the method parameter would be a breaking change

image

src/extension.ts
```diff
@@ -983,6 +984,34 @@ async function invokeScaleKubernetes(kindName: string, replicas: number) {
...
+        if (kindName) {
+            const namespace = await kubectlUtils.currentNamespace(kubectl);
+            await invokeRestartKubernetes(kindName, namespace);
+        }
+    }
+}
+
+async function invokeRestartKubernetes(kindName: string, namespace: string) {
+    const trimmedNamespace = (namespace || '').trim();
+    if (trimmedNamespace.toLowerCase() === 'all') {
```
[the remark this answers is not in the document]
AzhanL: Resolved with b

AzhanL:
Handling of `all` namespaces should not be done here and is a latent bug in the project that should noted as an issue. The only place where it is handled is in [getPods](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/blob/73e8c0fafb8ac9f8ab2e25c23bdebb1d5e4fea3f/src/kubectlUtils.ts#L312) at the moment

AzhanL:
@tejhan  this is ready for review. Let me know if you find anything else

Tatsinnit (commented):
Thank you so much for this contribution, @AzhanL 🙏  and thanks for your patience through the review rounds. This is a genuinely useful gap-fill: I do feel that `rollout restart` is one of the most common day-2 operations (picking up a rotated Secret/ConfigMap, re-pulling a mutable image tag, clearing a wedged pod), and today it forces users out to a terminal. The implementation also mirrors the existing `scaleKubernetes` shape closely, which makes it easy to reason about. Really appreciate it.

I've left a few small, non-blocking thoughts inline, mostly around safety-of-default and post-action UX rather than correctness. The one I'd gently push on before merge is the confirmation prompt (comment on `restartKubernetes`), since this lands as a one-click context-menu item directly beneath Scale.

Also, you're completely right in [your comment](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1901#issuecomment-5081896021) that the `all` namespaces handling is a pre-existing, project-wide gap and not something this PR should have to solve. Would you mind opening a tracking issue for it so it doesn't get lost? Happy to label it `good first issue`.

Thanks again!

src/extension.ts
```diff
@@ -983,6 +984,30 @@ async function invokeScaleKubernetes(kindName: string, replicas: number) {
     await kubectl.reportResult(er, {});
 }
 
+async function restartKubernetes(target?: any) {
```
Tatsinnit, line 987: Nit / the one thing I'd like to land before merge: could we add a confirmation prompt here?

This surfaces as a single-click context-menu entry sitting immediately below **Scale**, so a misclick triggers a rolling restart of what may be a production Deployment, with no undo. `deleteKubernetes` already gates on a confirmation, and I think restart deserves the same treatment.

There's a ready-made module-local helper at `warnConfirm` (src/extension.ts:1615), already used for the namespace-deletion guards, so this is a small addition:

```ts
const confirmed = await warnConfirm(
    `Do you want to restart '${kindName}' in namespace '${namespace}'? This will roll all of its pods.`,
    "Restart", "Cancel");
if (!confirmed) { return; }
```

(The `deleteMessageItems` pattern used at line 1552 would work equally well if you prefer consistency with delete.)

Very happy to be overruled if you feel it adds too much friction for a routine action but given it's irreversible, I lean toward asking.

src/extension.ts
```diff
@@ -983,6 +984,30 @@ async function invokeScaleKubernetes(kindName: string, replicas: number) {
...
+    if (target && explorer.isKubernetesExplorerResourceNode(target)) {
+        const kindName = target.kindName;
+        const namespace = target.namespace || await kubectlUtils.currentNamespace(kubectl);
```
Tatsinnit: Small observation, not a change request: nice defensive fallback. Worth noting `ResourceNode.namespace` reads straight from `metadata.namespace` (src/components/clusterexplorer/node.resource.ts:25), which is genuinely absent for cluster-scoped resources, so the `||` is doing real work here. Good catch.

src/extension.ts
```diff
@@ -983,6 +984,30 @@ async function invokeScaleKubernetes(kindName: string, replicas: number) {
...
+        const namespace = await kubectlUtils.currentNamespace(kubectl);
+        const kindName = await findKindNameOrPrompt(kuberesources.restartableKinds, 'restart', { skipFreeTextPrompt: config.isMinimalWorkflow() });
+        if (kindName) {
+            await invokeRestartKubernetes(kindName, namespace);
+        }
+    }
+}
+
+async function invokeRestartKubernetes(kindName: string, namespace: string) {
+    const trimmedNamespace = namespace.trim();
```
Tatsinnit: Minor: `currentNamespace()` (src/kubectlUtils.ts:348) only ever returns `""` when the kubeconfig can't be read, otherwise it falls back to `"default"`. So the `trimmedNamespace ? ... : ''` branch is close to unreachable in practice.

Not worth churning the PR over, but there's a helper for exactly this: `kubectlUtils.currentNamespaceArg(kubectl)` (src/kubectlUtils.ts:361) already does the empty-check and returns the formatted flag. Might be a small simplification if you'd like it — entirely your call.

src/extension.ts
```diff
@@ -983,6 +984,30 @@ async function invokeScaleKubernetes(kindName: string, replicas: number) {
...
+
+    const namespaceArg = trimmedNamespace ? ` --namespace=${trimmedNamespace}` : '';
+    const er = await host.longRunning(`Restarting ${kindName}...`, () =>
+        kubectl.invokeCommand(`rollout restart ${kindName}${namespaceArg}`)
+    );
+    await kubectl.reportResult(er, {});
```
Tatsinnit, line 1008: Suggestion: could we refresh the cluster explorer on success? After a restart the pod list underneath the workload is immediately stale, and users will likely expect the tree to catch up on its own.

`refreshExplorer` is already imported in this file (line 62) and used the same way after other mutating commands (e.g. lines 1543, 2212), so no new imports needed:

```ts
await kubectl.reportResult(er, {});
if (ExecResult.succeeded(er)) {
    refreshExplorer();
}
```

(Note `treeProvider.refresh()` isn't reachable from here — it's a `const` local to `activate()` — so `refreshExplorer()` is the right call.)

package.json
```diff
@@ -633,6 +633,11 @@
                     "group": "2@6",
                     "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource\\.(deployment|job|rc|rs|statefulset)/i"
                 },
+                {
+                    "command": "extension.vsKubernetesRestart",
+                    "group": "2@7",
+                    "when": "view == extension.vsKubernetesExplorer && viewItem =~ /vsKubernetes\\.resource\\.(deployment|statefulset|daemonset)/i"
+                },
```
Tatsinnit, line 640: This all looks correct to me — I traced it through: `contextValue` is built as `` `vsKubernetes.resource.${this.kind.abbreviation}` `` (src/components/clusterexplorer/node.resource.ts:48) and the abbreviations are `deployment` / `statefulset` / `daemonset`, so the regex matches. Slotting at `2@7` right after Scale is the right neighbourhood too. 👍

src/kuberesources.ts
```diff
@@ -46,6 +46,12 @@ export const scaleableKinds = [
     allKinds.statefulSet,
 ];
 
+export const restartableKinds = [
+    allKinds.deployment,
+    allKinds.statefulSet,
+    allKinds.daemonSet,
+];
```
Tatsinnit, line 53: Nicely done: mirrors `scaleableKinds` exactly, and restricting to the three genuine rolling-restart-capable workload controllers is the correct scope. No notes.

Tatsinnit:
One last small housekeeping request, sorry: the branch currently carries 7 commits, several of them `Potential fix for pull request finding` autofix commits from the review rounds. Could you squash before we merge, so `main`'s history stays tidy? Purely cosmetic (But just. note, we can squash merge at our end), thanks again for sticking with this one. 🙏

AzhanL:
Created https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/2005 for namespace handling issue

src/extension.ts
AzhanL: Definitely agree; won't hurt to add confirmation, especially if it's a large deployment

src/extension.ts
AzhanL: Makes sense, simplifies the namespace handling as well

AzhanL:
@Tatsinnit, thanks for the review. I've made the suggested changes. Squashed and pushed. Let me know if anything else needs changes

Tatsinnit (approved):

PR #1923, opened 2026-05-28
Updating auth prompt detection to match additional known login URLs

tejhan:
The authentication prompt detection added in #1795 only matched the `microsoft.com/devicelogin` URL. Newer versions of kubelogin and Azure CLI may use `login.microsoft.com/device` or `login.microsoftonline.com/common/oauth2/deviceauth`, causing the extension to hang indefinitely when Azure authentication is required.

This updates the regex pattern to cover all known device login URL variants:
- `microsoft.com/devicelogin` (legacy)
- `login.microsoft.com/device` (current)
- `login.microsoftonline.com/common/oauth2/deviceauth`

Fixes infinite loading when expanding namespaces or other resources on Azure clusters that require re-authentication.

### DIFF
--- a/src/shell.ts
+++ b/src/shell.ts
@@ -175,8 +175,10 @@ function execCore(cmd: string, opts: any, callback?: ((proc: ChildProcess) => vo
     });
 }
 
-// Pattern to detect Azure AD device login prompts
-const AUTH_PROMPT_PATTERN = /microsoft\.com\/devicelogin/i;
+// Pattern to detect Azure AD device login prompts (covers multiple URL variants)
+// Known URLs: microsoft.com/devicelogin, login.microsoft.com/device, login.microsoftonline.com/.../oauth2/deviceauth
+// Requires https:// prefix to avoid matching substrings in unrelated URLs
+const AUTH_PROMPT_PATTERN = /https:\/\/(?:[a-z0-9-]+\.)?microsoft(?:online)?\.com\/(?:device|devicelogin|common\/oauth2\/deviceauth)/i;
 
 let authNotificationShown = false;
 


Tatsinnit (dismissed):

src/shell.ts
```diff
@@ -175,8 +175,9 @@ function execCore(cmd: string, opts: any, callback?: ((proc: ChildProcess) => vo
     });
 }
 
-// Pattern to detect Azure AD device login prompts
-const AUTH_PROMPT_PATTERN = /microsoft\.com\/devicelogin/i;
+// Pattern to detect Azure AD device login prompts (covers multiple URL variants)
+// Known URLs: microsoft.com/devicelogin, login.microsoft.com/device, login.microsoftonline.com/.../oauth2/deviceauth
+const AUTH_PROMPT_PATTERN = /microsoft(?:online)?\.com\/(?:device|devicelogin|common\/oauth2\/deviceauth)/i;
```
Tatsinnit: This regex is a good start, but it is not fully safe for security-sensitive validation.

Current pattern:

```js
const AUTH_PROMPT_PATTERN =
  /microsoft(?:online)?\.com\/(?:device|devicelogin|common\/oauth2\/deviceauth)/i;
```

Potential issue: it can also match malicious strings like:

```txt
evilmicrosoftonline.com/devicelogin
attacker.com/?next=microsoftonline.com/devicelogin
foo.microsoftonline.com.evil.org/device
```

because the regex is:
- not anchored
- not validating hostname boundaries
- matching anywhere in the string

Safer alternative:

```js
const AUTH_PROMPT_PATTERN =
  /^https:\/\/([a-z0-9-]+\.)?microsoft(?:online)?\.com\/(?:device|devicelogin|common\/oauth2\/deviceauth)(?:\/|$)/i;
```

We could latter on change is to helper function and use the `URL` parser and validate:
- exact hostname allowlist
- pathname separately

That is generally more robust than regex-only URL validation.

tejhan: The regex is unanchored intentionally. The pattern appears midline within the string (For ex: `To sign in, use a web browser to open the page https://login.microsoft.com/device...`) , a pattern at a variable location within the string is what's expected. Anchoring with `^` would break detection completely. 

On the point of not validating hostname boundaries, this is valid, I've updated the regex to ensure we are specifically capturing Microsoft domains.

And as stated above, matching anywhere in the string is intentional & necessary for our flow. We're only surfacing the authentication link, which we have to find with a general regex pattern since we know variability exists in link location & type but our pattern allows all those cases to be handled, that's why these extra two were added. 

It's worth noting that this isn't a security-sensitive validation- it's a UX detection on a trusted subprocess that we spawned ourselves. That said, the tightened pattern is better, so I've applied changes in amended commit.

Tatsinnit (approved):

PR #1941, opened 2026-06-08
Add resource name filter to the Clusters tree view

Bogghi:
## Summary

Adds a **name filter** to the **Clusters** tree view so users can quickly narrow long resource lists (Pods, Services, Ingresses, Deployments, …) instead of scrolling.

A **filter** icon in the Clusters view title bar opens an input box; typing matches resource leaf nodes by name (case-insensitive substring). While a filter is active, a **clear** icon appears next to it and the view shows a `Filtering resources by: <text>` header. Folders and the cluster structure stay intact — only the resource *items* inside them are filtered.

## Motivation

Real clusters routinely list dozens or hundreds of resources of a single kind. The Clusters explorer had no way to filter them, so locating a specific pod/ingress/etc. meant scrolling through the whole list.

VS Code's extension API does not allow docking a persistent text input above a native `TreeView`, so this uses the idiomatic pattern — a title-bar button that drives the filter — which keeps all native tree behavior (context menus, watch/refresh, reveal).

## What changed

- **`src/components/clusterexplorer/explorer.ts`** — `KubernetesExplorer` now holds an optional `resourceNameFilter`. The filter is applied in `getChildren()` (the single point all nodes pass through): only nodes with `nodeType === resource` are matched against the needle; folders, contexts, config items, helm and message nodes always pass. New `getResourceNameFilter()` / `setResourceNameFilter()` methods set the filter, toggle a `setContext` key, surface the active filter as the tree view's header message, and refresh.
- **`src/utils/providerresult.ts`** — Added a `filter()` helper that mirrors the existing `map()`, handling the array-or-`Thenable` duality of `ProviderResult<T[]>`.
- **`src/extension.ts`** — Registered two commands: `extension.vsKubernetesFilterExplorer` (opens the input box, pre-filled with the current filter) and `extension.vsKubernetesClearFilterExplorer`.
- **`package.json`** — Declared both commands (codicons `$(filter)` / `$(clear-all)`, no new image assets) and added them to the Clusters `view/title` `navigation` group. The clear button is gated on `extension.vsKubernetesExplorerFilterActive` so it only shows while filtering.

## Behavior notes

- Matching is case-insensitive substring (e.g. `tecno` matches `tecnocasa-portale-dev-0`).
- The filter is provider-wide (applies across all listed contexts) and is re-applied on every refresh, including watch-driven refreshes.
- Submitting an empty input box, or pressing the clear button, removes the filter; pressing Esc leaves it unchanged.

## Testing

- `npm install`
- `tsc -p ./ --noEmit` → no type errors
- `eslint ./src` on the touched files → clean
- Manually in the Extension Development Host: expanded a cluster → Workloads → Pods, used the filter icon to narrow the list, verified folders remained, the header message appeared, and the clear button reset the view. Confirmed non-resource items (e.g. ConfigMap keys, pod status info lines) are unaffected, and that the filter persists across refresh.

## Out of scope (possible follow-ups)

- Per-resource-kind filtering (current design is provider-wide).
- Fuzzy matching (the `fuzzysearch` dependency is available; substring chosen for predictability).
- Auto-hiding empty folders / auto-expanding folders that contain matches (would require eager fetching across folders).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -452,6 +452,16 @@
                     "group": "navigation"
                 },
                 {
+                    "command": "extension.vsKubernetesFilterExplorer",
+                    "when": "view == extension.vsKubernetesExplorer",
+                    "group": "navigation"
+                },
+                {
+                    "command": "extension.vsKubernetesClearFilterExplorer",
+                    "when": "view == extension.vsKubernetesExplorer && extension.vsKubernetesExplorerFilterActive",
+                    "group": "navigation"
+                },
+                {
                     "command": "extension.vsKubernetesCreateCluster",
                     "when": "view == extension.vsKubernetesExplorer",
                     "group": "0"
@@ -1030,6 +1040,18 @@
                     "light": "images/light/refresh.svg",
                     "dark": "images/dark/refresh.svg"
                 }
+            },
+            {
+                "command": "extension.vsKubernetesFilterExplorer",
+                "title": "Filter Resources by Name",
+                "category": "Kubernetes",
+                "icon": "$(filter)"
+            },
+            {
+                "command": "extension.vsKubernetesClearFilterExplorer",
+                "title": "Clear Resource Filter",
+                "category": "Kubernetes",
+                "icon": "$(clear-all)"
             },
             {
                 "command": "extension.vsKubernetesShowEvents",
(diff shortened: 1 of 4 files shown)


tejhan:
Right now this filter applies to every kind in the tree, not just the list you're on. If you type a pod name, it also culls namespaces, services, nodes, etc. The idea's heading in the right direction, but I think a resource-scoped filter (applied to the resource kind you invoke it from) would match user expectations better than a global catch-all.

PR #1969, opened 2026-07-13
config option suppress missing kubeconfig

davidgamero:
## Summary

Adds a config option to disable the startup "Kubeconfig not found … Add a new one?" alert that requires acknowledgement.

Some users intentionally reference kubeconfig paths that don't always exist on disk (e.g. transient/generated configs, remote or on-demand setups). For them this startup prompt is noise that must be dismissed repeatedly. This lets them opt out.

## Change

New boolean setting **`vs-kubernetes.suppress-kubeconfig-not-found-alerts`** (default `false`), following the existing `suppress-*-not-found-alerts` pattern (e.g. helm).

- `package.json` — contributes the new configuration option.
- `src/components/config/config.ts` — adds `suppressKubeconfigNotFound()` / `setSuppressKubeconfigNotFound()` accessors.
- `src/extension.ts` — when the setting is enabled, suppresses the startup **"Kubeconfig not found at: … Add a new one?"** prompt in `validateKubeconfigPath()`.

Default behavior is unchanged; the prompt still appears unless a user explicitly enables the setting.

## Notes

- Scoped to feature-only changes — earlier formatting churn and an unrelated dependency `overrides` block were removed so the diff reflects just this feature (3 files, +16/-1).
- All commits are DCO signed-off.

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,6 +214,11 @@
                             "default": false,
                             "description": "Suppresses the display of warnings if no helm binary is available"
                         },
+                        "vs-kubernetes.suppress-kubeconfig-not-found-alerts": {
+                            "type": "boolean",
+                            "default": false,
+                            "description": "Suppresses the display of warnings if no kubeconfig file is available"
+                        },
                         "vs-kubernetes.ignore-recommendations": {
                             "type": "boolean",
                             "default": false,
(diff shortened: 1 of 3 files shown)


tejhan (approved):
LGTM. 

If we think this is an option that a decent portion of our users might like, we can also add it to our `README.md` so that it's documented. 

Otherwise, I think this is fine as a semi-hidden config setting as is.

PR #2007, opened 2026-08-10
fix: do not block activation on the kubeconfig-not-found prompt (#1999)

### LINKED ISSUE #1999: latest version 1.132 clashes with kubernetes
the attached cluster info receives the error: extension.vsKubernetesRefreshExplorer: command 'extension.vsKubernetesRefreshExplorer' not found. This is likely caused by the extension that contributes extension.vsKubernetesRefreshExplorer.

no matter how many time i refreshed. 

downgrading to ver 1.131 solved the issue. please investigate

Tatsinnit:
Fixes #1999

## Summary

On VS Code **1.132.0** the extension never finishes activating, so none of its commands exist. Every action fails with:

> `command 'extension.vsKubernetesRefreshExplorer' not found. This is likely caused by the extension that contributes extension.vsKubernetesRefreshExplorer.`

Refreshing does not help, because the command is never registered in the first place. Downgrading VS Code to 1.131.0 hides the problem.

Note the versions in the issue title are **VS Code** versions, not extension versions.

## Root cause

`activate()` began with:

```ts
export async function activate(context: vscode.ExtensionContext): Promise<APIBroker> {
    await validateKubeconfigPath();
    ...
```

`validateKubeconfigPath()` shows a notification when the kubeconfig is missing:

```ts
const choice = await vscode.window.showWarningMessage(
    `Kubeconfig not found at: ${path}. Add a new one?`,
    'Add',
    'Cancel'
);
```

A VS Code notification with buttons only settles once the user clicks one or dismisses it. If the user ignores it, that `await` never returns, `activate()` stays pending forever, and the ~120 `registerCommand` calls further down the function (plus the tree data providers, file system providers and status bar items) never run.

`vs-kubernetes.suppress-kubeconfig-not-found-alerts`, added in 1.4.1, lets a user opt out of the prompt, but it does not fix the ordering — anyone who hits the prompt still gets a dead extension.

`getKubeconfigPath()` can also throw (for example when the WSL shell-out fails), and that was likewise unguarded on the activation path.

## Reproduction

Same extension code, same conditions, only the VS Code version differs:

| VS Code | `activate()` | `extension.vsKubernetesRefreshExplorer` registered |
| --- | --- | --- |
| 1.131.0 | completes in ~6s | yes |
| 1.132.0 | never resolves (30s timeout) | no |

## Fix

Move the kubeconfig check off the activation critical path: run it after everything has been registered, do not await it, and swallow its errors so it can never fail activation.

```ts
    recommendExtensions(kubectl, context);

    // The kubeconfig check can prompt the user, and a VS Code notification with buttons only
    // resolves once the user responds to (or dismisses) it. ...
    validateKubeconfigPath().catch((err) => {
        console.warn(`Error validating kubeconfig path: ${err}`);
    });
```

The user-facing behaviour is unchanged — the prompt still appears, and choosing **Add** still opens the kubeconfig picker. It just no longer holds activation hostage.

Also fixes the indentation inside that `if` block, which was misleading.

## Tests

New regression test `test/suite/activation.test.ts` stubs `showWarningMessage` so it never settles (a user who ignores the notification), points `KUBECONFIG` at a missing file, and asserts that `activate()` completes and the refresh command is registered.

Reverting the source fix makes it fail with:

```
AssertionError [ERR_ASSERTION]: activate() did not complete - it is blocked on a user prompt
```

Two supporting changes were needed to make activation testable at all:

* `test/runTest.ts` installs `redhat.vscode-yaml` into the test instance. It is declared in `extensionDependencies`, and without it VS Code refuses to activate the extension: `Cannot activate the 'Kubernetes' extension because it depends on unknown extension 'redhat.vscode-yaml'`. No existing test activated the extension, so this had gone unnoticed.
* `test/runTest.ts` honours `VSCODE_TEST_VERSION` so a specific VS Code build can be pinned, e.g. `VSCODE_TEST_VERSION=1.132.0 npm test`. Defaults to `stable`, so CI is unaffected.
* `kubeconfig-merge.test.ts` only registers its stand-in refresh command when the real extension has not already registered it — registering the same command twice throws.

## Verification

* `VSCODE_TEST_VERSION=1.132.0 npm test` — 59 passing
* `VSCODE_TEST_VERSION=1.131.0 npm test` — 59 passing
* `npm run compile` and `npm run lint` clean

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -135,7 +135,6 @@ export const HELM_TPL_MODE: vscode.DocumentFilter = { language: "helm", scheme: 
 // this method is called when your extension is activated
 // your extension is activated the very first time the command is executed
 export async function activate(context: vscode.ExtensionContext): Promise<APIBroker> {
-    await validateKubeconfigPath();
     setAssetContext(context);
 
     await fixOldInstalledBinaryPermissions(shell);
@@ -387,6 +386,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<APIBro
     vscode.workspace.registerTextDocumentContentProvider(configmaps.uriScheme, configMapProvider);
 
     recommendExtensions(kubectl, context);
+
+    // The kubeconfig check can prompt the user, and a VS Code notification with buttons only
+    // resolves once the user responds to (or dismisses) it. Awaiting that during activation
+    // leaves activate() pending forever, so no commands - including
+    // extension.vsKubernetesRefreshExplorer - ever get registered. Kick it off in the
+    // background instead, and never let it fail activation. See issue #1999.
+    validateKubeconfigPath().catch((err) => {
+        console.warn(`Error validating kubeconfig path: ${err}`);
+    });
 
     return apiBroker(clusterProviderRegistry, kubectl, portForwardStatusBarManager, treeProvider, cloudExplorer, localTunnelDebugger, onDidChangeKubeconfigEmitter, activeContextTracker, kubectlUtils.onDidChangeNamespaceEmitter);
 }
@@ -2182,15 +2190,15 @@ async function validateKubeconfigPath() {
         // suppressKubeconfigNotFound (vs-kubernetes.suppress-kubeconfig-not-found-alerts)
         // hides this startup "add a new one?" prompt when enabled.
         if (!exists && !config.suppressKubeconfigNotFound()) {
-        const choice = await vscode.window.showWarningMessage(
-            `Kubeconfig not found at: ${path}. Add a new one?`,
-            'Add',
-            'Cancel'
-        );
-        if (choice === 'Add') {
-            await useKubeconfigKubernetes();
-        }
-        break;
+            const choice = await vscode.window.showWarningMessage(
+                `Kubeconfig not found at: ${path}. Add a new one?`,
+                'Add',
+                'Cancel'
+            );
+            if (choice === 'Add') {
+                await useKubeconfigKubernetes();
+            }
+            break;
         }
     }
 }
--- a/test/suite/activation.test.ts
+++ b/test/suite/activation.test.ts
@@ -1,0 +1,66 @@
+import * as assert from 'assert';
+import * as os from 'os';
+import * as path from 'path';
+import * as sinon from 'sinon';
+import * as vscode from 'vscode';
+
+const EXTENSION_ID = 'ms-kubernetes-tools.vscode-kubernetes-tools';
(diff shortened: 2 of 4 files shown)


src/extension.ts
```diff
@@ -388,6 +387,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<APIBro
 
     recommendExtensions(kubectl, context);
 
+    // The kubeconfig check can prompt the user, and a VS Code notification with buttons only
+    // resolves once the user responds to (or dismisses) it. Awaiting that during activation
+    // leaves activate() pending forever, so no commands - including
+    // extension.vsKubernetesRefreshExplorer - ever get registered. Kick it off in the
+    // background instead, and never let it fail activation. See issue #1999.
+    validateKubeconfigPath().catch((err) => {
```
Tatsinnit: Could we refresh kubeconfig-derived state when the background prompt successfully selects a file? The current namespace and YAML schema support are initialized before this call, but `useKubeconfigKubernetes()` currently only updates the setting, fires `onDidChangeKubeconfigEmitter`, and invalidates telemetry. After choosing **Add**, the namespace status can remain blank or stale, while context and schema state are delayed until the context poll. The old ordering completed kubeconfig selection before those initial reads.

test/suite/activation.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        assert.ok(extension, `extension ${EXTENSION_ID} not found`);
+
+        const TIMED_OUT = Symbol('timed out');
+        let timer: NodeJS.Timeout | undefined;
+        const timeout = new Promise<symbol>((resolve) => {
+            timer = setTimeout(() => resolve(TIMED_OUT), 30000);
+        });
+
+        try {
+            const outcome = await Promise.race([extension.activate(), timeout]);
```
Tatsinnit: Could we make this test guarantee that it exercises the unresolved-prompt path? CI runs `YAML consideration-filter` first, and that suite opens a YAML document, which activates this extension through `onLanguage:yaml`. By the time this calls `extension.activate()`, activation may already be complete, so neither the `KUBECONFIG` override nor the warning stub participates; the test can therefore pass even if the production fix is reverted. Please run this case in a fresh extension host, or otherwise assert that the extension starts inactive and that the kubeconfig warning stub is actually called.

Tatsinnit:
The core activation fix looks good, and the rest of the changes look reasonable. I left two focused comments around deterministic regression coverage and refreshing kubeconfig-derived state after the background prompt.

FYI @bosesuneha @davidgamero @yolossn

PR #2008, opened 2026-08-16
fix: honour the consideration filter when linting

huxiaohei:
## What

`kubernetesLint` now consults the consideration filter, so the
`# vscode-kubernetes-tools: exclude` annotation suppresses linting the same way it
already suppresses schema support.

## Why

`shouldProvideSchemaFor` / `shouldSkip` in `src/yaml-support/consideration-filter.ts` is
the extension's answer to "is this document ours?". `requestYamlSchemaUriCallback` honours
it; the linter never has. So a document can be excluded from schema support and still be
linted, which is surprising and, as far as I can tell, unintentional.

The case that led me here is Kustomize patch files. They are deliberately partial
resources, so `resource-limits` flags every container in them:

```yaml
# overlays/local/patch-deployment.yaml -- merged with a base that is complete
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-server
spec:
  replicas: 2
  template:
    spec:
      containers:
        - name: server
          resources:
            requests:
              cpu: "500m"
              memory: "768Mi"
```

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -69,6 +69,7 @@ import { setActiveKubeconfig, getKnownKubeconfigs, addKnownKubeconfig } from './
 import { HelmDocumentSymbolProvider } from './helm.symbolProvider';
 import { HelmBlockMatchingProvider } from './helm.blockMatchingProvider';
 import { findParentYaml } from './yaml-support/yaml-navigation';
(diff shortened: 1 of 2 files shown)


tejhan (approved):
LGTM. Thanks for this.

PR #2009, opened 2026-08-16
feat: don't treat Kustomize patch files as complete resources

huxiaohei:
## What

Implements the TODO in `inferShouldProvideSchemaFor`:

> In future we may want to be smarter, e.g. detecting things which look like Kustomize
> patch files.

Files a `kustomization.yaml` names as patches are now excluded by the consideration
filter.

## Why

Patch files are partial resources by design — they carry only the fields to be merged over
a base — so validating them as complete objects reports problems with a file that is
correct:

```yaml
# overlays/local/patch-deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: server
spec:
  replicas: 2
  template:
    spec:
      containers:
        - name: server
          resources:
            requests:
              cpu: "500m"
              memory: "768Mi"
```

`Missing property "selector"` from the schema, plus three `resource-limits` warnings.
Nothing here is wrong; the merged output is complete.

## How

A patch isn't recognisable from its own contents — it looks like a truncated resource, and
so does a resource someone is halfway through writing. What identifies it is that a
kustomization names it. `kustomize-patch-index.ts` indexes what the workspace's
kustomizations point at.

The lookup has to be synchronous (vscode-yaml's contributor API is synchronous, and
linting runs on every keystroke), so the index is maintained in the background by a
`FileSystemWatcher` rather than reading kustomization.yaml on demand. Activation doesn't
block on the initial scan; `onDidChange` tells consumers to reconsider when it lands.

Only patch fields are indexed — `patches[].path`, and the deprecated
`patchesStrategicMerge` and `patchesJson6902[].path`. Entries under `resources` are
complete objects and keep their schema and lint support. Inline patches and remote
references name no local file and are skipped.

## Trade-offs I'd like your view on

1. **Patch files lose completion too**, since this excludes them from schema support
   rather than only from diagnostics. That seems the right trade against reporting errors
   that aren't real, and `# vscode-kubernetes-tools: include` overrides it per file — but
   if you'd rather suppress only diagnostics, that's a different mechanism and I'm happy
   to go that way instead.

2. **Already-open documents don't refresh** when the index changes. `updateYAMLSchema`
   can only invalidate — per the vscode-yaml limitation your own comment there records.
   Newly opened documents are correct.

## Relationship to #2008

Independent; they compose. This one fixes the schema errors on its own. The
`resource-limits` warnings go away once the linter consults the same filter, which is
#2008. Neither depends on the other to merge.

## Testing

- `npm test` — 73 passing (was 57). Path extraction is covered directly (11 cases:
  each patch field, relative paths, `resources` left alone, inline patches, remote
  references, malformed input). The index itself is exercised against a fixture
  workspace, which `runTest.ts` now opens — tests that use untitled documents are
  unaffected by having a workspace open.
- I checked the workspace tests actually bite by breaking the lookup and confirming the
  positive case fails.
- `npm run lint` — clean, no new warnings.

The index is driven directly by those tests rather than through extension activation:
tests run against the TypeScript output in `out/`, whereas activating the extension loads
the webpack bundle in `dist/`, which `npm test` doesn't rebuild.

### DIFF
--- a/src/yaml-support/kustomize-patch-index.ts
+++ b/src/yaml-support/kustomize-patch-index.ts
@@ -22,13 +22,22 @@ const KUSTOMIZATION_GLOB = '**/[kK]ustomization.{yaml,yml}';
 // kustomization only re-indexes that file.
 const patchesByKustomization = new Map<string, readonly string[]>();
 
+// Reads are asynchronous and can overlap, so a slow read of an earlier revision could
+// otherwise finish last and republish stale paths. Every attempt to change what we know
+// about a kustomization takes a new token; only the newest token may write.
+const generations = new Map<string, number>();
+
+let rescanGeneration = 0;
+
 let allPatchPaths = new Set<string>();
+
+let published = false;
 
 const onDidChangeEmitter = new vscode.EventEmitter<void>();
 
-// Fires when the set of known patch files changes - on the initial scan and whenever a
-// kustomization is created, edited or deleted. Consumers should re-evaluate any documents
-// they have already processed.
+// Fires when the initial scan completes, and thereafter whenever the set of known patch
+// files actually changes. Editing a kustomization without altering the files it names as
+// patches does not fire. Consumers should re-evaluate any documents already processed.
 export const onDidChange = onDidChangeEmitter.event;
 
 export function isKustomizePatch(uri: vscode.Uri): boolean {
@@ -45,7 +54,10 @@ export function initialise(context: vscode.ExtensionContext): void {
     watcher.onDidCreate(reindexOne, undefined, context.subscriptions);
     watcher.onDidChange(reindexOne, undefined, context.subscriptions);
     watcher.onDidDelete((uri) => {
-        patchesByKustomization.delete(normalisePath(uri.fsPath));
+        const key = normalisePath(uri.fsPath);
+        // Take a token as well, so that a read still in flight can't resurrect the entry.
+        generations.set(key, nextGeneration(key));
+        patchesByKustomization.delete(key);
         republish();
     }, undefined, context.subscriptions);
 
@@ -58,8 +70,22 @@ export function initialise(context: vscode.ExtensionContext): void {
 }
 
 async function rescan(): Promise<void> {
+    const generation = ++rescanGeneration;
     const kustomizations = await vscode.workspace.findFiles(KUSTOMIZATION_GLOB);
-    patchesByKustomization.clear();
+    if (generation !== rescanGeneration) {
+        return;  // another rescan started while we were walking the workspace
+    }
+
+    // Prune what has gone rather than clearing outright: clearing would leave the index
+    // empty for the duration of the scan, briefly un-excluding files that are still
+    // patches.
+    const found = new Set(kustomizations.map((uri) => normalisePath(uri.fsPath)));
+    for (const key of Array.from(patchesByKustomization.keys())) {
+        if (!found.has(key)) {
+            patchesByKustomization.delete(key);
+        }
+    }
+
     await Promise.all(kustomizations.map(indexKustomization));
     republish();
 }
@@ -71,15 +97,27 @@ async function reindexOne(uri: vscode.Uri): Promise<void> {
 
 async function indexKustomization(uri: vscode.Uri): Promise<void> {
     const key = normalisePath(uri.fsPath);
+    const generation = nextGeneration(key);
+    generations.set(key, generation);
     try {
         const bytes = await vscode.workspace.fs.readFile(uri);
         const parsed = yaml.load(Buffer.from(bytes).toString('utf8'));
+        if (generations.get(key) !== generation) {
+            return;  // superseded while we were reading
+        }
         patchesByKustomization.set(key, patchFilePaths(parsed, path.dirname(uri.fsPath)));
     } catch {
         // Unreadable or mid-edit and not yet valid YAML. Drop what we knew rather than
         // keeping a stale answer: over-reporting warnings is better than hiding them.
+        if (generations.get(key) !== generation) {
+            return;
+        }
         patchesByKustomization.delete(key);
     }
+}
+
+function nextGeneration(key: string): number {
+    return (generations.get(key) ?? 0) + 1;
 }
 
 function republish(): void {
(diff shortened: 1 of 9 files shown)


huxiaohei:
Thanks for the review — all four were real. Fixed in a1b36be, kept as a separate commit so the delta since your review is visible.

**Windows path casing in the tests.** You're right, and it would have been a hard CI failure rather than a flake: the matrix includes `windows-latest`, `path.resolve` keeps the drive-letter casing from cwd, and `normalisePath` lowercases it on win32, so every positive assertion would have compared `D:\…` against `d:\…`. The expectations now go through the same normalisation, which makes both sides literally the same expression on any platform. I can't run the windows leg locally, so that argument is what I'm relying on — if you'd rather I made the test not care about casing at all, say so.

**Overlapping reads publishing stale paths.** Fixed. Every attempt to change what we know about a kustomization now takes a token, and only the newest token may write. Deletion takes a token too, otherwise a read still in flight could resurrect an entry that had just been deleted.

While in there I also stopped `rescan()` clearing the index up front. It left the index empty for the whole workspace walk, which briefly un-excluded files that were still patches — the same class of problem, just triggered by a rescan rather than by overlapping reads. It prunes what has gone instead, and drops its own result if another rescan started meanwhile.

**`republish()` firing when nothing changed.** Fixed — and thanks for catching it, because the code was contradicting the event's own doc comment, which claimed it fired when the set changed.

One deliberate exception: the first publish always fires, even if the set is empty. Consumers need to know the initial scan has landed so they can reconsider documents they processed while the index was still empty. That's now stated in the comment on `onDidChange` rather than implied.

**What isn't covered by tests.** The path extraction and the workspace lookups have tests; the token guard and the no-op suppression don't. Both would need to drive watcher timing or write to the fixture during the run, and there's no precedent for FS-mutating tests in the suite. Happy to add them if you want that precedent set — I'd rather ask than introduce a flaky test.

huxiaohei:
Hi @Tatsinnit — gentle nudge on this one.

Since your review request on Aug 17: Copilot's four comments are all addressed in a1b36be and now resolved, and DCO, Dependency Quality, License Compliance and Security Analysis are green. The Development build and CodeQL Advanced runs are sitting in "waiting for approval" — if you're able to approve those, it would also confirm the `windows-latest` leg of the matrix, which is the one part of the fix I couldn't verify locally.

No rush on the review itself, but two open questions from the description would help me know whether this is on the right track:

1. Excluding patch files from schema support loses completion on them too. I think that's the right trade against reporting errors that aren't real, but if you'd rather suppress only diagnostics, that's a different mechanism and I'm happy to rework it.
2. The token guard and the no-op suppression aren't covered by tests — both would need FS-mutating tests, and there's no precedent for those in the suite. Happy to add them if you want that precedent set.

Also flagging #2008 — independent of this one, all checks passing, no comments.

src/yaml-support/kustomize-patch-index.ts
```diff
@@ -0,0 +1,220 @@
...
+    return allPatchPaths.has(normalisePath(uri.fsPath));
+}
+
+export function initialise(context: vscode.ExtensionContext): void {
+    const watcher = vscode.workspace.createFileSystemWatcher(KUSTOMIZATION_GLOB);
+    context.subscriptions.push(watcher, onDidChangeEmitter);
+
+    watcher.onDidCreate(reindexOne, undefined, context.subscriptions);
+    watcher.onDidChange(reindexOne, undefined, context.subscriptions);
+    watcher.onDidDelete((uri) => {
```
Tatsinnit: This deletion path is not guaranteed to run when a directory containing a kustomization is deleted or renamed. VS Code documents that recursive watchers may report only the parent-folder event, and the parent does not match this file-only glob. In that case `patchesByKustomization` retains the deleted file indefinitely and those patch paths remain excluded. Could directory deletion/rename trigger a reconciliation scan, or could the watcher include ancestor events?

src/yaml-support/kustomize-patch-index.ts
```diff
@@ -0,0 +1,220 @@
...
+// We can't tell that from the file itself - a patch looks like a truncated resource, and
+// so does a resource someone is halfway through writing. What does distinguish them is
+// that a kustomization.yaml names the patch. So we index what the kustomizations in the
+// workspace point at, and let the consideration filter exclude those files.
+//
+// The lookup must be synchronous: vscode-yaml's schema contributor API is synchronous, and
+// linting runs on every keystroke. Hence an index maintained in the background rather than
+// reading kustomization.yaml on demand.
+
+const KUSTOMIZATION_GLOB = '**/[kK]ustomization.{yaml,yml}';
```
Tatsinnit: This glob misses the canonical extensionless `Kustomization` filename. Kustomize recognizes exactly `kustomization.yaml`, `kustomization.yml`, and `Kustomization` (`api/konfig/general.go`), while this also matches unsupported `Kustomization.yaml`/`.yml` on case-sensitive systems. Could we match the exact three names and add an extensionless fixture/test?

src/yaml-support/kustomize-patch-index.ts
```diff
@@ -0,0 +1,220 @@
...
+        patchesByKustomization.delete(key);
+        republish();
+    }, undefined, context.subscriptions);
+
+    vscode.workspace.onDidChangeWorkspaceFolders(() => { rescan(); }, undefined, context.subscriptions);
+
+    // Deliberately not awaited: activation shouldn't block on walking the workspace. Until
+    // the scan lands nothing is excluded, and the onDidChange event tells consumers to
+    // reconsider once it does.
+    rescan();
```
Tatsinnit: Could we handle the initial scan before registering schema support, or otherwise make vscode-yaml reconsider the triggering document? Opening a YAML file is what activates this extension, so that document is already open when this fire-and-forget scan starts. If vscode-yaml asks `shouldProvideSchemaFor` before the scan completes, the file is treated as a normal resource. The later `onDidChange -> updateYAMLSchema()` only refreshes the cached cluster schema and, as `yaml-schema.ts` notes, does not make vscode-yaml reassess an already-open document. This means the first patch opened in a fresh window can keep the false diagnostics until it is reopened.

src/yaml-support/kustomize-patch-index.ts
```diff
@@ -0,0 +1,220 @@
...
+    // empty for the duration of the scan, briefly un-excluding files that are still
+    // patches.
+    const found = new Set(kustomizations.map((uri) => normalisePath(uri.fsPath)));
+    for (const key of Array.from(patchesByKustomization.keys())) {
+        if (!found.has(key)) {
+            patchesByKustomization.delete(key);
+        }
+    }
+
+    await Promise.all(kustomizations.map(indexKustomization));
```
Tatsinnit: The rescan generation is only checked before these asynchronous per-file reads. An older rescan can therefore publish after a newer workspace-folder rescan: scan A starts reading a kustomization, its folder is removed and scan B prunes it, then A finishes and writes the entry back because pruning did not advance the per-file generation. Could the scan build a private result and commit it only if its rescan token is still current, or recheck and invalidate tokens after the reads?

Tatsinnit:
Thank you for this. The feature direction and the rest of the changes look good. I left four focused comments around first-open schema timing, the canonical Kustomize filename set, rescan ordering, and directory deletion handling.

FYI @bosesuneha @davidgamero @yolossn @tejhan ❤️

PR #2017, opened 2026-09-18
ci: group dependabot minor and patch updates

bosesuneha:
Dependabot opens one PR per dependency right now, which adds up fast.

What changes:

- Minor and patch bumps are grouped into a single PR per ecosystem (npm and github-actions).
- Major bumps are deliberately left out of the groups, so they still arrive as individual PRs and can be reviewed on their own.
- Schedule moves from weekly to the 1st and 15th of each month at 09:00 UTC. Dependabot has no twice-a-month interval, so this uses a cron expression.

No `ignore` rules were added, so security updates are not suppressed.

One thing worth flagging: the old `codeql` group is folded into the general actions group. That group had no `update-types`, so it was grouping majors too. Codeql major bumps will now come through as individual PRs. Happy to keep codeql separate if you would rather not change that.


Tatsinnit (approved):
💡 Thank you so much for this, for wider audience notes:

What this PR is doing: 

- npm and GitHub Actions are checked on the 1st and 15th at 09:00 UTC.
- All npm minor/patch updates are combined into one PR.
- All GitHub Actions minor/patch updates are combined into one PR.
- Major npm and Actions updates remain individual PRs.
- CodeQL minor/patch updates join the general Actions group.
- CodeQL major updates become individual PRs.

Security Assessment

Overall, the change looks safe overall:

- It changes only Dependabot scheduling and PR grouping.
- It does not run code or change workflow triggers.
- It does not change permissions, credentials, secrets, dependencies, or release behavior.
- The YAML parses and validates against the current Dependabot schema.
- Dependabot security updates are enabled.
- No ignore or allow rule suppresses security fixes.
- All current PR checks pass.

Approving and merging it ❤️

